PR #302, opened 2016-03-16
Re-factor README

tienne-B:
This change removes dead links and contractions from the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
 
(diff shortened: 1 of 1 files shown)


philipbelesky:
Hi Tienne, thanks for bringing this to our attention. In particular those dead links need to be corrected, although in most case we would like to relink them to their new locations, rather than remove them. In the case of the contraction removals we are not sure why these need to be removed, and the current usage is consistent with our approach to the other documentation. 

For both of these reasons I'll be closing this without merging it, although you are welcome to follow up with a more focused pull request if there are particular grammar or readability issues in the README you'd like to bring to raise. I'll also relink those dead links ASAP.

PR #520, opened 2017-11-11
Small tweak to management command confirmation

BenMusch:
This change makes the management command confirmation operate more like traditional shell commands with yes/no input

### DIFF
--- a/tabbycat/utils/management/base.py
+++ b/tabbycat/utils/management/base.py
@@ -202,11 +202,12 @@ class RoundCommand(TournamentCommand):
     def _confirm_rounds(self, rounds, **options):
         if not options["confirm"]:
             self.stdout.write(self.style.WARNING("WARNING! You are about to {} from the following rounds:".format(self.confirm_round_destruction)))
+
             for r in rounds:
                 self.stdout.write(self.style.WARNING("  [{t}]: {r}".format(
                     t=r.tournament.name, r=r.name)))
-            response = input("Are you sure? [y/N]").lower()
-            if response != "yes" and response != "y":
+
+            if input("Are you sure? [y/N]").lower() not in ["yes", "y"]:
                 raise CommandError("Cancelled by user.")
 
         elif options["confirm"] != options["round_selection"]:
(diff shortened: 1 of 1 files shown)


tabbycat/utils/management/base.py
```diff
@@ -202,11 +202,12 @@ def get_rounds(self, options):
...
         if not options["confirm"]:
             self.stdout.write(self.style.WARNING("WARNING! You are about to {} from the following rounds:".format(self.confirm_round_destruction)))
+
             for r in rounds:
                 self.stdout.write(self.style.WARNING("  [{t}]: {r}".format(
                     t=r.tournament.name, r=r.name)))
-            response = input("Are you sure? [y/N]").lower()
-            if response != "yes" and response != "y":
+
+            if input("Are you sure? [y/N]").lower() not in ["yes", "y"]:
```
BenMusch: Removing the `response` variable made codeclimate green. Happy to revert if that's not a concern

philipbelesky:
Thanks for the PR! I'd actually forgotten we had CodeClimate hooked up to pull requests; in general we don't worry about full compliance with it. Although we probably should at some point go through and address at least the most egregious cases it has flagged.

czlee:
I had actually deliberately made this different to traditional shell commands because it typically shows just before irreversible loss of a lot of data (as opposed to just a "check before proceeding"). How strongly do you feel about this? I'd prefer to keep it as requiring something more than `y` to proceed; if `yes` is too similar to bash y/n responses, then I'd move towards a system where they have to type in the tournament slug to proceed (similar to Heroku's prompts when doing destructive actions).

BenMusch:
@czlee Is this usually used by people who are running a tournament? My thought was that it's really only used by developers, since it's unlikely that any other user would not be modifying the tournament from the CLI, and probably wouldn't even know how to find which command to run

czlee:
It might be used by a developer who is running a tournament—the point of these checks isn't to protect against technical illiteracy, it's to protect against human error.

BenMusch:
gotcha -- That makes sense. I'm fine with reverting this PR

PR #545, opened 2018-01-21
Tests in 'participants' and test score validation

daigo-kimura:
This change ...
* makes the tests in 'participants' to be picked up and compatible with the develop branch
* adds adjudicator test score validation tests

### DIFF
--- a/tabbycat/importer/forms.py
+++ b/tabbycat/importer/forms.py
@@ -317,15 +317,13 @@ class AdjudicatorDetailsForm(SharedBetweenTournamentsObjectForm, BaseInstitution
             'test_score': _("Rating"),
         }
 
-    def clean(self):
-        clean_data = super().clean()
-        if 'test_score' not in clean_data:
-            return clean_data
+    def clean_test_score(self):
+        test_score = self.cleaned_data['test_score']
         min_score = self.tournament.pref('adj_min_score')
         max_score = self.tournament.pref('adj_max_score')
-        if clean_data['test_score'] < min_score or max_score < clean_data['test_score']:
+        if test_score < min_score or max_score < test_score:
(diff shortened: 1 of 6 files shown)


czlee:
Nice catch with the participants tests, thanks so much!

I'd actually rather not catch out-of-range adjudicator scores at the model level—because its validity can be modified by changing tournament preferences, it's not a hard constraint, and an out-of-range score isn't fatal to any operation. We should already have a warning message for this though (see 3bbe7f132a62d4ff624f235174846c751727b57b), did that work?

daigo-kimura:
Yes, it works. Well, I revert commits regarding the validation and its tests.

daigo-kimura:
There is out-of-range score validation on "add feedback" form. So, how about applying the same thing on simple importer form?

philipbelesky:
I think adding it to the importer form makes sense — both to catch typos and instances where users don't realise the ranges need to be configured and are inadvertently creating scores out of range.

daigo-kimura:
Thanks. I added it.
Do I need to modify codes to make commits pass codeclimate tests?

czlee:
No need, thanks! I'll look at merging this in next week.

czlee:
All merged in, thanks! (I squashed some commits and rebased it, so they're now d8180907a2974c0b43924348193b7a885c044b69, 821a15c929b1077fa7c7abb2fb9e786dbc0f1b8f, db5de82554ef5dbd9d1afd5b5f9bb9c40db0f41d and 23b88798b70c321350190009978e6f8a2285210d.)

philipbelesky:
Thanks @daigo-kimura! Let us know if you were interested in doing any further development work; we could point you to some issues/features that we would love a hand working on :)

PR #579, opened 2018-04-12
Create page for important feedback with default query

### LINKED ISSUE #477: Add an 'important feedback' page to highlight feedback that is outside the expected range for a chair/panellist/trainee
So, at the risk of expanding the feedback system, there is some low hanging fruit in terms of more easily surfacing 'important' feedback. Currently we need to find things by source/target; which makes it hard to look across all pieces of feedback to highlight the most critical ones. A simple 'win' here would be to make a feedback page that highlights feedback that is outside the 'expected' range based on position; say:

- A chair that receives a 0 or 1 (normalised to a 0-5 range)
- A trainee that receives a 3-5 (normalised to a 0-5 range)
- A panellist that receives a 4-5 or a 0-1

The other way to play thing would be to show results that are more than 2 points higher or lower than an adj's current rating. While some configuration might be valuable here (and easy given the queries can be parametrised) a set of sensible defaults would still be a major win and relatively easy to create.

tfpk:
Fixes #477. Pretty basic, the only notable thing is that it uses the formula `feedback_importance=F('score') - F('adjudicator__test_score')`, `abs(feedback_importance) > 2`.


philipbelesky:
Hey, thanks for the PR! Just a quick thing, would you mind reissuing this against the `feature/579-important-fb` branch? I might make a few changes and run them by you before we merge in to develop.

tfpk:
Is doing that good enough, or do I have to create a new pull request?

Issue #580, opened 2018-04-13
Fix nav bar code


tfpk:
The current code for the navbar is unintelligent and messy, including a large if block that causes issues (for example, checking if "import" is in the path will mean that "important" below does not get recognised). It can be refactored into a neater javascript solution.

philipbelesky:
Closed in #589. Thanks again!

PR #589, opened 2018-04-25
replace context processor for menu selection with jquery

tfpk:
From: #580


philipbelesky:
Hey thanks for adding this! The code that this remove was a particularly cumbersome hack, so great to have something much more elegant put in place instead.

czlee:
Just wanted to echo what @philipbelesky said, this is great, thanks so much for doing it!

czlee:
Just a heads-up, this changed the color of the links in this menu:

image

I don't think this is necessarily a bad thing, just checking if it was intentional?

philipbelesky:
I think that was me in 471ea476400f1ebe600cd449985d11aa2e37c327, not in this PR. But yes those menu links in the drop downs should now match the general top-menu.

PR #599, opened 2018-05-12
Fix dead link for motion statistics

tienne-B:
This change updates the link to the code for running the chi-squared test to the new dedicated file. The line anchor was also changed to the `_annotate_χsquared` method, as that seems to be the equivalent position, and it also includes a comment that explains the statistic in more technical terms.


philipbelesky:
Great catch @tienne-B, thanks for the PR!

PR #622, opened 2018-06-04
Make private URL distribution e-mails customisable

### LINKED ISSUE #531: Make private URL distribution e-mails customisable
A nice-to-have, allow private URL distribution e-mails to be edited before sending.

- Should be considered an advanced feature; i.e., it shouldn't be part of the standard workflow.
- Should include default text (basically the string we currently use).
- Validation should check that all merge fields are present (to allow for typos).
- Should show a completed example, or even better allow you to switch between all of them like in mail merge (but the latter is probably a bit overkill). Actually, this part probably should be part of the standard workflow.

tienne-B:
This resolves #531, although in a somewhat different direction.

What this change does is adds text fields (and areas) in the forms which were previously just buttons to send the mass emails. Using the cursory CSS I saw being used, the forms are styled (well, I hope). With the messages visible, there is a sense of what the emails will look like. Basic use is just to not touch the message, but a more advanced workflow would allow for that.

Point 3 is not implemented; just trying out Python / Django and haven't quite figured out that yet (or much if anything). If validation is made, it should only take `%url` into account.

I also added a line about email debugging to the Contributors file. However, I did not add anything to the documentation about this (as the only mention to email was in passing, and the instructions are below the textarea), nor added an entry in the changelog.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -34,6 +34,10 @@ Getting started with development
 
     $ pip install -r 'requirements_development.txt'
 
+- The email backend should be changed in ``local_settings.py`` to display sent messages in ``STDOUT``, not by real email. Insert::
+
+    EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+
 Style guide
 ===========
 
--- a/tabbycat/privateurls/forms.py
+++ b/tabbycat/privateurls/forms.py
@@ -1,0 +1,32 @@
+from django import forms
+
+from django.utils.translation import gettext as _
+
+
+class EmailBodyField(forms.CharField):
+
+    label = _("Message")
+    required = True
+    widget = forms.Textarea
+    help_text = _(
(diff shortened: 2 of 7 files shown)


tienne-B:
I guess another way would be to make the email form a true form, and the feedback one a form-set of such forms. The views would then also be `FormView`s for sending emails. With that, the `{{ field.error }}` can be used, checking to see if the variables are present. (would take quite some time for me to figure any of this out), but what I have here should be already good.

tienne-B:
No problem, @philipbelesky. Take your time. I'll have an updated patch-set with the other route in a day.

As for an update on the other route, a formset would not be possible as the initial values would have to be identical between debaters/judges in the feedback. However, this route does get rid of some complexity in my view and is better than what I provided.

czlee:
Just wanted to echo what @philipbelesky said, this is something we've really wanted to do so it's exciting to see a potential implementation! As it turns out, I too will have more time to turn to this next week, hopefully.

Issue #532, opened 2018-01-11
Allow current wins/team points to be e-mailed to all speakers


czlee:
As a check, à la @philipbelesky's magic at WUDC México 2018.

- E-mail text should be editable as an advanced step, similarly to #531.
- We'll need two default texts, one for two-team formats ("wins") and one for BP ("team points").
- Text should (be required to) include a link to the current team standings, if that page is enabled.
- I think it's best for the count to be merged into the subject line, so that people see the crucial information in their notifications.

philipbelesky:
It might as well also provide a link to the team-specific pages if those are enabled?

If we wanted to get fancy we could also send an email to adjudicators confirming their ballot results —  this wasn't possible, but was desirable at Mexico.

philipbelesky:
Oops — saw this before #535

tienne-B:
The message would not be editable as #531, rather like #535 in a option in the email section. With that, the settings could be added to the BP preset. I might take this on after the others are reviewed; some prerequisites to this are in the PRs, and I don't want to swamp both of you with these.

tienne-B:
So a simple message, just point count (and link). The emails would thus not include information such as the team they just debated against, referring them to the standings URL. I'll take this task. With the default messages in the preferences, a button to send these emails to the speakers could be added to the Confirm Round Advance page.

PR #626, opened 2018-06-06
Send receipt email to judges with their ballot

### LINKED ISSUE #535: E-mail entered ballots to adjudicators once entered
A check similar to #532.

For consensus formats, either e-mail to chair or to all voting members.
For voting formats, e-mail each adjudicator just their own ballot.

tienne-B:
The commit message has some pertinent information.

This resolves #535 by executing an emailing routine after ballot submission, through the `form_valid()` method of `BaseBallotSetView`. Through that, the new routine takes the `ballotsub` and `DebateResults` separately (as it would give an error otherwise).

Review when you have time; no rush :)

### DIFF
--- a/tabbycat/results/templates/public_enter_results.html
+++ b/tabbycat/results/templates/public_enter_results.html
@@ -69,7 +69,10 @@
         <input class="save btn btn-success btn-block " type="submit" value="Submit Ballot(s)" tabindex="{{ form.nexttabindex }}"/>
         <div class="text-center pt-3 small text-muted">
           {% trans "When submitting this form your IP address will be stored for logging purposes." %}
-          {% trans "Emails will be sent to adjudicators to validate their responses." %}
+          {% if pref.enable_ballot_receipts %}
+            <br />
+            {% trans "Emails will be sent to adjudicators when the ballot is confirmed." %}
+          {% endif %}
         </div>
       </div>
     </div>
--- a/tabbycat/results/templates/public_enter_results.html
+++ b/tabbycat/results/templates/public_enter_results.html
@@ -69,6 +69,7 @@
         <input class="save btn btn-success btn-block " type="submit" value="Submit Ballot(s)" tabindex="{{ form.nexttabindex }}"/>
         <div class="text-center pt-3 small text-muted">
           {% trans "When submitting this form your IP address will be stored for logging purposes." %}
+          {% trans "Emails will be sent to adjudicators to validate their responses." %}
         </div>
       </div>
     </div>
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -31,7 +31,7 @@ from .models import BallotSubmission, TeamScore
 from .tables import ResultsTableBuilder
 from .result import DebateResult
 from .prefetch import populate_confirmed_ballots
-from .utils import get_result_status_stats, populate_identical_ballotsub_lists, send_email_to_adjs
+from .utils import get_result_status_stats, populate_identical_ballotsub_lists, send_ballot_receipt_emails_to_adjudicators
 
 logger = logging.getLogger(__name__)
 
@@ -269,7 +269,12 @@ class BaseBallotSetView(LogActionMixin, TournamentMixin, FormView):
 
     def send_email_receipts(self):
         # For proper error handling for admin/assistants, overwrite this
-        send_email_to_adjs(DebateResult(self.ballotsub).as_dicts(), self.debate)
+        try:
+            send_ballot_receipt_emails_to_adjudicators(DebateResult(self.ballotsub).as_dicts(), self.debate)
+        except (SMTPException, ConnectionError):
+            return False
+        else:
+            return True
 
     def form_valid(self, form):
         self.ballotsub = form.save()
@@ -279,7 +284,7 @@ class BaseBallotSetView(LogActionMixin, TournamentMixin, FormView):
             self.ballotsub.save()
 
             if self.tournament.pref('enable_ballot_receipts'):
-                self.send_email_receipts()
+                self.email_receipts_sent = self.send_email_receipts()
 
         self.add_success_message()
         self.round = self.ballotsub.debate.round  # for LogActionMixin
@@ -308,21 +313,19 @@ class BaseBallotSetView(LogActionMixin, TournamentMixin, FormView):
 
 class BallotEmailWithStatusMixin:
     def send_email_receipts(self):
-        success = True
         try:
-            send_email_to_adjs(DebateResult(self.ballotsub).as_dicts(), self.debate)
+            send_ballot_receipt_emails_to_adjudicators(DebateResult(self.ballotsub).as_dicts(), self.debate)
         except SMTPException:
             messages.error(self.request, _("There was a problem sending ballot receipts to adjudicators."))
-            success = False
+            return False
         except ConnectionError:
             messages.error(self.request, _("There was a problem connecting to the e-mail server when trying to send ballot receipts to adjudicators."))
-            success = False
+            return False
+        else:
+            return True
 
-        if success:
-            messages.success(self.request, _("Email receipts sent successfully."))
 
-
-class AdministratorBallotSetMixin(BallotEmailWithStatusMixin, AdministratorMixin):
+class AdministratorBallotSetMixin(AdministratorMixin):
     template_name = 'enter_results.html'
     tabroom = True
 
@@ -330,7 +333,7 @@ class AdministratorBallotSetMixin(BallotEmailWithStatusMixin, AdministratorMixin
         return reverse_round('results-round-list', self.ballotsub.debate.round)
 
 
-class AssistantBallotSetMixin(BallotEmailWithStatusMixin, AssistantMixin):
+class AssistantBallotSetMixin(AssistantMixin):
     template_name = 'assistant_enter_results.html'
     tabroom = True
 
@@ -338,7 +341,7 @@ class AssistantBallotSetMixin(BallotEmailWithStatusMixin, AssistantMixin):
         return reverse_tournament('results-assistant-round-list', self.tournament)
 
 
-class BaseNewBallotSetView(SingleObjectFromTournamentMixin, BaseBallotSetView):
+class BaseNewBallotSetView(SingleObjectFromTournamentMixin, BallotEmailWithStatusMixin, BaseBallotSetView):
 
     model = Debate
     tournament_field_name = 'round__tournament'
@@ -347,7 +350,10 @@ class BaseNewBallotSetView(SingleObjectFromTournamentMixin, BaseBallotSetView):
     pk_url_kwarg = 'debate_id'
 
     def add_success_message(self):
-        messages.success(self.request, _("Ballot set for %(debate)s added.") % {'debate': self.debate.matchup})
+        message = _("Ballot set for %(debate)s added.") % {'debate': self.debate.matchup}
+        if getattr(self, 'email_receipts_sent', False):
+            message += _(" Email receipts sent.")
+        messages.success(self.request, message)
 
     def get_error_url(self):
         return self.get_success_url()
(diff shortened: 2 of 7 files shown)


tienne-B:
Darn useless Codeclimate...

Also, according to your `CONTRIBUTORS` file, I should not be submitting a PR against the `develop` branch. However, there is no other branch that I could reasonably use (can't make a PR against a non-existant topic branch...) Is this fine?

czlee:
Wait, yeah, that instruction seems weird, since you can't create a branch on our repository yourself. @philipbelesky, should that just say to submit PRs against develop?

Also don't stress too much about Code Climate, we haven't perfectly tuned settings to what we want, I'll let you know if we actually want your code to be more decomposed, though it sure seemed fine to me on a first glance.

philipbelesky:
Great to see the subject and message are configurable! One thing to note is that there a number of circumstances when people may not want ballots to be auto-emailed — either because they are using the kitten version (no third party plugins), running offline/locally, or have imported data with little to no emails. 

As such my first thought is that there should be a general preference boolean to enable/disable this behaviour (perhaps default to disable?). It might also be worth spinning out such 'emails' preferences into their own configuration section; if just so they are more visible. 

One other minor thing is that while the `SMTPException` or `ConnectionError` conditions correctly log the exception that process is generally invisible to users. It might be worth passing a `message` (i.e. an inline alert to next rendered page) notifying of those failures (and perhaps a success state also).

I'd be happy to add those changes myself though after this is merged. Are you ok merging this @czlee ?

tienne-B:
1. (Option to email) The default should be enabled in my mind:
- 1. Tabbykitten is a minor case, and if there are not many inputted email addresses,
- 2. I've addressed this somewhat in #622, where local Tabbycat installs could use the terminal email setting (could even be added to `local_settings.example`) so that the full Tabbycat functionality is still visible with offline installs.
- 3. Adjudicators without emails are ignored by this system. Even if there is a small number of ones with email, this could still be useful.
2. A new settings panel could be made, and this enable/disable switch could also be used to display the links to email private URLs.
3. Good idea, but these messages about email success/failure should not be viewable in the public view; it should not concern them. Without creating a new table (like `privateurls_privateurlsentmailrecord`), it would be impossible to show failure/success for each sent ballot in a meaningful way, such as indicating the state in the debate table. That would be overkill in my mind (as is the former DB table...)

I will submit a patch that deals with these points, but my choice of message for `messages` may have to be refined a bit, as well as the last part of point 3 about selectively showing these messages.

czlee:
I'll have a look in the next couple of days, but I'm really excited about this! 😃 (sorry, just behind with other parts of life)

philipbelesky:
For (1) I'd be interested in CZ's thoughts. My thinking is that our out of the box experience is catered for novice users and standard tournament setups; any unexpected or new behaviours are generally things that people need to opt-in to. At present emails are only ever sent due to explicit user action rather than being an implicit side effect of the standard round workflow. 

It's also hard to get exact stats, but looking at error logs I would guess Tabbykitten installs are 15-40% of the base. Although if the relevant email service is not available that could be mentioned within the error messages themselves.

With regards to (3) I was thinking of only doing so in the admin-facing version of the results form. Even at large majors the norm seems mostly to still use a paper ballot based workflow with Tabbycat, so it would be valuable (and actionable) for tab admins/assistants to see whether emails are being sent successfully or not.

tienne-B:
I can revert the addition of the new options section very easily if it is not wanted. It could end up having disparate settings, such as this and #532 for instance. The switch for whether to send receipt emails could be to blank the subject line as an indication.

We agree on point 3, but I don't really see how we can selectively show these messages, as sending email is done no matter the submission method (through the base view) yet the messages must take it into account. Am I missing something (a way to do this)?

czlee:
Thoughts, having reviewed the commits but not tried it out:

(1) I don't have strong feelings about whether the default should be enabled or disabled, given that you probably entered emails into the system if and only if you intended to use them. But I do think it's important for users to be made aware that the system is going to send an email (approximately) _at the point where it's about to happen_; it shouldn't be something that just happens. I think this applies regardless of what the default is.

(3) The public, assistant and admin result views are all distinct, so you should be able to condition the messages on whether it's public or assistant/admin by only including the relevant method in the relevant subclasses. That is, at the moment, it's in the very top-level base class; you could put it lower in the class hierarchy. You might need to add a level to the hierarchy or make it a mixin in order to get it to be included in the subclasses you want without duplicating code.

(4) Question: Does this send emails only when they're confirmed, or on any submission? Would it avoid sending emails when a ballot is discarded?

As an aside, I hadn't thought of storing failures in the database, but it's not a bad idea. I disagree that it's overkill; situations arise where you need to dive into history. But that's a subject for another day; it's not necessary for an MVP and it needn't be part of this pull request.

Thanks again! I should get a chance to take it for a test spin by the end of Tuesday (PDT).

tienne-B:
Should point 2 (about spinning off email-type preferences) be implemented? I do not think it should, as it fractures the current model of preferences being by "application," and emailing crosses (or could in in the future) these various functionalities of Tabbycat.

(3) [x] Done

(4) It sends on the creation of a new ballot set. My thought is that it should not wait for the ballot to be confirmed, but after any submission so that there is more time for judges to spot if there's any error. As it is not prompted when editing submissions, only on creation, it would not send emails if the ballot is then marked as discarded.

(5) If there is a table created for this, it should not be of just the errors, but of all the emails sent (like `privateurls_privateurlsentmailrecord`) to give a complete log, possibly also to avoid double-sending if a new ballot submission was made without any changes to the specific judges' marks. That would be for another day (what do you mean by "MVP" in this context?)

philipbelesky:
(2). We can definitely hold on. We've been looking to reorganise the preferences as is to increase their clarity so this can be done as part of that. 
(4). I'm a little unsure of. If a ballot is entered with an error but then edited that would then not send an email. In a 'correct' double checking workflow the ballot would be discarded/recreated generating a second email, but this doesn't necessary happen in all cases. Also from my experience at WUDC results-confirmation emails are checked very slowly so I wonder if its better to hold off for the confirmed state rather than risk confusion with multiple emails (assuming the time gap has a very minor chance of surfacing a correction).
(5). MVP = minimum viable product in this case, so yea definitely for another day.

Thanks again for all your effort implementing these!

tienne-B:
(4) Had a different workflow in mind, where ballots are just edited, not discarded/recreated. With that, the judges would not need another confirmation email, as they pointed out the error. I recognized the confusion multiple emails could have, which is why I limited it to ballot creation. We can wait for the confirmed state. The new patch does this (and is easily reverted if requested)

tabbycat/results/views.py
```diff
@@ -295,15 +306,31 @@ def post(self, request, *args, **kwargs):
...
+            send_email_to_adjs(DebateResult(self.ballotsub).as_dicts(), self.debate)
+        except SMTPException:
+            messages.error(self.request, _("There was a problem sending ballot receipts to adjudicators."))
+            success = False
+        except ConnectionError:
+            messages.error(self.request, _("There was a problem connecting to the e-mail server when trying to send ballot receipts to adjudicators."))
+            success = False
+
+        if success:
+            messages.success(self.request, _("Email receipts sent successfully."))
```
czlee: We probably want to roll this into the message in `add_success_message()` method, since currently there are always two messages that show when you confirm a ballot set, that arise from the same action. The catch is that it's not obvious to me how to do this, because if the ballot set is saved but the email sending failed, you still want two messages—a success message for the saving, and an error message for the email.

I don't have any fantastic suggestions, so if you have a better way feel free to run with that, but here's one possibility:
- Have `send_email_receipts()` return `True` if it succeeded or `False` if it failed.
- `send_email_receipts()` still adds the error messages if it failed, but adds no messages if it succeeded.
- Pass the result of `send_email_receipts()` as an optional argument into `add_success_message()`, the optional argument defaulting to `False`.
- If this argument is `True`, then `add_success_message()` appends the "Email receipts sent to adjudicators." sentence to the message. (No need to check if it was confirmed here, since it should only pass in `True` if it was in the first place. Since the two sentences are grammatically independent, I'm fine with these messages being concatenated, as opposed to string-formatted.)

On another note, this structure might help simplify things a little:
``` python
try:
    send_email_to_adjs(DebateResult(self.ballotsub).as_dicts(), self.debate)
except SMTPException:
    messages.error(self.request, _("There was a problem sending ballot receipts to adjudicators."))
except ConnectionError:
    messages.error(self.request, _("There was a problem connecting to the e-mail server when trying to send ballot receipts to adjudicators."))
else:
    messages.success(self.request, _("Email receipts sent successfully."))
    # or `return True`, if implementing the message merger above
```
(If you're wondering why this pattern is used at [privateurls/views.py:219-238](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/privateurls/views.py#L219-L238), but not at [privateurls/views.py:325-371](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/privateurls/views.py#L325-L371), it's because the latter encompasses two `try`...`except` blocks.)

tienne-B:
Also slightly changed the inheritance structure to make more sense with the success messages.

tabbycat/results/views.py
```diff
@@ -265,14 +267,28 @@ def add_success_message(self):
...
+
     def form_valid(self, form):
         self.ballotsub = form.save()
         if self.ballotsub.confirmed:
             self.ballotsub.confirmer = self.request.user
             self.ballotsub.confirm_timestamp = datetime.datetime.now()
             self.ballotsub.save()
+
+            if self.tournament.pref('enable_ballot_receipts'):
+                self.email_receipts = self.send_email_receipts()
```
czlee: This line doesn't always execute, which can cause the `self.add_success_message()` call to crash, because no other line sets `self.email_receipts`—so if the ballot isn't being confirmed, or ballot receipts are disabled, then `add_success_message()` crashes with an `AttributeError`.

I think this is the right idea, but either we need to guarantee that `self.email_receipts` always exists, or `add_success_message()` should check for its existence using `getattr(self, 'email_receipts', False)`. I think I'd prefer the latter, because it's more future-proof—the only way to _guarantee_ an instance attribute exists is to put it in the constructor, and even then that makes assumptions coupling this method to the corresponding constructor.

Also, can we call it `self.email_receipts_sent` rather than the ambiguous `self.email_receipts`? Thanks!

tabbycat/results/templates/public_enter_results.html
```diff
@@ -69,6 +69,9 @@
         <input class="save btn btn-success btn-block " type="submit" value="Submit Ballot(s)" tabindex="{{ form.nexttabindex }}"/>
         <div class="text-center pt-3 small text-muted">
           {% trans "When submitting this form your IP address will be stored for logging purposes." %}
+          {% if pref.enable_ballot_receipts %}
+            {% trans "Emails will be sent to adjudicators to validate their responses." %}
+          {% endif %}
```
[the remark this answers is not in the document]
tienne-B: Or should that one have another check for if it automatically confirms the ballot? I'd say no, adjudicators should not concern themselves with that.

czlee:
Re check for showing the "emails will be sent" message in `public_enter_results.html` (because you force-pushed, I can't comment on the line anymore): I hadn't thought of that, and now that you mention it I think it's a good idea. If it'll just send them an email I'd agree with you, but if they have a panel it sends emails to everyone on the panel, and I tend to think you should be aware if you're about to send an email to someone else. Is that okay?

tabbycat/results/views.py
tienne-B: I can still comment here!

czlee:
Looks good for merging to me, feel free to put the `public_enter_results.html` change in a new pull request if you want to do it (if I don't get to it first).

tienne-B:
I got to it first :) It was in this PR.

tabbycat/results/views.py
czlee: Oh right finding this was easier than I thought lol, thanks!

PR #628, opened 2018-06-09
Internationalize forgotten help_text string

tienne-B:
This string, a help_text in the judge feedback app was forgotten to be wrapped in the gettext format.

Just found while exploring the code-base.


tienne-B:
Going to bundle this with #629.

Issue #287, opened 2016-02-28
Buttons that change data should always be POSTs


czlee:
If a button changes data, then it should be a button that submits a POST request, and it should go to a view that only accepts POST requests.

`PostOnlyRedirectView` can facilitate this.

czlee:
In things I stumbled across today: [Django docs for `get_or_create()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create) say:
> Finally, a word on using get_or_create() in Django views. Please make sure to use it only in POST requests unless you have a good reason not to. GET requests shouldn’t have any effect on data. Instead, use POST whenever a request to a page has a side effect on your data. For more, see [Safe methods](https://tools.ietf.org/html/rfc7231.html#section-4.2.1) in the HTTP spec.

The HTTP spec essentially says that GET, HEAD, OPTIONS and TRACE requests have to be "safe", _i.e._, must not affect data, so that clients know when to present appropriate warnings to the user (and when they can use automated tools without harm).

On another note, do we have any data-changing POST requests left?

philipbelesky:
Do you mean data-changing GET requests?

czlee:
Whoops, sorry! Yes, I meant do we have any data-changing GET requests left.

philipbelesky:
Hmm not that I can think of or ID easily. Will keep an eye out.

czlee:
Availability still has several, I came across them last night (the check in all buttons)

philipbelesky:
Ah, I see. So they're:

- [ ] `update_availability_breaking_adjs` / `AvailabilityActivateBreakingAdjs`
- [ ] `update_availability_all` / `AvailabilityActivateAll`
- [ ] `update_availability_previous` / `AvailabilityActivateFromPrevious`

czlee:
Also:

- [ ] `toggle_postponed` in results/views.py

czlee:
I think this now ties to #279, in the sense that it's most easily done when the availability app itself is overhauled.

czlee:
- [ ] `adjallocation.view.CreateAutoAllocation`

czlee:
I think that might be it?

philipbelesky:
👏. Can always reopen if not?

PR #629, opened 2018-06-09
Add field to disregard select judge feedback scores

### LINKED ISSUE #538: Add a means to disregard a feedback's impact on score without discarding it
Currently if you want to remove a piece of feedback by discarding it then it still shows up on the feedback missing page. It was cited at Mexico as desirable to have an option to stop a piece of feedback impacting on a score but still keep it around. This could be a boolean type field whereby the score would then filter to exclude either `discarded` or `disregarded`.

tienne-B:
This is the meat of #538. Check the commit messages and task for further details and for moving forward.

A new field `ignored` allows scores to be marked as ignored, so that they are still counted yet do not affect the feedback score of the subject. No method to mark them after submission (other than through the DB) was added. However, admins/assistants can mark them as such during input.


philipbelesky:
Hey thanks again for the pull request @tienne-B. I've had a look through and happy to merge this so will throw to @czlee to push the button. I'll add a few extra comments in the original issue also.

tabbycat/adjfeedback/views.py
```diff
@@ -704,6 +706,31 @@ class PublicFeedbackProgress(PublicTournamentPageMixin, BaseFeedbackProgressView
...
 
 
+class IgnoreFeedbackView(SingleObjectFromTournamentMixin, RedirectView):
+    model = AdjudicatorFeedback
+    pk_url_kwarg = 'feedback_id'
+    allow_null_tournament = False
+    private_url = True
+    tournament_field_name = 'adjudicator__tournament'
+
+    def get_redirect_url(self, *args, **kwargs):
```
czlee: Just a couple of pointers for future use:

By convention, HTTP GET requests aren't meant to be used for data-changing actions, in principle. We're meant to use HTTP POST (or PUT) requests for those instead. This way, browsers will know not to cache them, allow them to be bookmarked, or store them in the user's history. We've made this mistake before ourselves, and actually once had to convert all our data-changing GETs to POSTs (see issue #287).

The flipside is that writing a web page (on the front-end) to send a POST request can be nontrivial relative to a GET request—by design, you can't just use a standard `<a>` tag for a POST request (they're always GETs). We actually haven't decided on a standard pattern for how we do POST requests yet, so we'll take care of it after merge—no need for action your side. I just wanted to give you a heads-up about the convention.

The other note is that I try to write it so that `get_redirect_url()` does only what it says, construct and return a redirect URL. The actual data-changing action should be in other methods—for form views there's a structure for this, but for something like this I just put it in `post()` and then call the super at the end of the `post()` method. There's a `PostOnlyRedirectView` for this purpose which provides a standard pattern for the back-end (but not the front-end). Again, don't worry about this, we'll deal with it after merging, once we've settled on a pattern for the front-end.

tienne-B: I know its more of an anti-pattern, but a proper button would seem out-of-place.

czlee: Yeah, this is why we weren't decided on the pattern 🙂. There are ways to do it with a link, but they involve fiddling with ajax or something like that (@philipbelesky knows more than I do).

PR #642, opened 2018-06-18
Add help_text to URL email fields

tienne-B:
Points 1 & 3 of #531.

For some reason, `help_text` would not work from inside the class. Customized the message for which parameters are available. Also reduced the variables to use angle brackets.

### DIFF
--- a/tabbycat/privateurls/utils.py
+++ b/tabbycat/privateurls/utils.py
@@ -55,9 +55,9 @@ def send_randomised_url_emails(request, tournament, queryset, url_name, url_key_
         url = request.build_absolute_uri(path)
 
         formatted_subject = subject
-        formatted_message = message.replace('%name', instance.name).replace('%url', url)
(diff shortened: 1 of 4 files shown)


tabbycat/privateurls/utils.py
```diff
@@ -55,9 +55,7 @@ def send_randomised_url_emails(request, tournament, queryset, url_name, url_key_
         url = request.build_absolute_uri(path)
 
         formatted_subject = subject
-        formatted_message = message.replace('%name', instance.name).replace('%url', url)
-        if hasattr(instance, 'team'):
-            formatted_message = formatted_message.replace('%team', instance.team.short_name)
+        formatted_message = message.replace('<NAME>', instance.name).replace('<URL>', url).replace('<TEAM>', instance.team.short_name)
```
czlee: This crashes with adjudicators: the `instance.team.short_name` isn't short-circuited when `'<TEAM>'` isn't found, so with adjudicators it raises an `AttributeError`. The old `hasattr()` check is good.

PR #646, opened 2018-06-23
Allow for team points to be emailed to teams

### LINKED ISSUE #532: Allow current wins/team points to be e-mailed to all speakers

tienne-B:
This fixes #532 by adding a button (more appropriate and looks better here) to the round confirmation view (before entering the next). Three new preferences were added to the email settings for the subject, message body, and a preface to the standings link. The third one is so it can be skipped if not enabled. The messages might need some finessing though.

The method that sends these emails is very flexible, and so can be added elsewhere. In its use here, it only send emails to the members of teams who have participated in the round, as there could not be a change for those who did not.

Get all these new email features ready :)


philipbelesky:
This looks great, thanks again @tienne-B. I'm happy to merge this and close #532 if you are @czlee ?

tienne-B:
Should I add entries for these changes to the CHANGELOG?

philipbelesky:
@tienne-B feel free to include patches for the CHANGELOG with the PRs; otherwise we will generally do an audit and update the log + provide credit just before a release is made.

tienne-B:
The newer commit changes the variables to `{{ USER }}` for use by Django templating. Easier to handle and extend.

czlee:
Hmm, I actually have mixed feelings about using Django template language in the user-defined fields—if #653 materializes, I was actually thinking of implementing HTML templates, with the user-defined text to be substituted into the template. Any thoughts? I'd say I'm happy to leave this system in place until the time comes, but it'll break existing instances if we do later change our minds about our substitution syntax.

tabbycat/privateurls/utils.py
```diff
@@ -50,16 +51,15 @@ def send_randomised_url_emails(request, tournament, queryset, url_name, url_key_
...
-        formatted_subject = subject
-        formatted_message = message.replace('<NAME>', instance.name).replace('<URL>', url)
+        variables = {'NAME': instance.name, 'URL': url}
         if hasattr(instance, 'team'):
-            formatted_message = formatted_message.replace('<TEAM>', instance.team.short_name)
+            variables['TEAM'] = instance.team.short_name
 
-        messages.append((formatted_subject, formatted_message, settings.DEFAULT_FROM_EMAIL, [email]))
+        formatted_message = message.render(Context(variables))
+        messages.append((subject, formatted_message, settings.DEFAULT_FROM_EMAIL, [email]))
```
czlee: Did we forget to render the subject template here? The emails I'm getting when testing this have subject line: "<django.template.base.Template object at 0x7fa7c87ac710>". (The other emails look fine.)

tabbycat/tournaments/utils.py
```diff
@@ -174,3 +182,41 @@ def _wrapped(tournament):
...
+        context['POINTS'] = str(team.points_count)
+        context['TEAM'] = team.short_name
+
+        subject = subject_temp.render(Context(context))
+
+        for speaker in team.speakers:
+            if speaker.email is None:
+                continue
+
+            context['USER'] = str(speaker)
```
czlee: Probably better to use `speaker.name`, since that's what we want, and the behavior of `str(speaker)` (which you should think of as "a sensible string representation of the speaker" rather than "the speaker's name") is liable to change. 🙂

tabbycat/tournaments/views.py
```diff
@@ -157,6 +159,21 @@ def post(self, request, *args, **kwargs):
             return super().post(request, *args, **kwargs)
 
 
+class SendStandingsEmailsView(RoundMixin, AdministratorMixin, PostOnlyRedirectView):
+
+    def post(self, request, *args, **kwargs):
+        active_teams = Team.objects.filter(debateteam__debate__round=self.round)
+
```
czlee: This isn't obvious, and I should really add a docstring about this, but `Team.points_count` is actually query-heavy if it's not prefetched—it'll generate a SQL query per team. There's a function for this purpose called `populate_win_counts` in participants/prefetch.py, so if you add to the imports
``` python
from participants.prefetch import populate_win_counts
```
and then after line 165, add
``` python
populate_win_counts(active_teams)
```
then it'll save a lot of SQL queries when sending this emails.

czlee:
Oh don't worry about HTML messages for now, unless you're dying to do them, they're super low priority (like I care more about the infrastructure for abstracting and logging notifications), I just wanted to leave the door open for them!

tabbycat/tournaments/templates/round_advance_check.html
```diff
@@ -13,6 +13,18 @@
...
+  <form action="{% roundurl 'tournament-send-standings-emails' %}" method="POST">
+    {% csrf_token %}
+    <button class="btn btn-outline-primary active" id="sendPointsEmails" type="submit">
+      Email Team
+      {% if pref.teams_in_debate == "two" %}
+        Wins/Losses
+      {% else %}
+        Points
+      {% endif %}
+    </button>
```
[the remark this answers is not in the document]
tienne-B: 好的，翻译挺麻烦。
C'est vrai, la traduction est assez épineux.

Noticed I didn't even translate it at all... I do have experience with MediaWiki though, where i18n/l10n is a priority :)

While I'm here, might as well put translate tags everywhere.

tabbycat/options/preferences.py
```diff
@@ -1294,3 +1294,34 @@ class BallotEmailMessageBody(LongStringPreference):
...
+    name = 'team_points_email_subject'
+    default = "Your current number of wins for <TEAM> (<TOURN>): <POINTS>"
+
+
+@tournament_preferences_registry.register
+class PointsEmailMessageBody(LongStringPreference):
+    help_text = _("The message body for emails sent to speakers with their team points. "
+                  "Use '<TOURN>' as a placeholder for the tournament, '<POINTS>' for the team points, "
+                  "'<USER>' for the speaker, and '<TEAM>' for the associated team.")
+    verbose_name = _("Team points subject line")
```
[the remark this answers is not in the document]
tienne-B: That was already fixed.

tabbycat/tournaments/views.py
tienne-B: Commit to avoid repeating this I see: 38c6473 :)

tienne-B:
"Failed building wheel for PyYAML" - don't think my code's the problem for travis-ci
(and the completely arbitrary "failing" codeclimate, but that's another thing)

tabbycat/tournaments/templates/round_advance_check.html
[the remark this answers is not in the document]
czlee: Haha, yeah, we've got most of it now but marking translatable string is still a work in progress.

tabbycat/tournaments/templates/round_advance_check.html
```diff
@@ -1,26 +1,25 @@
 {% extends "base.html" %}
-{% load debate_tags %}
+{% load i18n debate_tags %}
 
-{% block page-title %}Confirm Round Advance{% endblock %}
+{% block page-title %}{% trans "Confirm Round Advance" as page_title %}{% endblock %}
```
czlee: If you use the `{% trans string as name %}` syntax, the text actually doesn't render—instead, it just saves it as a local variable. So you'll find on this page the title is blank.

tabbycat/tournaments/templates/round_advance_check.html
```diff
@@ -1,26 +1,25 @@
...
 {% block head-title %}
-  <span class="emoji">🙏</span>Confirm Round Advance
+  <span class="emoji">🙏</span>{{ page_title }}
```
czlee: Just to make things even more confusing, this doesn't work either. Variables defined outside a `block` can't be used inside it: https://docs.djangoproject.com/en/2.0/ref/templates/language/#template-inheritance

It might seem like it violates DRY but I think it's actually fine to have `{% trans "Confirm Round Advance" %}` in there twice, as you might notice we often do. It's just UI text, and the `gettext` tool picks up on duplicate strings and lists them in the translation files only once, unless you provide a `context` to differentiate between them.

tabbycat/tournaments/templates/round_advance_check.html
```diff
@@ -45,31 +44,32 @@
   {% if increment_ok %}
     <div class="alert alert-success">
       <i data-feather="alert-circle"></i>
-      It looks like all the ballots from {{ round.name }} are confirmed, you should be good to go.
+      {% blocktrans trimmed %}It looks like all the ballots from {{ round.name }} are confirmed, you should be good to go.{% endblocktrans %}
```
czlee: Tags with dots in them aren't allowed inside `blocktrans` tags, you need to use with `with` syntax to defined variables: `{% blocktrans trimmed with round_name=round.name %}`

czlee:
I've fixed up all the minor translation things as an example in 49f298b02639da9ce12cc92659ba6176a0a27864, thanks for picking up on this!

tienne-B:
Thanks for the review and showing the wrinkles in blocktrans

PR #672, opened 2018-07-01
Make corrections to new 'Scaling' documentation

tienne-B:
There were a few typos and phrasing problems that are fixed through this commit. I might have also tweaked some punctuation as well if it sounded better to me.

What more could be done is to explain what are these R-series errors.


philipbelesky:
Great, thanks 🙏

PR #677, opened 2018-07-04
Bug fix: Add switch to save ignored field

tienne-B:
Didn't notice that my patch broke the public feedback submission form.

If there is no `ignored` field, then it would not be present in the cleaned data, causing an error in the public submission form when attempting to save the feedback.


czlee:
This is right! `hasattr` checks for the existence of an attribute, not its value.

PR #685, opened 2018-07-14
Add signup page, and controls for it

### LINKED ISSUE #502: Add more user-friendly way to add user accounts
Now that, after five years, we have the luxury to be able to think about such niceties…

What I'm guessing we _probably_ want is a single form that bulk-invites many users, using (ideally) just:
- e-mail address
- whether they should have a tab director account or an assistant account

The user wouldn't be created upon invitation. Instead, it would e-mail each of them a link to a sign-up form, which creates an account upon submission. But in our case, they shouldn't be allowed to sign up _unless_ they've received an invitation. This means the signup form needs to be controlled with access tokens, and you shouldn't be able to access the signup form without it.

The other use-case is the one we assumed would be best back in 2014, before we started setting up sites a few days before the tournament: where you're physically next to the tab assistants and just have them create accounts while logged into your account, one by one. This workflow would basically be the same as how you create accounts in Django admin, with a less intimidating UI.

Django's built-in authentication system doesn't support this. There are probably third-party apps that do, but I think our use-case is a little niche, so the apps don't seem to cover exactly what we want. For example:
- https://github.com/pennersr/django-allauth/
- https://github.com/bee-keeper/django-invitations/

tfpk:
Commit to fix #502


tabbycat/options/preferences.py
```diff
@@ -1334,3 +1335,34 @@ class PointsEmailLinkText(StringPreference):
     section = email
     name = 'team_points_email_link_text'
     default = "To consult the current team standings, visit:"
+
+# ==============================================================================
+account_creation = Section('accounts', verbose_name=_('Account Creation'))
+# ==============================================================================
+
+
```
tienne-B: Adding and blanking the fields are a bit awkward; could there be a toggle "activate" switch or a way to automatically close the keys?

tabbycat/options/templates/preferences_index.html
```diff
@@ -49,6 +49,18 @@
       {% trans "How adjudicators are ranked and who can submit feedback" as description %}
       {% include "components/item-action.html" with type="primary" emoji="🙅" text=title extra=description %}
 
+      {% if show_leagues %}
+        {% tournamenturl 'options-tournament-section' section='league_options' as url %}
+        {% trans "League Options" as title %}
+        {% trans "Special options for running large multi-venue multi-level debating competitions. Warning: these features are not fully tested or supported." as description %}
+        {% include "components/item-action.html" with type="primary" emoji="🏫" text=title extra=description %}
+      {% endif %}
```
tienne-B: Why is this block being shown as added? It is in `develop`.

tabbycat/tournaments/templates/tournament_index.html
```diff
@@ -16,6 +16,16 @@
     {% include "components/item-info.html" with type="info" icon="heart" %}
     <hr>
   {% endif %}
+  {% if global_preferences.accounts__admin_account_key or global_preferences.accounts__assistant_account_key %}
+    {% tournamenturl 'options-tournament-section' section='accounts' as url %}
+      <p class="card-text">
```
tienne-B: `<p` tags are not needed (and go against the style)

tabbycat/tournaments/templates/tournament_index.html
```diff
@@ -16,6 +16,16 @@
...
+    {% blocktrans trimmed asvar text%}
+      <a href="{{ url }}">You have enabled a key to create an account. You should disable it when you can, especially once the tournament starts!</a>
+    {% endblocktrans %}
+    {% include "components/item-alert.html" with type="info" icon="heart" %}
```
tienne-B: `item-alert.html` does not have either a `type` or `icon`. Might be confounding it with `"components/alert.html" with type="info"` (where the `asvar` would have to be `message`)

tabbycat/options/templates/preferences_index.html
philipbelesky: The 'league' class of preferences were a commissioned set of features for running a large-scale high school league tournament; however they deviate significantly from how 'standard' tournaments work and have some incompatibilities with the standard preferences. They are also at this stage not fully tested with some changes in the `2.X` branch. For this reason we have basically hidden them behind a flag so that normal users don't come across them.

tabbycat/options/preferences.py
philipbelesky: Yea, I think this needs some form of toggle to enable. By default with a blank key the signup page is enabled to users at `/accounts/signup/%20/`

philipbelesky:
Hey thanks for adding this @tfpk! I made a number of minor fixes in d6b925cc5135313ade652fb572d58423975d0b88 — the extended commit message has some details specifying them. 

A couple of things to resolve before merging this:

- [ ] @czlee are you OK with the current setup of having a new `users` app and having a new `users` preference page? I could see the case for consolidating the preference page but I think we should probably be moving in the direction of having more (but more explicitly-define) preference pages rather than big grab-bags of semi-related preferences.
    - [ ] If we are splitting off the users app it would be a great chance to move some of the views/templates related to user sign up and login into this new app
- [ ] Currently the default state of each password is blank; which essentially means sign ups are open at `/accounts/signup/%20/`. It looks like there should be an explicit 'enable sign ups using key' style preference for both types, and that the view/form checks this is enabled. IIRC correctly there isn't an obvious way in dynamic preferences to have a string value that can also be false. A similar mechanism is implemented for the `PublicPassword` preference  
- [ ] I'm fine with the URL just being noted on the preference page but am open to other ideas as to where it can go
- [ ] It would be nice, but not essential, if there was an addition to the docs mentioning this feature

tabbycat/options/templates/preferences_index.html
tfpk: It's being shown as added because I moved it from one side to the other - it made more sense in terms of UI balance if we ever enable it.

tabbycat/options/preferences.py
tfpk: Damn I should have noticed that - I'll add a toggle

tabbycat/options/templates/preferences_index.html
tienne-B: So it's duplicated on line 95.

tfpk: Oops - that's an accidental artefact of the manual merge I had to do. I'll fix it in a commit in the next few hours.

philipbelesky:
Hey @tfpk sorry for the delay in getting back to you. I'll give @czlee a chance to get back to my question above otherwise I'm happy to merge this and tweak as needed in future!

czlee:
Actually, almost next. The next thing is a cherry-pick of removed code for EUDC, but this is second, after that 😊

czlee:
Re preference pages—I think our preferences structure is now mostly broken anyway (it's outgrown itself, again!), so I'm pretty indifferent as far as this PR goes. The way it's done now is fine, in the short term I agree that I'd rather have more, shorter preference pages than fewer, longer ones.

Correct me if I'm wrong, but currently the signup page renders for _any_ key, and only checks that the key is correct upon `POST`? I feel like it should just give a 404 if the key is bad? (I'm happy just to merge and then fix if that would be easier.)

Otherwise this looks good! Before this hits production I'd like it if we implemented a mechanism to make the link automatically expire after some amount of time (say, 24 or 48 hours), but that's nontrivial to implement, so we can merge and then do later 🙂

Thanks @tfpk, and sorry again it took me so long to get to!

tfpk:
I agree that this should 404 - that was a poor choice on my part but I'll fix it now.

I don't think the link should expire - I think that there's little reward for a lot of work. The use case for this seems to be that sign-ups would be open from when the instance is created until round one (because more tab staff are added, or someone gets sent an email and forgets until 3 days before... etc...).

czlee:
Neither of you are worried about users forgetting to turn it off?

tfpk:
Isn't there already a prominent warning? Or did I not merge something in?

czlee:
Hmm, I see it in the files changed, probably I just missed it when trying it out. Let me figure out why…

Partly it's that I didn't open the tournament index page when testing it, I think.

tabbycat/templates/components/item-action.html
```diff
@@ -3,18 +3,18 @@
   <a href="{{ url }}" {% if blank %}target="_blank"{% endif %} {% if id %}id="{{ id }}"{% endif %}
    class="list-group-item list-group-item-action
    {% if to_complete %}
-    bg-primary text-white
+    bg-{% if type %}{{ type }}{% else %}primary{% endif %} text-white
```
czlee: just pinging @philipbelesky in case you'd object to this (and change below in the same file)

philipbelesky: Hmm, I think the intent of the `to_complete` flag was to standardise on the primary colour as the indicator for 'key todo action'. Not fundamentally opposed to having an override to that if its used sparingly

philipbelesky:
On every admin page would be slightly overkill and in many cases probably end up having unfortunate side effects on the layout. I'd prefer just on the dashboard — if we wanted to be extra conservative it could perhaps also be a few other critical pages in the flow?

tabbycat/templates/components/item-action.html
czlee: Of course. In this particular case it's security implications that I'd argue warrant the override; I can't think of any others right now, but in my head the threshold is something like "OH MY GOD THIS IS WEIRD AND DANGEROUS YOU DEFINITELY WANT TO CHECK THIS FIRST". (Not just, "You should do this!")

If you want to restructure to _e.g._ make it a different argument (to prevent mishaps or misuse) I'd be fine with that too.

czlee:
I realized by "every admin page" I meant to exclude pages like adj allocations where the layout's really delicate. But for most pages it would just go at the top of the page like any other alert, no?

That said, since this is something that users will leave enabled for some period (and have to put up with the warnings for that whole time), I'd also be okay with putting it just on a few other major pages, which in my head would be the main pages for availability, draw, display, results, feedback, simple importer and the main options index. I'm trying to cover all the pages a user would make a beeline for. Of course, hopefully the user won't even get to the draw page without fixing this, but the purpose of the alert is as much to prevent accidental enabling of the URL (owing to whatever mishap) as it is to remind the user to disable it.

tfpk:
What's remaining to do on this? Just the UI changes?

philipbelesky:
@tfpk I *think* so, do you mean the placement of the alerts/warnings?.

tfpk:
Yeah that's what I was thinking - but was there anything else?

philipbelesky:
@tfpk sorry I missed your last message here. I've lost track of this feature/issue a little bit but I think just adding the warnings is what is needed, although @czlee may have different thoughts. If you can push up those when you have the time I can have a fresh look at the branch and see if there is anything else.

tienne-B:
@tfpk: Are you still interested in this feature?

We have used your central idea (of using global preferences) since 2.4, and to un-stall this PR, you could add your new preferences to the Global Preferences section, reducing this PR.

tfpk:
Hi! Yeah I'd be happy to take another look.
I keep meaning to come back and add more to tabby, so I'll do some work on this this week!

tabbycat/options/preferences.py
tienne-B: I've had a change in mind about this; I'm fine with blanking/adding strings rather than having toggles. @philipbelesky , blank keys are treated as `False` and so would give permission denied now.

philipbelesky:
Thanks for continuing with this feature @tienne-B. I've made a couple of small tweaks to some text. Looks ready to merge unless there was anything else you wanted to add?

tienne-B:
Don't think so

philipbelesky:
Ah, I actually just realised I couldn't push the commit as I don't have access to @tfpk's branch. Could you cherry-pick 5ffe658345f8cd6cc104994352d384650c934b29 into it and merge?

PR #689, opened 2018-07-15
Fix bugs in Notification app

tienne-B:
Two bugs are fixed in this patch:
- The database link in the 'feedback' email form had incorrect capitalization.
- Reply-to and List-Unsubscribe headers had to be optional.

Oddly, in Chrome and Safari, the reply-to fields seem to be mandatory, even though I had not specified that. Is it a browser problem, and could it be overridden without DB editing?

### DIFF
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -20,7 +20,6 @@ class TournamentEmailMessage(mail.EmailMessage):
         self.from_email = "%s <%s>" % (self.tournament.short_name, settings.DEFAULT_FROM_EMAIL)
         self.reply_to = None
         if self.tournament.pref('reply_to_address') != "":
-            headers['List-Unsubscribe'] = "<mailto:%s?subject=unsubscribe>" % (self.tournament.pref('reply_to_address'))
             self.reply_to = ["%s <%s>" % (self.tournament.pref('reply_to_name'), self.tournament.pref('reply_to_address'))]
 
         super().__init__(subject, body, self.from_email, self.emails, bcc, connection, attachments,
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -15,14 +15,16 @@ class TournamentEmailMessage(mail.EmailMessage):
         self.round = round
 
         self.event = event
-
-        self.reply_to = "%s <%s>" % (self.tournament.pref('reply_to_name'), self.tournament.pref('reply_to_address'))
-        self.from_email = "%s <%s>" % (self.tournament.short_name, settings.DEFAULT_FROM_EMAIL)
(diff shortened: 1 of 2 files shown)


philipbelesky:
Do the list-unsubscribe headers actually work? If so is that handled by the mail provider?

I haven't checked but I think the mandatory-ness of the fields should be universal across all browsers — iirc a string preference needs *some* value in order for the form to submit properly.

tienne-B:
What it does is just a convienience link to send a message with a "Unsubscribe" subject to the reply-to address. It would be necessary to do DB modification to remove the address to prevent Tabbycat to send back to the address.

There are a few things that could be done to improve it. The unsubscribe link could be to a new "person" option to disable email. Or to a new page that just removes their email from the DB (using private URLs as verification)

philipbelesky:
Hmm, yea, I'd be somewhat-wary of allowing some sort of automatic opt-out process — that's the only location where the email is stored and tab directors who do send them do so for pretty critical information.

tienne-B:
Removed the header in the new commit. It is however true that we would want to find a way to "unsubscribe" yet keep the email in the DB for the tab director. Hope it makes this pull-request acceptable.

tienne-B:
About that, would it make sense to have a page with the participants and their email addresses for the Tab Director (so that they do not have to go to the DB area)?

czlee:
Umm, good question, I'm actually not sure. I've never felt like I've needed one before, though—I don't even check the email addresses on the private URLs page just before I hit "send". If I'm changing an email address it's normally of a single participant, in which case DB area works fine.

tienne-B:
I was replying to:

> (@philipbelesky) Hmm, yea, I'd be somewhat-wary of allowing some sort of automatic opt-out process — that's [`Person` model] the only location where the email is stored and tab directors who do send them do so for pretty critical information.

With that, my response was not about the modification of the email of a participant; as you say the DB area works fine. I'm pondering a place to show all the participants' email addresses so that the tab director(s) may know the addresses to be able to send non-form messages more easily. This use is implied, but going to the DB area to retrieve an address is somewhat laborious.

philipbelesky:
Hmm, yea, in most cases you imported the emails for everyone (either one by one or via the command line) from a spreadsheet in the first place, so could go back there if you ever needed them in bulk. I guess a page listing them all wouldn't hurt though. The natural place for it would seem to be as another tab in the participants area; I assume basically the same page as `
/admin/participants/list/` but with emails instead of teams/institutions.

czlee:
What would be the use case, though? If it's for copying and pasting into MailChimp, it needs to be in a copy-pastable format, preferably plain text with comma or tab separators. If it's for sending individual emails, I've never had to do that…

Issue #684, opened 2018-07-11
Participant check-in could be done by the individuals


tienne-B:
This may also be a contentious feature, but having participants check themselves in could be added. With the live views of check-ins (even with the "assistant" view which is also for display), instructions for self-check-ins could be shown, with officials overriding and complementing this self-system. Of course there would be settings to enable this feature.

There are two methods I thought of for it:
* **Inbound email**: Participants will be requested to send a message (say `!checkin`) to the sendgrid email address (which does take inbound messages). Then, the linked participant to the email gets their status updated live (or with a reply). Un-linked addresses could get a reply indicating its failure.
* **Private URLs**: A page for each participant/team with a simple "Check-in" button is created using their related private URLs. Pressing it updates (toggles?) their status.

Both these methods would take the timeframe into account for rejections.

philipbelesky:
Hi @tienne-B. I think the 'private URL' method would be most fitting with how user interaction is handled across Tabbycat. Inbound email sounds interesting although ideally it would be nice if features were not overly attached to specific addon providers like sendgrid.

One note for the private URLs thing, I'm not sure if a POST request can be issued from an email (or if so it might be blocked for cross site scripting reasons). But that could probably be avoided by having the check-in private URL just be a page with a check-in button that then POSTs. As a side benefit that page could probably confirm if a user has been checked in (and when) and potentially allow them the ability to revoke a check-in if you think that would be useful.

tienne-B:
No, `POST` requests cannot be done through email. That is why the link would have a "confirm" button, in addition to specifying the applicable timeframe.

Should individual debaters have their own URL key (being aligned with the current framework)? Then, the key could be associated with the `participants` model; not in the `adjudicator` subclass. This would potentially be a separate issue, but there could be a merging of the checkin IDs with the URL IDs (switching to alphanumeric barcodes). However, there could be concerns about data security; exposing the participants' "private key"s at the registration desk. Thoughts?

philipbelesky:
Hmm, I think using a URL key in a similar fashion to what we do for eballots/efeedback would be best. Part of the reason we chose just numeric rather than alphanumeric barcodes is that this speeds manual entry, particularly on mobile devices, so it would be great not to need to compromise this. And while unlikely, it would probably be relatively easy (if desired) to check in other people as their barcode is currently shown to the public in the UI (and on the barcode itself often)

tienne-B:
[...]

PR #690, opened 2018-07-18
Give all participants a private URL

tienne-B:
Sorry for this huge PR, couldn't neatly divide the commit due to DB migrations. (Also mostly worked on it in the plane...) The `Person` table now has a field for `Tournament` as well as for the private URL.

This commit does the side-discussion from #684, and checkins will be added in another commit (separated that at least :) )

Check the commit message for more details.


tienne-B:
Is there a particular way to update the `fixtures`? Doing a `loaddata`/`dumpdata` gives a somewhat different form than what was there.

philipbelesky:
The method I use is:

'python manage.py loaddata data/fixtures/completed_demo.json'

[do migrations and stuff]

'python manage.py dumpdata --natural-foreign --natural-primary -e options -e sessions -e contenttypes --format=json -o data/fixtures/completed_demo.json'

czlee:
Just commenting for clarity that my thoughts are in #684 (where they seem to flow better with existing discussion)

tienne-B:
@philipbelesky: May I have more details on your method here? I think you're missing `-e auth.permission` first of all but what JSON formatter do you use?

Split the private URL feedback view in two so that `self.get_object()` works. That was the stumbling block that prompted me to move `Tournament` to `Person`.

czlee:
@tienne-B Actually, I wrote a script that migrates the fixtures, it's called `migrate_fixtures.py` and it's under the `data` directory. The script is based on @philipbelesky's method. It'll migrate to the old commit's schema, load the fixture, then check out the new commit and migrate to the new commit's schema, then replace the fixture. The one important thing I should point out is that you should set up a blank, **unmigrated** database before you start it, and point your local settings file to it.

Let me know if it doesn't work. It auto-detects most things, but it's not perfect, and I'm definitely keen to patch it if it fails.

czlee:
Oh! Fair enough. The `--indent=4` part actually isn't mandatory—JSON is valid if the whole file has not a single newline—but I put it in the script so that git-diffs would show in some sensible manner, since git-diffs run line by line. I'll make a note to add this information to the contributing docs.

czlee:
I was just thinking about the [data migration](https://docs.djangoproject.com/en/2.0/topics/migrations/#data-migrations) for this, and realized it's not obvious. I mean, half of it is obvious—every adjudicator's URL key should be copied to the corresponding `Person`. But there's a uniqueness constraint on URL keys, which means we can't copy every team's URL key to the two or more speakers on the team. Probably the least bad thing we could do is just choose any _one_ speaker on the team to copy the team's URL key to, and leave the other speakers blank?

@tienne-B Do you want to have a crack at the data migration? I'd be happy to do it but wanted to give you first dibs. Here's a (not very analogous) example: [participants/migrations/0016_populate_team_names.py from version 1.4.6](https://github.com/TabbycatDebate/tabbycat/blob/v1.4.6/tabbycat/participants/migrations/0016_populate_team_names.py). (We ran a giant squash in version 2.0 so we don't have many data migrations in our current code base.)

tienne-B:
The migration now moves the data, including to a speaker on each team.

Issue #671, opened 2018-07-01
Send notification to adjudicators on draw release


tienne-B:
Continuation of #652 but within the framework being created with #653. The main details are in the former task, which was closed due to the infeasibility of the proposed method. This task takes the extensible e-mail approach.

tienne-B:
Having #679 merged would be best before starting work of this feature.

philipbelesky:
Sounds good! Is #679 ready — if not let us know when you'd like it reviewed

tienne-B:
Yes, the pull request is ready; additional work would be in future ones.

tienne-B:
Should the email contain

* The adjudication panel (with institutions), or
* Available motions

in addition to the basic "w"s: **w**hen (round) / **w**here (venue) /(against) **w**ho / **w**hich (side)

Also, should adjudicators get a similar notification for them to know where they are judging?

czlee:
Adjudicator notifications are actually probably more useful than debater notifications. Adjudicators sometimes hang around not in GA (but just outside or something) to avoid crowds; debaters can't really do this because then they'll discover the motion after everyone else.

I don't really see any harm in including the adjudicators with institutions and available motions, though institutions should depend on the "show adjudicator institutions" setting and motions on whether motions are released. (They typically won't be at draw release, since the draw is released before motions are in limited prep tournaments.)

tienne-B:
I do have a patch ready for adjudicator notifications, but which does not include institutions for them (does show Team affiliations) nor the motions, as they would come after, and are less relevant for the adjudicators.

tienne-B:
This task is not yet finished, with some more items on the list before completion: (Might add some _au fur et à mesure_)
* [ ] How to deal with emails when un-releasing/re-releasing draws
* [ ] Being able to send similar notifications to speakers for their team's assignment

Emails sent when re-releasing a draw could check for whether the venue is the same before re-sending, as that is the most vital information.

The second point would be less useful (as @czlee points out), however they could still be useful as a reminder if forgot or something.

philipbelesky:
@tienne-B @czlee do you think the unrelease/release issue and speaker assignment notification are enough to prevent this being released?

tienne-B:
The lack of speaker assignment notifications shouldn't prevent release. The
email notice could be updated when they are included.

Less sure for the unrelease/release issue. Deferring it. Don't think it
should be a blocker, but it could be possible to add a sentence like "This
[...]

Issue #653, opened 2018-06-25
Modularize Notifications


tienne-B:
With the new email features and changes coming, it might be important to put all these email functions as a new module (app?) like `printing`. This would allow for more abstraction of the forms and the functions provided. This task is branched off #644, and inspired with @czlee's FB post liked in #652.

For this app (`notifications`?), there are a few main points:
* [x] Be able to set name/reply-to for all emails (#644)
* [x] Insert headers/metadata to emails
* [x] Better mail-merge (rather than the forest of `replace`)
* [x] Keep record of sent messages (à la `privateurlsentmailrecord`) (mentioned in #626)
  * [ ] Avoid duplicate notifications
* [ ] Support alternatives to emails (#652)
* [x] HTML-type emails (#800)
* [x] ~Easy "unsubscribe"~

Point 2 would be for information as to why the message is sent could be attached, along with an opt-out or something (that would get into dynamic_preferences I guess).

For point 3, that table could perhaps be generalized to something like `notifications_log` with a structure:

Field | Type | Notes
------|-----|-----
id | INT PRIMARY
person | INT FOREIGN | `participants_person`
event | VARCHAR (ENUM?) | What type of email was sent (_e.g._ 'ballot_confirmed')
timestamp | DATETIME
method | VARCHAR (ENUM?) | email or other enabled method
round | INT FOREIGN NULL | `tournaments_round` - As most notifications are round-specific, like 'ballot_confirmed'.

Some of the cruft is left off, such as different fields for adj/speaker, nor are the contents of the email in the table as they could be easily re-constructed from the other data (I'd think).

czlee:
This is a good idea, I was thinking about it too. One dumb question before we get too far in though: Does there exist a third-party Django app that does something similar to what we want? This isn't really a debating-specific problem, and we can't be the first people to want to send a bunch of automated template emails at particular user-triggered times and keep records of it.

By analogy—when we were figuring out how to overhaul options the first time we found django-dynamic-preferences, and now we rely on it and it's been really helpful.

I also agree that once we have other email sent mail records, the `PrivateUrlSentMailRecord` model should be removed and merged into a more generic `SentMailRecord` or `NotificationLog` like the one you describe in this table. (Or the third-party app's records, if they exist.)

[...]

PR #698, opened 2018-07-25
Expand the sent messages record

tienne-B:
Continues on the work of #653.

Makes emails easier to send while adding data to each record. Check the commit message.

Options added in anticipation of #671, #684, and the continuation of #653 itself (SMS).


tienne-B:
Should "Splash page" be renamed to "Landing page"? Seems more applicable.

philipbelesky:
@tienne-B yea, I think of splash page as like a page click through before getting to the 'main' website — or at least that's what they used to be referred to as when that sort of thing was popular in web design like 15 years ago. Landing page definitely sounds more accurate!

PR #699, opened 2018-07-25
Add link for participant self-check-in

### LINKED ISSUE #684: Participant check-in could be done by the individuals

tienne-B:
Adds the participant check-in button to their splash page. Fixes #684.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Change Log
 ------------------
 *Release date: TBD*
 
+- Implemented participant self-check-in through the use of their private URLs.
+- Gave all participants to a tournament a private URL key rather than being by team, and added a landing page for the participants using this key.
 - Implemented templated email notifications with ballot submission and round advance with the messages in a new settings panel. Private URL emails are now also customizable.
 - Added a page to the documentation that details how to scale a Tabbycat site that is receiving large amounts of traffic; and another page that documents how to upgrade a Tabbycat site to a new version.
 - Added a number of performance improvements that should help ensure pages load quickly when receiving large amounts of traffic.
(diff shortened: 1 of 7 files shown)


tienne-B:
The difficulty I'm having is that the label of the button does not consistently update after being pressed, nor does the success message appear. I'd appreciate suggestions to resolve that.

philipbelesky:
Hmm, just looking at the code the success message looks like it should be appearing. I'll hopefully be able to pull locally and check soon. In terms of the button updating, do you mean that it updates after it is clicked (to show some loading state) or that it doesn't update to reflect the changed state once someone has been redirected back to that page?

tienne-B:
The button text should change from "Check-in" to "Sign-out."

philipbelesky:
Hey @tienne-B just a quick note that I pushed up two commits here that fix the button text issue (and the underlying issue) as well as a few other fixes. The commit descriptions have a more detailed set of changes — let me know if you had any questions about them. And thanks again for the PR!

tienne-B:
Didn't solve the issue. The messages are shown on the next page the person visits, and the text does not update.

tienne-B:
Please see #701 for the proper fix.

PR #701, opened 2018-07-28
Fix caching problem in personalized landing pages

tienne-B:
The underlying issue was that the page was being cached (as I had thought in #684 [[comment](https://github.com/TabbycatDebate/tabbycat/issues/684#issuecomment-406579786)]). Saw 304s being sent out as the landing page. To fix, I created a new mixin which is like the PublicTournamentPage mixin but without the caching. Caching can't be used well for personalized pages, which PublicTournamentPageMixin wasn't for.

Also fixed a few typos relating to the feature.

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -800,7 +800,7 @@ class AssistantAccess(ChoicePreference):
 
 @tournament_preferences_registry.register
 class CheckInParticipantSubmit(BooleanPreference):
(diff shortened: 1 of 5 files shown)


czlee:
Ah, yes, removing the caching on participant private pages definitely seems like something we want to do.

Caching actually sort of gets in the way more generally, because messages also get caught up in the cache, so we often have to be careful not to redirect to a cached page when showing a message.

PR #706, opened 2018-07-29
Send message to adjudicators on draw release

tienne-B:
This commit adds the ability to send email notifications to adjudicators informing them of their role and in which venue they are assigned to (but not of the affiliations of others in the panel, nor motions even if released. The notification is customizable, and also includes the debate matchup.

Part of #671.

### DIFF
--- a/tabbycat/draw/utils.py
+++ b/tabbycat/draw/utils.py
@@ -1,7 +1,18 @@
-from django.db.models import Count, Q
+import logging
+from smtplib import SMTPException
 
+from django.core.mail import get_connection
+from django.db.models import Count, Q
+from django.template import Template
+from django.utils.translation import gettext as _
+
+from adjallocation.allocation import AdjudicatorAllocation
+from notifications.models import SentMessageRecord
+from notifications.utils import TournamentEmailMessage
 from participants.models import Team
 from tournaments.models import Round
+
+logger = logging.getLogger(__name__)
 
 
 def annotate_npullups(teams, until):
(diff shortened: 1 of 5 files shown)


tabbycat/draw/utils.py
```diff
@@ -22,3 +31,44 @@ def annotate_npullups(teams, until):
...
+
+    def _assemble_panel(adjs):
+        adj_string = []
+        for adj in adjs:
+            adj_string.append("%s (%s)" % (adj.adjudicator.name, adj.get_type_display()))
+
+        return ", ".join(adj_string)
+
+    for debate in draw:
+        adjs = debate.debateadjudicator_set.all()
```
czlee: The method `debate.adjudicators.with_positions()` iterates through the list in order of position (chairs, their panellists, then trainees). It yields tuples `(adj, adjtype)` where `adj` is an `Adjudicator` (not a `DebateAdjudicator`) and `adjtype` is one of the `AdjudicatorAllocation.POSITION_*` values (which are not the same as the `DebateAdjudicator_*` values, because the former uses a different value for chairs who don't have panellists).

`debate.adjudicators` is an `AdjudicatorAllocation` object, which is a class whose purpose is to provide a nicer interface than the `DebateAdjudicator` [related manager](https://docs.djangoproject.com/en/2.0/ref/models/relations/) that `debateadjudicator_set` returns.

tabbycat/draw/utils.py
```diff
@@ -22,3 +31,44 @@ def annotate_npullups(teams, until):
 
     for team in teams:
         team.npullups = npullups_by_team_id.get(team.id, 0)
+
+
+def send_mail_to_adjs(round):
+    tournament = round.tournament
+    draw = round.debate_set.all().prefetch_related('debateadjudicator_set')
```
czlee: We probably want to prefetch `debateadjudicator_set__adjudicator`, since we need adjudicators' names?

tienne-B:
Having some difficulties, as while the panel string is successfully made, the second `for` loop for the adjudicators does not activate.

tienne-B:
Hope the new indications are good.

philipbelesky:
Added some minor touch ups in 695423d, will merge now. It would be great to also update #671 with what further needs to be done to close that (if anything)

PR #712, opened 2018-07-30
Search for duplicate key notifications

tienne-B:
This commit searches for duplicate private URL key notifications through the email recipient and the private key that was sent, rather than just the Person recipient object. This is to pick up and not count emails for when the private URL has changed or their email as duplicates. (Mentioned in #684)


philipbelesky:
Hey sorry for the delay in merging this. I've had a look through the code but am not super familiar with the new email app setup. This all looks sensible though so I'm going to merge in. I assume we will do a full round of testing/QA for the new email apps prior to the new release which will catch any issues.

Issue #676, opened 2018-07-03
Optimise N+1 end points


philipbelesky:
This is an ongoing tracker for identifying/discussing these views, although scout doesn't actually identify many over the 150ms threshold so they could definitely all be ticked off. None are egregious in terms of execution time (except for `/database/participants/adjudicator/`) but it would be nice to improve them:

- [x] `results.views.PublicResultsForRoundView`
- [x] `participants.views.AdjudicatorRecordView` + `participants.views.PublicTeamRecordView`
- [ ] `tournaments.views.TournamentAdminHomeView` +   `tournaments.views.TournamentAssistantHomeView`
- [ ] `adjallocation.views.CreateAutoAllocation`
- [ ] `draw.views.CreateDrawView`
- [ ] `results.views.AdminNewBallotSetView` + `results.views.AdminEditBallotSetView`
- [ ] `adjallocation.views.SaveDebatePanel`
- [ ] `django.contrib.admin.options.changelist_view` for `/database/participants/adjudicator/`
- [ ] `participants.views.UpdateEligibilityEditView`
- [ ] ` /django.contrib.admin.options.change_view` aka `/database/adjfeedback/adjudicatorfeedback/740/change/`
- [ ] `venues.views.AutoAllocateVenuesView`
- [x] `adjfeedback.views.FeedbackByTargetView`
- [x] `adjfeedback.views.FeedbackBySourceView`
- [x] `printing.views.AdminPrintScoresheetsView`
- [x] `adjfeedback.views.PublicAddFeedbackByRandomisedUrlView`

philipbelesky:
Yea, I've been playing with https://github.com/djsutho/django-debug-toolbar-request-history as a means to inspect both POSTS and the AJAX requests which has been working well.

The problems with SavePanel seem to come from the act of returning a new serialised debate means having to recalculate the feedback score for adjs and the liveness score for teams. It should be possible to more tactically construct the new serialisation so that it pulls those values from the object it receives.

czlee:
Re serialisation, you might remember this discussion from here: https://github.com/TabbycatDebate/tabbycat/commit/3545827ae8c08acaae8f8ff03f6c0cf0ba19bca8#commitcomment-28040144

I had the feeling our serialisation framework causes a lot of duplicate and redundant queries, I think we need to think more thoroughly through how to get it to fetch everything we want, and only want we want, efficiently?

philipbelesky:
[...]

PR #715, opened 2018-08-03
Improve queries in adjudicator feedback views

tienne-B:
To improve the performance of views for adjudicator feedback, Django annotations (`Count()`) were used with prefetches for the query-sets which are used by those views. This chips away a bit at #676.


philipbelesky:
Awesome, thanks @tienne-B! Just testing locally (with a big dataset) this takes each page down from several hundred queries to ~10.

PR #716, opened 2018-08-04
Prevent public forms from being cached

tienne-B:
The public forms to add feedback or ballots should not be cached as they are personalized to the submitter, and having them cached could mean success messages cannot be properly displayed. The fix is similar to that of the private URL landing page. (see #701)

Also moved `BallotEmailWithStatusMixin` to the `mixins.py` file as it seems more appropriate of a place for it than in `views.py`.


philipbelesky:
Sounds good! We have probably been a bit to aggressive on the caching — even disregarding the messages issue there isn't much need to cache pages that are only hit by essentially a single user very rarely. Although I guess the personal pages represent most of those cases.

PR #728, opened 2018-08-07
Use more select_related in get_table_by_team

### LINKED ISSUE #727: Implement prefetching in get_table_by_team
Currently the per debate results view employs prefetching. 7e4a21bfcbc49c69cbc05580da9d85cc0b2ee2d9 implements prefetching adjs/institutions which seems to have cut down on the duplicate queries in that view. 

However the per per team view does not employ prefetching as it constructs the list of debates from team scores. I think this should be able to switch to use `debate_set_with_prefetches` if the appropriate filter can be set. 

Relates to #676 but would be good to have in the hotfix as its public-facing.

tienne-B:
Hopefully fixes #727. Candidate for inclusion in hotfixes.
Some fields were mis-categorised, and debate_team__debate was missing, causing more queries to be made.

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -143,10 +143,11 @@ class PublicResultsForRoundView(RoundMixin, PublicTournamentPageMixin, VueTableT
     def get_table_by_team(self):
         teamscores = TeamScore.objects.filter(debate_team__debate__round=self.round,
                 ballot_submission__confirmed=True).prefetch_related(
-                'debate_team__team__speaker_set', 'debate_team__team__institution',
+                'debate_team__team__speaker_set',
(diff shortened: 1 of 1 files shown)


tienne-B:
Sorry. Opened this without testing, as I was prevented due to #729. I think I was looking at the wrong view anyway as I was not getting an increase of queries when trying it after removing Summernote.

On another note, how are queries analyzed in POST views? Don't think the toolbar can do that.

philipbelesky:
There is an [add-on library](https://github.com/djsutho/django-debug-toolbar-request-history) for DDT that allows for analysing AJAX and POSTs — I've just added it to requirements_develop (note that it will also require an extra line in local_settings).

PR #730, opened 2018-08-07
Optimize participant eligibility query

tienne-B:
The `.exists()` method causes a DB hit, and so it is replaced by finding the category in the list of the speaker's categories, which should be more efficient.

#676 
* [x] `participants.views.UpdateEligibilityEditView`

### DIFF
--- a/tabbycat/participants/views.py
+++ b/tabbycat/participants/views.py
@@ -407,7 +407,7 @@ class UpdateEligibilityEditView(LogActionMixin, AdministratorMixin, TournamentMi
 
     def set_category_eligibility(self, speaker, sent_status):
         category_id = sent_status['type']
-        marked_eligible = speaker.categories.filter(pk=category_id).exists()
+        marked_eligible = category_id in [c.id for c in speaker.categories.all()]
         if sent_status['checked'] and not marked_eligible:
             speaker.categories.add(category_id)
             speaker.save()


czlee:
Hmm, I can see how this reduces the N+1 queries, but I feel like there's still an unnecessary O(n) inefficiency. Here's the offending call:
``` python
speaker_ids = [int(key) for key in posted_info.keys()]
speakers = Speaker.objects.prefetch_related('categories').in_bulk(speaker_ids)
for speaker_id, speaker in speakers.items():
    self.set_category_eligibility(speaker, posted_info[str(speaker_id)])
```

That method does _many_ things one by one:
``` python
def set_category_eligibility(self, speaker, sent_status):
    category_id = sent_status['type']
    marked_eligible = speaker.categories.filter(pk=category_id).exists()   # <======
    if sent_status['checked'] and not marked_eligible:
        speaker.categories.add(category_id)                                # <======
        speaker.save()
    elif not sent_status['checked'] and marked_eligible:
        speaker.categories.remove(category_id)                             # <======
        speaker.save()
```

But [`RelatedManager.add()`](https://docs.djangoproject.com/en/2.0/ref/models/relations/#django.db.models.fields.related.RelatedManager.add) and [`RelatedManager.remove()`](https://docs.djangoproject.com/en/2.0/ref/models/relations/#django.db.models.fields.related.RelatedManager.remove) both support taking many objects as an argument. Can we ditch `set_category_eligibility()` altogether and just do everything in three db hits: one to understand the current status of all posted speakers, a second one to add missing items, and a third to remove unmarked ones?

There's a similar pattern in [`availability.utils.set_availability_by_id()`](https://github.com/TabbycatDebate/tabbycat/blob/cd4ec60/tabbycat/availability/utils.py#L42-L78), except that instead of using the ~`QuerySet.delete()`~ and ~`QuerySet.bulk_create()`~ methods in lines 70 and 75, in this case we should just use `RelatedManager.add()` and `RelatedManager.delete()` (because we have access to a related manager here). (The `transaction.atomic()` in there is for data integrity, not performance.)

tienne-B:
I'll check that out. I do not think we can use `RelatedManager`s as while they can take many objects as arguments, we can't take advantage of that fact as the arguments would be the `categories` to add to the `speaker`s, not the inverse. But I'll see about improving it.

philipbelesky:
Hey just a heads up that this seems to improve query performance (at least in terms of the number) so I'm going to merge this is. If you'd like to continue a discussion of whether `RelatedManager` works in this context feel free to create a new issue or do so in #730.

PR #731, opened 2018-08-08
A few little fixes

tienne-B:
Expanding on 6f17c57 and e92481a. Should the `adjudicator.adj_core` and `.independent` `AlterField` statements stay?


philipbelesky:
Hmm, I think it would be best practice if the migration excluded the changes to those Boolean fields — I think they are a result of the 2.1 upgrade. And in general it would be nice if migrations in commits were atomic and minimised the scope of their changes. Are you ok to edit and update the pull request? Happy to merge then!

Also thanks for going through and reconciling the emoji comments. I was going to do so later but its a little tedious going through such a huge file!

tienne-B:
I didn't have the patience to go through all that file; just enough to go through your commit and do a little regex find/replace (or remove in this case)

Tell me if I'm stepping on your toes a bit.

tienne-B:
The first of the newest two commits addresses an omission in the `Summernote` workaround from fcfc37a in the tournament setup dialog.

The second changes the generation of barcodes from outputting `PNG`s to `SVG`s. It gives a sharper image (and thus might be better to scan). The change in file-size is negligible in my view, but it is now a bit more readable (not that it matters). Using vectors on paper is a preference of mine :)

tienne-B:
Fixes #727 for good!

philipbelesky:
`npm run watch` should enable auto-recompilation which should make things a little faster. I've been meaning to get proper live hot module reloading going although its difficult given the hybrid django/vue approach we have rather than a fully separated backend/frontend.

I might not be able to review these individually for a day or two. In future could distinct changes be submitted as distinct PRs, even if they are for very small? That might make them easier to review rather than needing to merge a series of distinct changes all at once.

tienne-B:
No problem to put them as separate PRs in the future. Hope related changes
can stay in the same PR; I've bemoaned the fact that PRs can't be dependent
on each other before.

philipbelesky:
@czlee as an fyi new emoji were added as the unicode set released last year was added, so that should explain the migration.

czlee:
Ah, and no migration in e92481ac2c69c55c71c3f08410262240c3c3bee3, okay, cool, thanks a relief 😌

tienne-B:
Strange; I checked the extension that was suggested in the interface, and it claimed to support SVGs.

philipbelesky:
Good point, have cherry picked.

PR #736, opened 2018-08-08
Add average speaker average to Team metrics

### LINKED ISSUE #733: Include average speaks in speaker standings/tab
Most people are quite bad at dividing large numbers by 4, 5, 9 etc. 

It would be useful if alongside (and probably on the left of all other scores), speaker tabs included a speaker's average.

tienne-B:
~~Fixes #733.~~

The average speaker average (ASA) is like the average speaker score (ASS), but it returns values within the tournament's scoring range as it averages the speaker scores between all speakers in all rounds, and not just between all rounds. Would there be a better name?

I kinda prefer this new metric honestly :)

### DIFF
--- a/tabbycat/standings/teams.py
+++ b/tabbycat/standings/teams.py
@@ -165,6 +165,22 @@ class AverageMarginMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
     exclude_forfeits = True
 
 
+class AverageIndividalScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
+    """Metric annotator for total constructive speaker score."""
+    key = "speaks_ind_avg"
+    name = _("average individual score")
+    abbr = _("AIS")
+
+    def get_annotation(self, round=None):
+        annotation_filter = Q(
+            debateteam__teamscore__ballot_submission__confirmed=True,
+            debateteam__debate__round__stage=Round.STAGE_PRELIMINARY,
+            debateteam__teamscore__forfeit=False,
+            debateteam__speakerscore__position__lte=round.tournament.last_substantive_position
+        )
+        return Avg('debateteam__speakerscore__score', filter=annotation_filter)
+
+
 class DrawStrengthMetricAnnotator(BaseMetricAnnotator):
     """Metric annotator for draw strength."""
     key = "draw_strength"
@@ -349,6 +365,7 @@ class TeamStandingsGenerator(BaseStandingsGenerator):
         "wins"          : WinsMetricAnnotator,
         "speaks_sum"    : TotalSpeakerScoreMetricAnnotator,
         "speaks_avg"    : AverageSpeakerScoreMetricAnnotator,
+        "speaks_ind_avg": AverageIndividalScoreMetricAnnotator,
         "speaks_stddev" : SpeakerScoreStandardDeviationMetricAnnotator,
         "draw_strength" : DrawStrengthMetricAnnotator,
         "margin_sum"    : SumMarginMetricAnnotator,
--- a/tabbycat/standings/teams.py
+++ b/tabbycat/standings/teams.py
@@ -124,8 +124,8 @@ class TotalSpeakerScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
 class AverageSpeakerScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
     """Metric annotator for total speaker score."""
     key = "speaks_avg"
-    name = _("average speaker score")
-    abbr = _("ASS")
+    name = _("average total speaker score")
+    abbr = _("ATSS")
     exclude_forfeits = True
 
     function = Avg
@@ -165,11 +165,11 @@ class AverageMarginMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
     exclude_forfeits = True
 
 
-class AverageIndividalScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
+class AverageIndividualScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
     """Metric annotator for total constructive speaker score."""
     key = "speaks_ind_avg"
-    name = _("average individual score")
-    abbr = _("AIS")
+    name = _("average individual speaker score")
+    abbr = _("AISS")
 
     def get_annotation(self, round=None):
         annotation_filter = Q(
@@ -365,7 +365,7 @@ class TeamStandingsGenerator(BaseStandingsGenerator):
         "wins"          : WinsMetricAnnotator,
         "speaks_sum"    : TotalSpeakerScoreMetricAnnotator,
         "speaks_avg"    : AverageSpeakerScoreMetricAnnotator,
-        "speaks_ind_avg": AverageIndividalScoreMetricAnnotator,
+        "speaks_ind_avg": AverageIndividualScoreMetricAnnotator,
         "speaks_stddev" : SpeakerScoreStandardDeviationMetricAnnotator,
         "draw_strength" : DrawStrengthMetricAnnotator,
         "margin_sum"    : SumMarginMetricAnnotator,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Change Log
 ------------------
 *Release date: TBD*
 
+- Added the "average individual speaker score" metric which averages the scores of all substantive speeches by the team within preliminary rounds.
+- Renamed the "average speaker score" to "average total speaker score."
 - Implemented a new server architecture on Heroku that should significantly improve performance under load. If upgrading an existing Heroku instance this requires a few tweaks:
     - Adding the `https://github.com/heroku/heroku-buildpack-nginx.git` build pack under the Settings area of the Heroku Dashboard and positioning it first
     - If your Heroku Stack is not "heroku-16" (noted under that same Settings page) it will need to be set as such using the Heroku CLI and the `heroku stack:set heroku-16 --app APP_NAME` command
--- a/docs/features/standings-rules.rst
+++ b/docs/features/standings-rules.rst
@@ -52,6 +52,12 @@ metrics that will be shown in the team standings, but not used to rank teams.
     - The average total speaker score over all debates the team has had, not
       counting debates where they or their opponents forfeited.
 
+  * - Average individual score
+    - The total substantive speaker score, over all debates the team has had and
+      the number of speakers. Provides an equivalent metric to average speaker
+      score in no-reply formats, but within the substantive speech scoring
+      range.
+
   * - Speaker score standard deviation
     - The standard deviation of total speaker scores over all debates the team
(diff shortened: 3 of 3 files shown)


tienne-B:
#737 should be merged before this.

tienne-B:
Changed it so it does not include reply speeches, as they would throw off the scale.

Another metric that I had in mind was "Lopped average" for the speaker tab, which acts like the trimmed average, but that excludes scores based on a (pre-)set confidence interval rather than top/bottom. As a result, it specifically excludes outliers. However this does seem a bit more specific and not really a widespread metric (I know of only the NSDS that does it--disclosure: I have a position there) Worth committing it here, or keep it "local?"

philipbelesky:
@czlee might have different thoughts, but given the metrics are all optional I'm happy to have extra ones in there even if they are not widely used. For both this average-of-average and the 'lopped' average could you provide both a patch for the docs (i.e. [this page](http://tabbycat.readthedocs.io/en/stable/features/standings-rules.html) that explains how the metric functions?

tienne-B:
The lopped average thing will be (possibly much) later; something like:
```
speakerscore__score__lte=F('average') + 2 * F('stddev'),
speakerscore__score__gte=F('average') - 2 * F('stddev'),
```
must be added to the `filter` for the feature to work, but that raises a `aggregate functions are not allowed in FILTER` error.

czlee:
I'm fine with metrics that are not widely used, but I would want all complex metrics to be used _somewhere_, even if it's only one league, otherwise there are literally endless possibilities of very reasonable metrics that we could dream up, cluttering the list.

Based on the information so far in this thread, this means I'm fine with the "lopped average" metric (maybe with a different name), but not so sure about the "average speaker average" metric. Are there tournaments or leagues that already require the use of this metric?

Also, am I right in believing that this metric is mathematically equivalent to the team score (with similar conditions), or are there edge cases where they'd differ that I've missed? (Other than formats with scored reply speeches, in which this metric wouldn't be sensible.)

tienne-B:
You are right that it is mathematically equivalent to the team score. The advantage is that the metric is within the scoring range, which I find makes it easier to make sense of the score. As #733 says:
> Most people are quite bad at dividing large numbers by 4, 5, 9 etc.

This metric is also something the NSDS uses, but it seems (to me) like a simple metric (easily understandable) that may be preferable in other tournaments as well, rather than the more esoteric lopped average which takes the scores as a normal distribution and applies a confidence interval (95% or 2sigma in NS).

czlee:
I can see why one might like it, for sure, but that's the criterion I was arguing against—I think you can come up with arguments for why most metrics "may be preferable". I just want a guarantee that it'll be used. In this case I think it's fine to add though, if nothing else it could be used in the "extra metrics", in the same way that the average speaker score often is in the speaker standings.

To implement the average excluding outliers (how's that for a name?), would it help to store aggregates in a field using `.annotate()`, and then filter on those? Or is that exactly what you were trying above?

philipbelesky:
Ok, am I right in thinking that for this to merge, we need to:

- [ ] Decide whether to include average speaker average
- [ ] Implement lopped average
- [ ] Document lopped average
- [ ] Consider renaming 'lopped average'

@tienne-B once this is done, or as part of it, did you want to implement a `NSDS` preset? It would be good to have (in code) a record of what that format uses.

For the first item there, is average speaker average used anywhere? 

Also, I'm still a little unclear, even with the docs, what its distinction is to the standard average metric. By "average speakers' average scores for all the constructive speeches the speakers have had" does that mean it, in effect, averages across individual ballots (in formats where there are multiple ballots per-adj) rather than across the 'final' combined ballot? If so it would be clarify this distinction in the docs and note that it only makes sense in formats with per-adj ballots.

tienne-B:
How about "The total substantive speaker score, over all debates the team has had and the number of speakers. Provides an equivalent metric to average speaker score in no-reply formats, but within the substantive speech scoring range."

Why wouldn't it make sense in consensus ballots, or was that a misunderstanding (that I hopefully clarified in my proposed description)?

Code for lopped average might as well be in a different PR (or kept local due as it would be not used often). Having some problems with making it, as I can't get it working for now.

czlee:
I think the difference is just that (in formats without replies) the average speaker average is equal to the average speaker score divided by the number of speakers in a team, right?

The implementation, if I understand it right, averages from the `SpeakerScore` model, not the `SpeakerScoreByAdj` model, so consensus/voting doesn't affect it. On that note, another way to see the distinction between ASA and ASS is that it averages from `SpeakerScore` rather than `TeamScore`.

Is my understanding right? The names are confusing and we should probably think of better ones, but I don't have any ideas.

0zlw:
Just came across this PR as it is a metric that I've been interested in and used to try and interpret scores in the past. I think "Average Score per Speech" or "Average Score per Individual Speech" is what makes most sense intuitively if I interpret the pull request correctly.

The way I use this metric is that taking the Australs 2018 tab - the top two teams teams by score scored 2200.50 and 2180.50 which is then divided by 8 rounds, and then divided by 3.5 (3 speeches + 1 reply) which ends up with a score of 78.59 and 77.88 respectively. 

Assuming that this is just the "Average Speaker Score" divided by the "Number of Speeches", then it produces the same ordering as "Average Speaker Score" so I don't imagine it is currently used anywhere for determining a break, however given its apparent usefulness, could gain adoption if it is an available feature in tabby.

In terms of whether it will get used, if it helps, I actually considered using this metric yesterday instead of Average Speaker Score when announcing the breaking teams at a tournament, but I didn't get time to convert the scores from tabbycat.

I would really like if this score was able to be calculated in formats with replies as well but I understand the difficulty. I think using the calculation "Number of Constructive Speeches" + "Max speaker score for reply speech/Max speaker score for constructive speech" would work for most formats, except where the proportions have discrepancies, i.e if the percentage was different for the min/max/average.

Hope those thoughts help!

czlee:
Yeah, thinking about this some more, I feel like the root problem is in the names of the existing metrics. "Total speaker score" makes sense, but "Average speaker score" really means "Average total speaker score". Then when you have "Average speaker score" (the most intuitive name for the addition in this PR), "Average total speaker score" (what is currently called _average speaker score_) and "Total speaker score" it's just a soup of crazy permutations of words.

But as far as I'm aware, as a sport, we don't really have a proper name for any of these metrics…

@0zlw I found the bit about replies a bit hard to follow. If I understand you right: In (say) Australs, would we want to divide the average total speaker score by 3.5?

tienne-B:
Thanks for your suggestions! I would go with "Average Score per Individual Speech" as it is clearer than my "average of average" naming approach, and it doesn't conflict with the "ASS" acronym ("average speaker score"). I'll make that change of wording.

As for reply speeches, I do not think it would be appropriate to include them. As their scoring ranges can be different to constructive speeches, we would get a bimodal distribution, and averages based on those would not have the same value. One of the benefits of averaging per speech is that the metric is within the scoring range, but including reply speeches can skew that. Inversely, it would be confusing to "transpose" the scoring range of reply speeches onto the constructive scale. On that, the range can differ between these types of score (maybe constructive speeches are between 20 while replies are only between 10) and so just comparing top of the ranges would be inadequate.

czlee:
For what it's worth, unless @philipbelesky objects, I wouldn't be opposed to renaming the existing "average speaker score" to something a bit clearer, if that would help.

czlee:
Yeah, but the ambiguity is whether you sum the scores of a team, or whether you average them. Both "average total speaker score" and "average individual speaker score" are averages across rounds/ballots.

Does "average total speaker score" sound confusing to anyone, particularly alongside "total speaker score"? I'm actually fine with it, mainly because I can't imagine what else you could possibly say.

tienne-B:
Changed the names in the last commit. Don't know why the tests are failing.

philipbelesky:
I think that's related to @czlee 's changes in fa43d4bcb18f04a21279159ce0fdd47ef0b2b5e7

philipbelesky:
Also I'm happy with the names are in the current `HEAD`. If you are also @czlee let's merge this in?

tabbycat/standings/teams.py
```diff
@@ -165,6 +165,22 @@ class AverageMarginMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
     exclude_forfeits = True
 
 
+class AverageIndividalScoreMetricAnnotator(TeamScoreQuerySetMetricAnnotator):
```
czlee: typo (`Individal`)

czlee:
Can I be slightly picky and ask for "average individual _speaker_ score" rather just "average individual score"? People will mostly figure out what it means either way, but I feel like "speaker score" should be treated as an inseparable phrase, to differentiate it from "team points" (even though "team score" isn't really a thing), and also because "individual" means both "relating to a single thing as opposed to an aggregate" (without specifying what the thing is, so the thing needn't necessarily be atomic, just in contrast to any aggregation) and "relating to a person" (which is our intended meaning here).

Other than that, I'm happy with it!

tabbycat/standings/teams.py
tienne-B: Fixed, and renamed.

czlee:
Regarding reply speeches: If the suggestion is to take (for Australs, whose substantive speeches as 70-80 and replies 30-40) the average total speaker score and divide it by 3.5 (not by 4), there wouldn't be any issues with bimodal distributions or skewed averages; your average would be conceptually in the 70-80 range, with replies given half weight (which is natural for Australs).

Nonetheless, I'm still inclined to agree that replies should be excluded, because this complicates both the implementation and what it would take to explain the system to someone (in documentation). You can't just explain it as "divide by 3.5", because it's actually "divide by (_n_ + _a/b_), where _n_ is the number of substantive speeches, _a_ is the maximum reply score and _b_ is the maximum substantive score".

For me to change my mind, someone would have to persuade me that a tournament already uses this system, _de facto_. @0zlw, does yours?

tienne-B:
Sorry, don't know much about Australs, or how people view the weight of reply speeches.

> I think using the calculation "Number of Constructive Speeches" + "Max speaker score for reply speech/Max speaker score for constructive speech"

@0zlw did explain it more mathematically, but I do agree that it is becoming more complicated than it should. Further, if the range of constructive speeches ran 20 points while replies were between 10 points, this simple weighting seems to fail (maybe I'm mistaken though)

0zlw:
I think Average Individual Speaker Score and Average Total Speaker Score makes the most intuitive sense so I think cz's nitpicking makes sense.

@czlee yup so (n + a/b) was the initial way I defined the calculation but I understand that replies somewhat complicates the interpretation of this metric. You'd probably want to include both options and I agree that this would be considered confusing/excess clutter. We don't use it as the standard at the moment and although I'd prefer it with replies I understand if we just leave the PR as is. 

@tienne-B so the simple weighting is fine when the range goes from 20 to 10. For example in world schools this works fine because the min/avg/max/range all reduce proportionally by 0.5 (from 60/70/80/20 to 30/35/40/10). An issue would only come about when the 4 numbers dictating the scoring system are out of proportion for some reason (from 60/70/80/20 to 35/37.5/40/5).

Issue #729, opened 2018-08-07
SummernoteWidget causes TypeError in Public Features Configuration


tienne-B:
Probably due to an update, or I wrongly updated something, but when trying to access `admin/options/public_features/` (Public Features Options Panel), there is a type error, claiming `render() got an unexpected keyword argument 'renderer'` here (`tabbycat/templates/components/form-field.html#14`):
https://github.com/TabbycatDebate/tabbycat/blob/d5f7e4afa3c7ea9449822523b552682275f275ba/tabbycat/templates/components/form-field.html#L14
Which leads to `tabbycat/utils/templatetags/add_field_css.py`:
https://github.com/TabbycatDebate/tabbycat/blob/d5f7e4afa3c7ea9449822523b552682275f275ba/tabbycat/utils/templatetags/add_field_css.py#L8

It might be that now, `SummernoteWidget` no longer has a `.as_widget()` method.

philipbelesky:
Closed in fcfc37afa5d63c3e5708bd2a3f54c5c3b81ca0a9

PR #737, opened 2018-08-08
Add 'renderer' argument to MultiField

tienne-B:
Needed for Django 2.1 compatibility. Similar issue to fcfc37a as reported in #729.


philipbelesky:
Thanks!

tienne-B:
Happy international cat day!

Issue #377, opened 2016-09-03
Allow for different names of sides


czlee:
In some styles (Asians, CP) it's convention for teams to be called "Government" and "Opposition" rather than "Affirmative" and "Negative". It'd be nice to make this an option.

This is largely an abstraction exercise, I guess, though we'd want to make sure there's a good caching system for this—views shouldn't have to hit the database just to figure out what to call teams, especially since these words tend to show up a lot.

We'd also need to fix standard abbreviations, like "gov". It's not always the first three letters ("prop" for "proposition". This means that the preference can't act as the text strings itself—it'd need to point somewhere else that has the information. Perhaps it'd be convenient to access these via `Tournament.side_names` or something, I'm not sure.

philipbelesky:
As an aside: maybe using the localisation options would be a good (if somewhat roundabout) way to go about this. For example we could have English (Asians Format) or English (Canadian Formats) that swaps out the terms (and potentially any other lingo).

Or we could just use recreate/use whatever methods localisation uses (assuming they are fast).

czlee:
Hmm, I don't think localisation should be used for this. I think we should save localisation for actual translation, as in, if we were to start supporting entire tournaments in Spanish; I think it'd be confusing if we attempted to use localisation for both.

(Also, I do actually think we might one day support other languages.)

czlee:
Funnily enough, this is a nightmare to work with translation, lol. It's fine when words are standalone, but it gets troublesome when you're trying to build strings like `"Negative's total score"` from `"Negative"` and `"total score"`, or `"Government's division rank"` or similar that we currently construct by piecing together `"

There are a couple of ways to go about this:

1. **Construct strings by formatting things together**.

    This is what I've done currently, for example:
    
        "%(side_possessive)s %(metric_name)s"
             side_possessive would be "affirmative's", "negative's", "opposition's", etc.
             metric_name would be "total speaker score", "who-beat-whom", "rank", etc.
             In English, the two are just concatenated, as shown.

        In French, this would be
        "%(metric_name)s %(side_possessive)s"
             side_possessive: "de l'affirmatif", "du négatif", "de l'opposition"
             metric name: "score totale d'orateur", "qui-a-battu-qui", "classement"
             Then you'd concatenate them, but the other way round.

    Note that the following strategy would _not_ work for French:

        "%(side_name)'s %(metric_name)"
              side_name: "affirmative", "negative", "opposition"

    because you can't hard-code how possession is indicated.

[...]

PR #738, opened 2018-08-09
Clarify speaker tab label

tienne-B:
As there is a different tab for reply speeches (without division classifiers), it makes more sense to use "Substantive speeches" (abbreviated to "Substantives") to describe that kind of tab.

Brings me to another question. How is (grammatical) gender dealt with in strings? Because if "All" is kept, there would have to be two different strings for that word in French: "toutes" for teams and "tous" for speakers.

### DIFF
--- a/tabbycat/standings/templates/standings_menu.html
+++ b/tabbycat/standings/templates/standings_menu.html
@@ -33,9 +33,22 @@
   <div class="btn btn-outline-primary disabled">{% trans "Speakers" %}</div>
   <a class="btn btn-outline-primary {% if standings_speaker_nav %}active{% endif %}"
      href="{% roundurl 'standings-speaker' current_round %}">
-    {% trans "All" %}
+    {% if pref.reply_scores_enabled %}
+      {% trans "Substantives" %}
+    {% else %}
+      {% trans "All" %}
+    {% endif %}
   </a>
+  {% if pref.reply_scores_enabled %}
+    <a class="btn btn-outline-primary {% if standings_reply_nav %}active{% endif %}"
+       href="{% roundurl 'standings-reply' current_round %}">
+      {% trans "Replies" %}
+    </a>
+  {% endif %}
   {% for category in tournament.speakercategory_set.all %}
+    {% if forloop.first and pref.reply_scores_enabled %}
+      <div class="btn btn-outline-primary disabled">{% trans "Categories" %}</div>
+    {% endif %}
     <a class="btn btn-outline-primary"
        href="{% roundurl 'standings-speaker-category' round=current_round category=category.slug %}">
       {% blocktrans trimmed with category_name=category.name %}
@@ -43,12 +56,6 @@
       {% endblocktrans %}
     </a>
   {% endfor %}
-  {% if pref.reply_scores_enabled %}
-    <a class="btn btn-outline-primary {% if standings_reply_nav %}active{% endif %}"
-       href="{% roundurl 'standings-reply' current_round %}">
-      {% trans "Replies" %}
-    </a>
-  {% endif %}
 </div>
 
 <div class="btn-group">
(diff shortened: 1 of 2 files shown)


philipbelesky:
The 'All' here is designed to contrast with the category-specific pages that follow it — for instance 'novice' or 'ESL'. It is a bit awkwardly phrased, but I think calling it just 'Substantives' would remove the clarity between what distinguishes that page and the following categories. 

We could perhaps check if `speakercategory_set` has more than 0 results, and then use a 'All Categories' label; but use 'Substantives' in the case where there are none? That said 'Substantives' is a little awkward in categories without replies, such as BP, where there really isn't a distinction between substantive and non substantive.

tienne-B:
Am I using the correct terminology though? I've always called them "Constructive speeches." I'll make the check.

philipbelesky:
Its probably a regional difference — I think substantive is more common in areas that do 3vs3 in high school whereas constructive seems more america centric?

Also note that this page is also linked to from the admin sidebar and has its own titles — whatever name it should have should be consistent across those. I'm kind of leaning on just keeping the 'all' (perhaps all speakers + all teams?) — if we use substantive/constructive thats going to be less familiar to one particular region. This has also been what it has been called for many many years know so any update to the text might be more clear semantically but less clear for usability.

tienne-B:
What the new patch does (`db7`...) is uses "Substantives" if there are reply speeches, and "All" otherwise. Categories are moved to be at the end, with an indicator "Categories."

czlee:
For what it's worth, I think it's implicitly understood that "speaker tab" means "substantive speeches" unless otherwise indicated. Or, to go into more detail:

In university debating, "substantive"/"reply" vs "constructive"/"rebuttal" is a regional difference. It's also not just terminology—Asia and Oceania think of reply speeches a little differently to how North America thinks of rebuttal speeches. They're close enough to be analogous, but they're not the same. Also, as far as I'm aware, neither CP nor APDA score rebuttal speeches, so our reply speech feature isn't relevant to them. In Asians, Australs, Australian Easters and all New Zealand tournaments, it's quite well-understood that the "speaker tab" relates to substantive speeches—long before Tabbycat even existed.

My understanding is that WSDC is similar to Asia and Oceania on this, in terminology, purpose and scoring. But I'm not so familiar with the schools circuits.

czlee:
Re grammatical gender, great question, @tienne-B—I've spent a lot of time trying to wrap my head around this (see #377).

In this particular case it's quite straightforward: We can provide a contextual marker, and `gettext` then considers two strings with different contexts to be distinct, and translated separately in the translation files. So it'd be like this:
```
{% trans "All" context "all [speakers]" %}
...
{% trans "All" context "all [teams]" %}
```

See: https://docs.djangoproject.com/en/2.1/topics/i18n/translation/#contextual-markers

In other contexts, where they're not conveniently standalone strings, this is another reason why DRY doesn't apply to human-readable strings, and more precisely why we try to avoid assembling sentences using concatenation where the word being substituted is one of a short list known ahead of time. In a purely hypothetical example: rather than `"All %(type)s"` with `type` being `"Speakers"` or `"Teams"`, you would enter `"All Speakers"` and `"All Teams"` as distinct strings.

philipbelesky:
Ok, I've pushed up 474262eb30d294dbcb52b4a81e943f727516648b as some tweaks. Having played around with different formats, options I've thought worked best (for both menu and menu bar)

- For both teams and speaker standings, if there are no replies and no categories just use "Speakers" and "Teams" in the page menu and Speakers/Teams in the side menu.
- For both teams and speaker standings, if there are categories, use "All" in the page menu and "All Speakers" / "All Teams" in the side menu. I've removed the categories prefix because we are somewhat stretched for space here, and the category names are sufficiently obvious on their own. 
- If there are reply speeches, change from the previous option to use "Substantives" in the page menu (where there is a clarifying "Speakers" prefix) but retain "Speakers" in the side menu

This is a little ad hoc but I think employs the substantives vs replies distinction when it is needed and in formats where it should be relatively familiar. Let me know what you think and I'm happy to merge this after any further work on the grammatical gender issue.

PR #739, opened 2018-08-09
Convert Motions-Round to a Many-to-many relation

### LINKED ISSUE #645: Make motion->round a many-to-many relation
Some tournaments re-use motions for many rounds, such as to let teams debate both sides of a prepared topic. Due to the one-to-many relation between rounds and motions, the motion statistics do not take advantage of the complete set of data available, splitting it between various rounds. As a model, I propose:

Motion | motion_rounds | round
------- | --------------- | ------
id | id (as required) | _no change_
text | motion_id
reference | round_id
info_slide | seq
flagged

Thoughts?

tienne-B:
As an advanced feature, while rounds can have many motions, this change makes it possible to have a motion be used for many rounds, without the need to create a new Motion object. This is the most useful to prevent fragmentation in motion statistics if a motion is re-used. Closes #645.


tienne-B:
Fixed the problem with BP. It would be useful if there is a completed BP demo (maybe based on `bp88test`) The printed ballots were an oversight that is also fixed now. What are the weird things Joynt and NZ Easters do? Not familiar with anything New Zealand. Thanks for the QA list.

For design, motions with multiple rounds will be listed under the first round it is in (the stacked card arrangement is back). However, they will also have a _purple_ (changed the colour) badge for the other rounds. I couldn't get `regroup` to work with the many-to-many, and so I used a workaround. The order should be correct.

philipbelesky:
Um, briefly, Joynt has pre-announced topics and pre-announced sides. With NZ Easters, neither topics nor sides are determined by the draw — both are selected by the teams debating so the ballots need to capture both bits of data (similar to how Australs captures the selected motion). In practice visually inspecting the printed ballots and checking ballot submission before/after the commit should be sufficient to determine regressions.

I'm happy with how the motion stats looks in 2-team formats, however in a completed BP format the stats page displays another error `Need 5 values to unpack in for loop; got 2.` when rendering `{% for side, advancing, advancing_pc, eliminated, eliminated_pc in motion.counts_by_side %}`

A completed bp demo would be a great fixture to have, and we could use it in `TournamentTestsMixin` to increase coverage. If you were interested we have a management command defined in `simulaterounds.py` that should make this relatively easy to setup.

tienne-B:
And the sequence of the motions is also important. I did keep that field :)

Fixed the BP problem by always choosing the first round the motion appears in (BP doesn't re-use motions so it's safer)

A BP demo will be in a different PR (in some time). It will require some work on associated scripts and tests. Also having a command to simulate elimination rounds could also be useful.

tienne-B:
So I can keep track of my testing:
 - [ ] Australs
    - [ ] Printable Ballots
    - [ ] Online Ballots
    - [ ] Motion Formset
    - [ ] Motion Page
    - [ ] Motion Tab
 - [ ] British Parliamentary
    - [ ] Printable Ballots
    - [ ] Online Ballots
    - [ ] Motion Formset
    - [ ] Motion Page
    - [ ] Motion Tab
 - [ ] NZEasters
    - [ ] Printable Ballots
    - [ ] Online Ballots
    - [ ] Motion Formset
    - [ ] Motion Page
    - [ ] Motion Tab
 - [ ] Joynt
    - [ ] Printable Ballots
    - [ ] Online Ballots
    - [ ] Motion Formset
    - [ ] Motion Page
    - [ ] Motion Tab

By "Inbuilt Motion Display," I take it that you're referring to the form-set to add motions to a round (or is there a place I'm not thinking of?). For the motion tab, would making the bar always go to 100% be acceptable, as now the number of debates for a motion may not add up to the sum of debates in a round. To meet in the middle, the number of debates that used a motion in a specific round could be added as a tooltip or something.

philipbelesky:
By inbuilt motion display I mean the 'display motions' link at the bottom of the Display page - I.e. the view that makes them big (designed to be shown to a general auditorium). 

The reason for the bar not going to 100% is that in formats like australs with motion selection we want to give an indication of what proportion of rooms choose that motion - a 80/20 balance split is less of a concern when only five rooms did the topic. Previously we found these stats are very easily misinterpreted if the quantities involved are not crystal clear. Happy for there to be a conditional or something that reverts to 100% in the instance motions were reused. 

I'm open to remove the need for sequence as long as there is some other method for easily altering the order although that may be challenging. Teams nominate selections/vetoes based on the order (and ballots record it using order) so the interface needs to make it obvious what the order is an allow it to be edited.

czlee:
In case we weren't aware, the current implementation displays multiply-used motions multiple times in motion statistics, and breaks grouping for the round headers in a motion's second appearance. Also, the BP average points calculation is broken; it seems to always display "1.50 points per team", and a couple of quick spot checks shows this is incorrect. I think it's not filtering for side properly, so it's showing the average points per position for all teams who debated the motion, which is (trivially) 1.5.

image

I have design questions about the `seq` field, the importers, and the appropriate solution to the statistics issue, but all of them are dependent on my enquiries in #645. (I realize it's sort of annoying to have discussions both there and here—in this case, given my questions, I think #645 is the better location.)

tabbycat/motions/statistics.py
```diff
@@ -160,25 +139,25 @@ def _prefetch_prelim_motions(self):
...
 
         annotations = {}  # dict of keyword arguments to pass to .annotate()
-        annotations['ndebates'] = Count('round__debate__ballotsubmission', distinct=True)
+        annotations['ndebates'] = Count('ballotsubmission', distinct=True)
 
         annotations.update({'%s_average' % side: Avg(
-            'round__debate__ballotsubmission__teamscore__points',
-            filter=Q(round__debate__ballotsubmission__teamscore__debate_team__side=side),
+            'ballotsubmission__teamscore__points',
+            filter=Q(rounds__debate__ballotsubmission__teamscore__debate_team__side=side),
```
czlee: So, this gets a little complicated. You noticed below that you needed to drop the `round__debate__` from the filter keyword arguments in order to get it to filter for the right things. Turns out the same is true in this line, where—at first glance—it should read
``` python
filter=Q(ballotsubmission__teamscore__debate_team__side=side)
```

You might then wonder, why did we go through that roundabout `round___debate__ballotsubmission__` path, when we could just look relevant ballots up directly from `ballotsubmission__`? The reason is that, in formats where everyone in a round does the same motion, we didn't want the lack of record of motion in a ballot to inhibit motion statistics. That is: if we know that everyone in a round is doing the same motion, we really just want statistics from the _round_, not the motion. So the motion statistics page looks for statistics for the round.

At this point the filters in the annotations can interact with the main filter up on line 141, which can make things tricky. It's also not entirely clear to me what behavior we'd even want—are we continuing the assumption that there is a one-to-one relationship between motion and round, in the BP code? If so, I think using `rounds__debate__ballotsubmission__` everywhere works?

tienne-B: That goes with the difference between motion-round statistics and motion statistics. I've fixed it so that motion-round statistics take precedence, and pure motion statistics take from the whole set though. Fixed (and seems like the code is simplified a bit)

tienne-B:
The re-configured stats page will come soon.

tienne-B:
Not ready yet; but not forgotten :)

tienne-B:
The branch cleanly merges, don't know why Github doesn't notice.

tienne-B:
What we could try to improve performance in motion statistics is to annotate vetos in a separate query, and just combine the values after. This way, `DISTINCT` clauses should be able to be removed, which are the performance hit.

tienne-B:
Split the statistics queries to remove the need for `DISTINCT`.

czlee (commented):
Looks good!

### Blocking errors

At http://[site]/bp88team/admin/motions/round/2/edit/
```
  File "/home/czlee/tabbycat/tabbycat/motions/views.py", line 72, in get_formset_queryset
    return self.round.motion_set.order_by('roundmotions__seq')
[...]

Exception Type: FieldError at /bp88team/admin/motions/round/2/edit/
Exception Value: Cannot resolve keyword 'roundmotions' into field. Choices are: ballotsubmission, debateteammotionpreference, id, info_slide, reference, roundmotion, rounds, text, tournament, tournament_id
```

At http://[site]/bp88team/admin/motions/round/7/display/
```
  File "/home/czlee/tabbycat/tabbycat/motions/views.py", line 179, in get_context_data
    kwargs['motions'] = self.round.roundmotions_set.select_related('motion').order_by('seq')

Exception Type: AttributeError at /bp88team/admin/motions/round/7/display/
Exception Value: 'Round' object has no attribute 'roundmotions_set'
```

Att http://[site]/bp88team/admin/printing/round/3/scoresheets/
```
  File "/home/czlee/tabbycat/tabbycat/printing/views.py", line 216, in get_context_data
    motions = self.round.roundmotions_set.order_by('seq').select_related('motion')

Exception Type: AttributeError at /bp88team/admin/printing/round/3/scoresheets/
Exception Value: 'Round' object has no attribute 'roundmotions_set'
```

I suspect these are all a hangover from the change in https://github.com/TabbycatDebate/tabbycat/pull/739/commits/db8e1ec99766d86f04e7cf9ec161f36bf2dea71f?

### Non-blocking issues
**(i.e. can be resolved post-merge, but once merged, these should be resolved before the next release)**

- The denominator in the position results bar of the motion statistics page is broken:
   image
   (I checked that this does not happen on `develop`.)

- I think we want to retain a UI distinction between formats that expect there to be one motion that everyone does in a round (like BP), and formats that don't operate like this. It doesn't really need to do much, like it should mostly just flash conspicuous warnings when the data input "looks wrong", to prevent human error—like, the software "watches out for you" when you're under pressure. Designing this for the new schema might be a bit of a project, so probably best left for a new PR.
    image

- The two columns on the Configuration page are currently uneven, we could move "Global Settings" to the right-hand column.
    image

czlee:
Confirming that the three blocking issues have been resolved, as well as the Configuration page columns.

czlee:
> I think we want to retain a UI distinction between formats that expect there to be one motion that everyone does in a round (like BP), and formats that don't operate like this. It doesn't really need to do much, like it should mostly just flash conspicuous warnings when the data input "looks wrong", to prevent human error—like, the software "watches out for you" when you're under pressure. Designing this for the new schema might be a bit of a project, so probably best left for a new PR.

Also, I think this is just a caching issue, so ignore this. Can't imagine what would've caused it, though…

czlee:
http://[site]/bp88team/admin/motions/round/1/copy/ still breaks with the same issue, though.
```
  File "/home/czlee/tabbycat/tabbycat/motions/views.py", line 107, in get_formset_queryset
    return self.round.roundmotions_set.all()

Exception Type: AttributeError at /bp88team/admin/motions/round/1/copy/
Exception Value: 'Round' object has no attribute 'roundmotions_set'
```

For what it's worth, I did a quick `git grep`:
```
$ git grep roundmotions_set
motions/templates/motions_edit.html:25:        {% if round.roundmotions_set.all.length == 0 %}
motions/views.py:107:        return self.round.roundmotions_set.all()
motions/views.py:130:        self.round.roundmotions_set.all().delete()
motions/views.py:131:        motions = self.round.prev.roundmotions_set.select_related('motion')
results/forms.py:250:        self.motions = self.debate.round.roundmotions_set.order_by('seq').select_related('motion')
```

For the rest of us for future reference, `round.motion_set` still exists, right? Any thoughts on which to use? I assume it's just, use `roundmotion_set` if you need any `RoundMotion`-specific information like the `seq` number, and use `motion_set` if it's more convenient to go for motions directly?

tabbycat/motions/models.py
```diff
@@ -51,3 +47,25 @@ class Meta:
...
+        verbose_name=_("motion"))
+    round = models.ForeignKey('tournaments.Round', models.CASCADE,
+        verbose_name=_("round"))
+
+    seq = models.IntegerField(default=1,
+        verbose_name=_("sequence number"),
+        help_text=_("The order in which motions are displayed"))
+
+    class Meta:
+        ordering = ('motion', 'seq')
```
czlee, line 65: The `seq` is with respect to each round, not each motion, right? Wouldn't we want to order by `('round', 'seq')` rather than within motion?

tienne-B: Thought about it myself.

tienne-B:
Thought I had done a search, but realized I was doing a case-sensitive search! All's fixed (and result forms too).

czlee:
Merged to try and put code through its paces as much as possible before Ocicat. With such a major refactor there are almost certainly lots of spots that we've missed, so we should expect more regressions to surface between now and then. The motion statistics bars need to be fixed before Ocicat—I created #1703 to make sure we don't lose it.

I have a question about the "copy" button. To me the word "copy" implies that a copy is made, i.e., that there is a new copy such that if you change the new one, it doesn't affect the old one. That's not true in our case, right? It's more like a "reuse", which is the word used in the tournament configuration? We want to make sure the "Edit Motions" interface makes clear what's going on when a motion is used in multiple places—I know it's technically an "advanced feature" but it will probably get confusing very quickly when someone accidentally or experimentally enables it.

PR #742, opened 2018-08-10
Remove localization on attribute numbers

tienne-B:
When numbers are being used as attributes in HTML tags, they should not be localized, as HTML can only deal with one. This is apparent in the motion statistics.


philipbelesky:
Good catch, thanks!

PR #745, opened 2018-08-11
Fix links in draw overview page

tienne-B:
`current_round` should not necessarily be used, and the motions link should direct to the motions page, not the draw.


philipbelesky:
Not sure how we didn't notice that one before, thanks!

czlee:
Thanks again for picking up on this! Just thought I should mention that this made me realize there was a deeper problem with this setup—we have options for both "all draws" and "current draw only" and we weren't properly distinguishing between them. I fixed this in fa43d4bcb.

PR #750, opened 2018-08-13
Bump Summernote version

tienne-B:
0.8.8.7 does not exist (it was skipped). 0.8.8.8 exists and fixes the incompatibility with Django 2.1. This commit also removes our workaround for the widget.


philipbelesky:
Great, thanks!

Issue #740, opened 2018-08-10
Implement Celery (and use for email notifications)


philipbelesky:
This is one for @tienne-B that hopefully ties in to the existing work on the email app but also adds some infrastructure that can be useful elsewhere.

For context, #590 (has more details) we want the ability to run longer tasks on workers. This is both to escape the 30-second dyno limit and to free up dyno resources that long-running tasks consume. My understanding is that its standard practice to offload the actual emailing process to a worker in this same fashion, so that could be a good starting point for this wider aim/infrastructure.

If you checkout the `celery` branch I've started to set this up although I was having some import path issues. If you were looking something to do getting celery up and going (and using it to implement the email sending) would be super useful. You could then have a crack at implementing it in other areas (say adj allocation) or we could take over at that point.

tienne-B:
Have 12 commits waiting (and #730 in study), guess I could have more :)

Should I continue on `celery` (which is not on `develop`), or cherry-pick to be on `develop`? The notification tasks would be in the `notifications` app (and not their respective applications) IMO, but `eudc` branched off before the app was created.

philipbelesky:
Um feel free to abandon/restart the `celery` branch — I didn't get very far with it. Or feel free to merge develop in to it. The EUDC branch is tournament specific so will be deleted without being merged back in at some point (it exist just to cherry pick specific commits from develop).

tienne-B:
I'm getting a lot of action time-outs with `.delay()`...
```
2018-08-13 03:07:31,876 - WARNING - server - Application instance <Task pending coro=<AsgiHandler.__call__() running at /tabbycat-cloned/venv/lib/python3.6/site-packages/channels/http.py:190> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/futures.py:403, <TaskWakeupMethWrapper object at 0x115034af8>()]>> for connection <WebRequest at 0x115296978 method=POST uri=/demo/admin/privateurls/email/ clientproto=HTTP/1.1> took too long to shut down and was killed.
```
I'll try with some other methods.

philipbelesky:
Hmmm, if you push up your branch I could have a look at it? Is this just when running with `runserver` or is a seperate celery worker also running?

tienne-B:
It's as that PR. It's with running with `runserver`.

philipbelesky:
[...]

PR #751, opened 2018-08-14
Add inheritance to Email factories

tienne-B:
Part of #740. Should not be merged right now. The private URL messages can be tested; but I haven't gotten them to work quite yet (but fixed all the serialization stuff).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,8 +28,9 @@ dj-database-url==0.5.0                  # To obtain the Heroku service's databas
 honcho==1.0.1                           # Allows a Procfile to run multiple processes
 gunicorn==19.9.0                        # Primary server for wsgi
 
-# Cache
+# Cache and Queue
 django_redis==4.9.0                     # Use redis for cache (on heroku; local optional)
+celery==4.2.1                           # Task queue (on heroku; local optional)
 
 # Channels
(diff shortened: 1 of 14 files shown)


tienne-B:
Well, doesn't seem like it was Celery's problem; doesn't work without it either :)

tienne-B:
Without the new classes, serialization is the main problem, as it seems only querysets can be serialized, not objects on its own or the dictionary generator of `DebateResult`.

philipbelesky:
Hmm what format are you serialising to? We have a few custom serialisers as model methods for creating JSON, would a similar approach work if the standard dictionary conversion is not working?

tienne-B:
For the ballot receipts, I tried with both Celery's serializer (which gave me a `'generator` is not serializable`) and Django's one. Where are TC's custom serializers?

philipbelesky:
There are a bunch in `partcipants/models` if you search for `serialize`. Some of these could probably use standard serializers but they make working with the resulting JSON easier. Not sure if these are applicable in this case, I'll try and look into the changes to see what kind of serializer Celery wants if I have the time.

tienne-B:
Getting nowhere. I'd be happy if @philipbelesky finds something but I'll close the PR for now.

Issue #744, opened 2018-08-11
Finish translating strings in Vue templates


tienne-B:
Strings provided by Vue templates do not seem to be translatable, of which is an omission of the i18n efforts. For consistency, these strings should use `gettext`.

philipbelesky:
There is the capacity to translate strings in Vue — checkout the `DiversityContainer.vue` and the corresponding `djangojs.po` files in the language folders for an example. Note that as per the contributing doc these need to be updated manually. 

At present we've focused only on the public-facing templates but intend to do the later ones at some point. Feel free to start adding them — we could use this issue as a tracker?

tienne-B:
This can be a tracker task. Thanks for telling me the method! I'll be adding some templates strings soon.

philipbelesky:
Yea I definitely looked in to more automated methods - the problem becomes that we want to be able to draw from the mo/po files for this (and transifex) whereas the Vue/is native approaches tend to use distinct formats that don't play nice with that.

Shall we track things file by file perhaps? Up to you. Just a heads up that the adj allocation templates will be redeveloped over the next few months so it might be worth skipping those and I'll add the translations as I refactor.

tienne-B:
By update manually, I assume it is to add entries like:
```
#: templates/tables/CheckboxTables.vue:47
msgid "Copy From Check-Ins"
msgstr ""
```
to `en/.../djangojs.po`, then run the make/compile commands. I don't understand what you mean by "Shall we track things file by file perhaps? Up to you." Could you clarify how it would work (hope it isn't that you pass through the file to add the individual strings). I'll skip adjudicator allocations, but I've already translated all the available strings to French :)

tienne-B:
For reference (without any priority assigned):
- [ ] ~~`AdjAllocation`~~ (ignored)
  - [ ] `EditAjudicatorsContainer.vue`
- [ ] `Availability`
  - [ ] `CheckInScanContainer.vue`
  - [ ] `CheckInStatusContainer.vue`
  - [ ] `PeopleStatusMixin.vue`
  - [ ] `VenuesStatusMixin.vue`
- [ ] `Divisions`
  - [ ] `AllocateDivisionsContainer.vue`
  - [ ] `DivisionDroppable.vue`
- [ ] `Draw`
  - [ ] `EditMatchupsContainer.vue`
- [ ] `Printing`
  - [ ] `PrintableBallotHeader.vue`
  - [ ] `PrintableDebateInfo.vue`
[...]

Issue #513, opened 2017-10-28
Time zone support


czlee:
At the moment the time zone is locked to Melbourne, which means all timestamps are Melbourne time. This should probably be configurable somehow. The easy way is to make it a deploy-to-Heroku environment variable, but maybe we could do a user-friendly way too.

This is subject to the same "do you follow the user or the tournament?" problem that translation support does, I guess.

philipbelesky:
The Django recommendation seems to be to store datetimes in UTC and then [translate that to the session's local timezone when the time is presented in the front-end](https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/). 

Looking through the templates we don't present times very often, or accept them as inputs very often, so this probably wouldn't be a huge effort to implement. 

That would probably be my preference as opposed to adding a user configuration. I guess it would look slightly odd for people watching the tournament form afar to see times presented in their own timezone format, but in many ways that would be helpful from an observation point of view.

philipbelesky:
Also is this in consideration for 2.0? Am I right in thinking that all front-end representations of times are incorrect?

czlee:
Yes, you're right. Most times we show are relative times ("2 hours ago") so it doesn't matter. But those times shown as absolute times are always in Australian Eastern time, including in the Edit Database area. The exception is the start time of a round, which is not time zone-aware.

Websites for sports events (like the RWC or Olympics) often give users the option to present times in either the user's time zone, or the event's time zone. I guess this is the ideal, but might be overkill for our purposes. I don't think Django's recommendation works as well for sports events, but for our purposes it's probably fine, I think.

czlee:
Notes:

1. First, a problem. Django doesn't actually automatically detect the end user's time zone; unlike with languages, [there's no easy way to do so](https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/#selecting-the-current-time-zone):

    > The current time zone is the equivalent of the current locale for translations. However, there’s no equivalent of the Accept-Language HTTP header that Django could use to determine the user’s time zone automatically. Instead, Django provides time zone selection functions. Use them to build the time zone selection logic that makes sense for you.

[...]

PR #752, opened 2018-08-14
Mark strings in HTML files to be translated

tienne-B:
Many strings and files have been forgotten about to be translated. This marks most missed strings to be translated.

Went through a lot of the HTML templates to mark strings for translation.
Some questions about the conventions to translate certain things in the Django way:
* Item separators (_i.e._ commas)
* Variables (saw some {{ var }} being enclosed by translation tags without any extra text)
* Strings in embedded `<script` tags (made an error in `scoresheets_list.html`)
* Are date formats hardcoded (_e.g._ for debate scheduling), and if so, is localization automatically dealt with?

More passes may be needed to pluralize some messages, and Vue templates (#744). The `.po` files are not modified here (the French ones will be in another PR with some translations soon enough though :) )

### DIFF
--- a/tabbycat/adjfeedback/templates/add_feedback.html
+++ b/tabbycat/adjfeedback/templates/add_feedback.html
@@ -1,8 +1,8 @@
 {% extends "feedback_base.html" %}
-{% load debate_tags %}
+{% load debate_tags i18n %}
 {% load static %}
 
-{% block head-title %}Who is the feedback from?{% endblock %}
-{% block page-title %}Enter Feedback{% endblock %}
+{% block head-title %}{% trans "Who is the feedback from?" %}{% endblock %}
+{% block page-title %}{% trans "Enter Feedback" %}{% endblock %}
 
 {% block page-subnav-actions %}{% endblock %}
--- a/tabbycat/adjfeedback/templates/feedback_base.html
+++ b/tabbycat/adjfeedback/templates/feedback_base.html
@@ -17,7 +17,7 @@
     </a>
     <a class="btn btn-outline-primary {% if feedback_important_nav %}active disabled{% endif %}"
        href="{% tournamenturl 'adjfeedback-view-important' %}">
-      Important
+      {% trans "Important" %}
     </a>
     <a class="btn btn-outline-primary {% if feedback_source_nav %}active disabled{% endif %}"
        href="{% tournamenturl 'adjfeedback-view-by-source' %}">
--- a/tabbycat/adjfeedback/templates/feedback_by_source.html
+++ b/tabbycat/adjfeedback/templates/feedback_by_source.html
@@ -1,8 +1,8 @@
 {% extends "feedback_base.html" %}
 {% load debate_tags i18n %}
 
-{% block head-title %}Feedback {{ source_type }} {{ source_name }}{% endblock %}
-{% block page-title %}Feedback {{ source_type }} {{ source_name }}{% endblock %}
+{% block head-title %}{% blocktrans trimmed %}Feedback {{ source_type }} {{ source_name }}{% endblocktrans %}{% endblock %}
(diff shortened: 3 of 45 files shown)


philipbelesky:
Awesome! I might leave the first two bullets for @czlee to answer. 

I think for strings within inline `<script>` tags it will be best to use the standard django approach so that the translated string is directly rendered into the template/source. If the javascript is in an external stylesheet that becomes a little problematic. There shouldn't be too many (if any?) cases where that happens though. One way would be to assign it as an inline variable within the js block then have the js read from that. 

For the date time formats I'm not sure. I think the standard time formatters should be user-language aware. But I'm not sure if they would, for instance, format a day/month output in a month-first or day-first variant. That said we don't distinguish between different forms of english, and may not actually ever display dates in a day-month format anyway. So this might be something we need to check and see.

tienne-B:
Another few things about `gettext`:
* Gender support? - as `participant_person` has a `gender` field, it could be useful in some circumstances
* Multiple pluralizations? When a string has two or more numbers with words that may require pluralization

philipbelesky:
Sorry, are you talking about `gettext` in a javascript context? Could you clarify also what you mean by "A method to see where strings are in dev on the web interface"?

tienne-B:
No, I'm talking about `gettext` generally.

Should have clarified what I meant by that. It would be to have strings "point" to where they are coded, such as on the final round's ballot entry page, by hovering (or something else) over "In the grand-final rounds of a British Parliamentary format this form...", it would indicate `results/templates/enter_results.html:49` to know where to find the string.

czlee:
I'll catch up on old messages eventually, forgive me:

**Item separators:** As far as I know, no good way to do it. I currently assemble a non-translatable comma-separated list, try to reword it to avoid the final "and", and hope for the best. This doesn't always work. Let me know if you figure it out.

**Variables:** If the _only_ thing inside a translated string is a single variable comprising user-defined text, it shouldn't be translated. So, for example, `{% blocktrans %}{{ name }}{% endblocktrans %}` is nonsense. If you see it, please fix it!

**Date formats:** In general, date and time formats are locality-aware. You can actually see this on the admin tournament overview page (I think), action log dates should change when you switch languages.

There were some complications though, see: https://github.com/TabbycatDebate/tabbycat/issues/513#issuecomment-366570036

Details are in there, but regarding what @philipbelesky said about MM/DD and DD/MM: It's because of the American–Commonwealth distinction/ambiguity on this that I [overrode English date formats](https://github.com/TabbycatDebate/tabbycat/blob/3d4614b7fc6565166bc899892f2d06717750a794/tabbycat/utils/formats/en/formats.py) to spell out the month (e.g. "18 Aug"), always. So in English that's already accounted for.

**Gender support:** As far as I'm aware, this isn't supported by `gettext`. Either we have to do the annoying gender-neutral thing, or we have to implement it ourselves. That aside, `Person.gender` is an optional field and most tournaments don't use it, so we shouldn't rely on it anyway, not even kind of. Its main use is for gender-balancing adjudicator panels, which only major tournaments tend to care about. For our application I don't think it's worth going through what [Facebook had to figure out about a decade ago](https://www.cnet.com/news/facebook-gender-policy-has-grammar-in-mind/) and [build into its translation platform](https://www.facebook.com/help/translations/1055862821149956).

**Multiple pluralizations:** Also isn't supported by `gettext`. I think I had to do this once with "_m_ adjudicators and _n_ speakers", you might have seen it when you were refactoring the private URLs emails. If you can, rephrase to avoid, otherwise, implement custom logic.

**Seeing where strings are on the web interface:** Not easily. Transifex offers the feature, but only as a (prohibitively expensive) paid feature.

Sorry I don't have any better news! Gettext is a good system compared to having nothing, but there are definitely times I've wished it was more sophisticated.

PR #757, opened 2018-08-15
Update package.lock

tienne-B:
How the `package-lock.json` file changed for me with the new changes. Is stuff like this to be committed or kept local?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,14 +23,14 @@
         "@babel/traverse": "7.0.0-beta.47",
         "@babel/types": "7.0.0-beta.47",
         "babylon": "7.0.0-beta.47",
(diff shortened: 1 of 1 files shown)


philipbelesky:
So the `package-lock.json` is meant to be committed. However its also meant to be deterministic. I think the difference is probably due to slightly different node/npm versions — could you check you are on versions that roughly match what is specified in `package.json`?

Also note that I'm going to reject/close this as I did an `npm audit fix` in cf47e2ee4864603c822b22ad2e52450c0be0a98d which would have altered the affected file anyway.

tienne-B:
Thanks for the reply; wasn't sure.

PR #758, opened 2018-08-16
Create completed BP fixture

tienne-B:
The fixture uses the bp88team preset as a base, with scores added for the 8 preliminary rounds. Contrary to other fixtures, the option tables had to be included as BP relies (and will fail without) certain non-default settings.

Should one be merged and not the other, as they are both huge files: 4 rounds is 35 000 lines and the whole 8 is 60 000?


philipbelesky:
Hmm, perhaps as a compromise could this be setup to match the existing fixture set — i.e. one set directly after round 1 (so availability/draw stuff can be checked) and one as a completed demo but one with only 4 rounds (just to cut down on the file size). 

With regards to tests, hopefully this should slot in to `TournamentTestsMixin`. It will be great to be able to have those view tests with BP coverage! As noted with things like motions the same view can have quite different code paths.

tienne-B:
Ok, I'll commit that over this PR. For them to slot in `TournamentTestsMixin`, I think the `id`s will have to be shifted, or there would be conflicts in foreign keys.

philipbelesky:
Ah true, I forgot that all fixtures are loaded at once. `TournamentTestsMixin` probably then also needs further modifications so that view tests and the like visit the paths for each tournament separately.

PR #762, opened 2018-08-17
Translate more messages into French

tienne-B:
Including the new Notifications application and assorted strings.

Some bigger changes:
- Feedback (Commentaires) has been renamed to Evaluations
  (Évaluations)
- Test scores (notes d'examen) becomes Base Score (note de base)
- Allocations (attributions) becomes Distributions (répartitions)

Fixed some grammatical errors, but may also have left some of my own. I'm not done with translations though.


tienne-B:
@czlee: Are those bigger changes fine with you? Didn't really like the use of "attribution"

Also wondering about the etymology for this use of "Test" as I wouldn't think judges get quizzed before the tournament. :)

czlee:
They actually do! At major international tournaments, like WUDC, EUDC, Australs and Asians, all judges are required to sit a test before the tournament begins. The test normally consists of some theoretical questions, as well as watching and delivering an adjudication on a test debate. This allows the chief adjudicators to assess the ability of, and assign a score to, all adjudicators in the tournament, especially those with whom they don't have a prior familiarity (which is actually most adjudicators). So the term "test score" comes from the fact that it is, in fact, a score derived from the adjudication test.

That said, while this software was originally written for Australs, you're right to point out that not all tournaments run an adjudication test. Most smaller, local tournaments do not. We kept the term "test score" around because people (who were typically also familiar with the major internationals) knew or could figure out what it meant. But "base score" is more general, so we might consider changing that name in English, too 🙂

Évaluations is definitely fine (as discussed offline, the English word "feedback" is mild misnomer, in my opinion). The translation of "allocation" I also struggled with. We tend to think of this process as "allocating an adjudicator to a debate" and "allocating a venue to a debate". Because of this I sort of looked over "répartitions" (est-ce qu'on "répartit" un juge à un débat? ou je suppose qu'on répartit _les_ juges _aux_ débats?) but this makes sense to me, so thanks for picking a more appropriate word.

One thing: I've noticed that verbs sometimes use different tenses in particular contexts. A common example is "Loading...", which is typically shown as the noun "Chargement..." in French (and similarly for other "in progress" messages that use the gerund in English). So phrases like "Saved 3 motions" might translate differently depending on whether they're a success message or just a historical log, or even "Edit Motions", which I thought might translate differently depending on whether it's on a button, or the title of a page.  We're not very good with context markers at the moment, so feel free to add them if you think they're necessary. Also, correct me if I'm wrong on this!

Thanks again for your work on this!

tienne-B:
Test score just made less sense when there are suggestions to tweak it in response to feedback or other metrics. I also looked at other terms such as "discretionary score" to reflect that fact, but it is too long and wouldn't be as suitable for scores determined by an actual test. Is "Curriculum vitae score"

Oui, il faut utiliser le verbe "répartir" avec des pluriels parce qu'on traite l'ensemble des débats non seulement un à la fois.

The next point would be better in #752, which has some `gettext` questions for you. About context markers generally, in poedit if I wasn't sure, I just looked at the code that used the string. I guess that's a bit harder on transifex or with non-programmer translators.

tabbycat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -368,197 +438,215 @@ msgid ""
...
 
-#: templates/nav/admin_nav.html:176 templates/nav/public_nav.html:162
+#: templates/nav/admin_nav.html:182 templates/nav/public_nav.html:165
 msgid "Standings"
 msgstr "Classement"
 
-#: templates/nav/admin_nav.html:186
+#: templates/nav/admin_nav.html:193
+msgid "All Teams"
+msgstr "Tout les équipes"
```
czlee: C'est "tout les équipes" ou "toutes les équipes?" (Aussi "tout les orateurs", ci-dessous.)

tienne-B: Bien vu! "équipes" est féminin, donc ça devrait être "toutes" comme aux autres places. "Orateur" par contre est masculin avec "tout". Le féminin serait "oratrice". Pour l'uniformité, c'est vrai qu'on devrait utiliser "orateur(ice)" partout. Aurez-vous une préférence de ponctuation/signes double pour montrer les deux genres? Je ferais les modifications plus tard. (Je suis aussi trop paresseux pour utiliser des guillemets).

czlee: Ben, je n'ai pas d'opinion tranchée sur la ponctuation pour indiquer les deux genres. Mais dans ce cas, on m'avais appris qu'au pluriel, sauf que tous les sujets soient féminins, on utilise le masculin, donc "tous les orateurs"?

Dans les cas où on fait référence à une personne particulière, j'ai jusqu'à présent utilisé par exemple `%(adj_name)s est le/la <strong>président(e)</strong> du jury`, parce que c'est ce que j'ai vu le plus souvent, mais auriez-vous une autre préférence?

tienne-B: C'est quand-même possible que tous les oratrices indiquées sont des femmes :)

On peu bien continuer avec les parenthèses. C'est le plus commun mais il y a plein d'autres conventions. Des barres obliques, des points, _etc._

czlee: Ben, oui, vous avez raison, il y a quelques tournois où toutes les oratrices sont des femmes 🙂

Qu'est-ce que c'est la pratique type dans ces cas? Quand l'expression fait référence à plus ou moins tout le monde, et habituellement les participants comportent des hommes et des femmes, mais dans certains cas, il est possible que toutes les participantes soient des femmes—qu'est-ce qu'on dirait?

PR #763, opened 2018-08-17
Some fixes to i18n

tienne-B:
Pluralized and added missing strings, fixed some parameters to translate
and made the HTML tag reflect the current language being used.


philipbelesky:
Thanks, this is a heroic effort on the translation front!

Issue #776, opened 2018-08-21
Change ballots graph to a more appropriate chart-type


tienne-B:
Originally reported in https://github.com/TabbycatDebate/tabbycat/issues/392#issuecomment-414520858, but it seems to be a different task. To re-state, the graph found in the admin's overview page should be a line chart for clarity. The changes in the lines for different statuses (unreceived, checked-in, unconfirmed, confirmed) and the times these statuses change is more easily seen and compared (with the intersection of series) through lines.

> @philipbelesky: Isn't that the function that the ballots status graph provides on the tournament dashboard/home page? (https://github.com/TabbycatDebate/tabbycat/issues/392#issuecomment-414566930)

No, the chart currently being used is a bar chart, which is a chart-type for discrete groupings (which time is not). As such, bars aren't ideal for this use.

philipbelesky:
I'm not sure people who use the graph are more interested in the time that those statuses change than the actual proportion of statuses in each category. The graph is always up to date and intends to answer the critical question of *how many more ballots are there to go right now?* while the general shape of the previous results given an idea of how quickly each type is decreasing or increasing. I'm not sure the exact times matter much. 

But definitely open to the idea — having a properly spaced time axis along the x might make it easier to construct the graph and presuming it clearly displays the proportion of status would provide the benefit of more accurately showing the overall time periods.

czlee:
For what it's worth, we used to have an area chart with a correctly spaced time axis there: https://github.com/TabbycatDebate/tabbycat/blob/77a078ebe962b558d80eb76a8ec2606f3e9a4b8b/tabbycat/templates/graphs/BallotsGraph.vue#L52-L103

We reverted it to a stacked bar chart in b98f0fe6743b475116543fdbaa2a09d3712781af:
https://github.com/TabbycatDebate/tabbycat/blob/b98f0fe6743b475116543fdbaa2a09d3712781af/tabbycat/templates/graphs/BallotsGraph.vue#L52-L127

My understanding was that this was due to a d3 package upgrade, and that we always intended to revert it back to the area chart with a correctly spaced time axis, just with low priority. @philipbelesky, was I mistaken?

Personally I'm definitely interested in understanding the time that those statuses change, which is why I had changed it to an area chart with correctly spaced times in the first place. I want to understand whether our delay is due to a handful of very slow ballots, or whether the ballot progression is roughly even at a slow rate.

czlee:
[...]

Issue #433, opened 2017-02-23
Allow results page's list of ballots to update in place 


philipbelesky:
Basically it occasionally becomes a pain (with large numbers of data enterers) that one has to refresh the results list in order to get an up to date list of what still needs to be entered, confirmed, etc. When lots of data is entered this can lead to people loading old versions of ballots or not realising something needs to be confirmed, not entered.

This should be relatively easily avoided by having the list be populated by a Vue object that it can then update every 10 seconds or so over JSON. If we wanted to be extra careful to avoid degrading performance we could cache the JSON response and manually invalidate it only when a ballot's information changes.

philipbelesky:
Ok this has been merged in. Keep an eye out for odd behaviour although I haven't found any unfixed cases in testing. Notes:

- In addition to being nice and rounding out the websocket infrastructure this should hopefully save a lot of load when lots of people are on that page at once.
- [ ] The JS translations files haven't been added (but the code uses `gettext`) as compiling them still seems broken.
- The advance round button still needs a manual refresh to show. This is a pain but too tricky too fix and maintain the layout. An alert notifies admins of this need to do this.
- Assistants can now no longer click through links to ballots that they are unable to confirm. Hopefully this is a convenience, but not sure if there's an edge case to consider here.
- This was a pain to get working with the ballots graph. I also found that the ballots graph doesn't really work well when the last confirmation is many many hours after the round advanced (I presume this is when a result is re-edited?). This doesn't fix this but shouldn't make it worse either.
- The confirmed timestamp of a ballotsub is now recorded in local time.
- [ ] Would it be possible to remove the inline display of results in the ResultsTable? Its a bit stretched for space in BP formats and I'm not sure its worth the extra cost of the page load.

czlee:
Nice! Looks great and sounds good.

-  Re clicking through to ballots they can't confirm, the edge case would be if they decide they need to discard it? (Should people be able to discard their own inputs?)

- Can you elaborate on the "confirmed timestamp in local time" thing? What was the case before? It's a timezone-aware object, right?

- I quite like having an inline display of results, it's nice to see what's been going on at a glance. I can't say I'm concerned about page load cost (unless I'm missing something?), but I see what you mean about horizontal space. 

    Do you just mean the chevrons? Do they make that much of a difference? Or do you also mean listing the teams?

    If you just meant the chevrons, would you have feelings about presenting it like this?

    | Government teams | Opposition teams |
    |--------|-----------|
[...]

PR #779, opened 2018-08-24
Change ballots graph to area chart

tienne-B:
Got interested in d3, and so tried my hand in doing #776 with the old colours (not check-ins though). It doesn't conflict with the work on the #433 working branch (and anyways, the data construction is separate to the graph so it should be possible to change one without touching the other).

This partially reverts 0555bcc68b39df57bd7f3be905568f33a793ffb2 to have the "square" effect on the chart.


philipbelesky:
#433 will probably see the data construction method change slightly — ideally the websocket updates should be transmitting just the changed ballot rather than reporting a full list. This then puts the onus on the client to figure out the delta to its existing data set. But that should be just as easy to change with this new graph form than it is for the old one. I might also make a few tweaks here as I go but in general this looks great!

czlee:
Just wanted to echo that this looks great, thanks @tienne-B!

PR #787, opened 2018-09-01
Fix a few French typos

tienne-B:
Also changed "réponse" for reply speeches to "réplique" which I have seen used, and has both senses of response and rebuttal.

Should this be cherry-picked for the 2.20 branch I see is created?


czlee:
Ah, thanks! I learnt a new word today 🙃

PR #799, opened 2018-09-04
Make private URL email message in preferences

### LINKED ISSUE #531: Make private URL distribution e-mails customisable

tienne-B:
Closes #531.

This removes the need for the separate page, with a warning list for
the participants where email failed. Now the message is customizable
though the email preferences, and the link to the email form is now a
form button which deactivates if everyone reachable has been sent their
key.

The delta check for if the key has changed does not work and so has been
removed.


philipbelesky:
Hey, thanks for the PR! I haven't been super on top of all the changes to the email notification system but this seems like a great wrap up to that original ticket. As I'm (hopefully) done with the server setup tinkering I'll have a bit more chance to play with this.

PR #805, opened 2018-09-08
Use team codes in emails according to options

### LINKED ISSUE #804: Email notifications should use team code names if applicable
Just making a note that currently the "draw released" emails sent to adjudicators use `debate.matchup`, which (always) uses real names. Ideally we'd want to use code names if code names are enabled for the public, and real names if they're not. (That is, we'd want to use the same names as are used in the public draw.)

Unfortunately there's currently no single convenience function for this _within_ `Debate` or `DebateTeam`, because it depends on a tournament preference. I think we'll need to construct the relevant string from within `draw/utils.py`, or do some refactoring. There _is_, however, a convenience function `use_team_code_names()` in `options/utils.py`, that interprets the preference cleanly, because there are five choices for various permutations.

Relevant bits of code (as of 531faa3056bd3892dcf3138a7bf86814a2666471):

https://github.com/TabbycatDebate/tabbycat/blob/531faa3056bd3892dcf3138a7bf86814a2666471/tabbycat/draw/utils.py#L60-L66

Example of existing code for choosing between code name and real name (this is the code used in the tables constructor):
https://github.com/TabbycatDebate/tabbycat/blob/531faa3056bd3892dcf3138a7bf86814a2666471/tabbycat/utils/tables.py#L197-L213

tienne-B:
The team_codes preference was ignored in emails, always opting for the teams' short names even when the codes are preferred (in the preferences). This adds switches to email generators between codes and short names.

Fixes #804.

### DIFF
--- a/tabbycat/results/utils.py
+++ b/tabbycat/results/utils.py
@@ -8,6 +8,7 @@ from django.template import Template
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy
 
+from adjallocation.models import DebateAdjudicator
 from draw.models import Debate
 from notifications.models import SentMessageRecord
 from notifications.utils import TournamentEmailMessage
@@ -228,17 +229,16 @@ def send_ballot_receipt_emails_to_adjudicators(ballots, debate):
         if 'adjudicator' in ballot:
             judge = ballot['adjudicator']
         else:
-            judge = debate.debateadjudicator_set.get(type="C").adjudicator
+            judge = debate.debateadjudicator_set.get(type=DebateAdjudicator.TYPE_CHAIR).adjudicator
 
         if judge.email is None:
             continue
 
-        scores = ''
+        scores = ""
         for team in ballot['teams']:
-            if use_codes:
-                scores += _("(%(side)s) %(team)s\n") % {'side': team['side'], 'team': team['team'].code_name}
-            else:
-                scores += _("(%(side)s) %(team)s\n") % {'side': team['side'], 'team': team['team'].short_name}
+
+            team_name = team['team'].code_name if use_codes else team['team'].short_name
+            scores += _("(%(side)s) %(team)s\n") % {'side': team['side'], 'team': team_name}
 
             for speaker in team['speakers']:
                 scores += _("- %(debater)s: %(score)s\n") % {'debater': speaker['speaker'], 'score': speaker['score']}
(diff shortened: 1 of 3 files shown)


tabbycat/tournaments/utils.py
```diff
@@ -199,7 +200,7 @@ def send_standings_emails(tournament, teams, request, round):
 
     for team in teams:
         context['POINTS'] = str(team.points_count)
-        context['TEAM'] = team.short_name
+        context['TEAM'] = get_team_short_name(team)
```
czlee: Hmm, does this one need to be coded? This one's only sent to the team to whom the points count relates, right? (I'm open to discussion on this, I just assumed it would be fine with the real name.)

tabbycat/results/utils.py
```diff
@@ -233,7 +234,7 @@ def send_ballot_receipt_emails_to_adjudicators(ballots, debate):
 
         scores = ''
         for team in ballot['teams']:
-            scores += _("(%(side)s) %(team)s\n") % {'side': team['side'], 'team': team['team'].short_name}
+            scores += _("(%(side)s) %(team)s\n") % {'side': team['side'], 'team': get_team_short_name(team['team'])}
```
czlee: I like this abstraction and use of DRY, but it's more complicated than it sounds. Because `team.tournament` is called in `get_team_short_name()`, it requires the `tournament` to be fetched for each team, and then the preference (`tournament.pref('use_team_code_names')`) to be called for each team in this loop, bearing in mind that teach `team.tournament` is a different _Python_ object referring to the same _database_ object, which causes a collection of redundant database and cache lookups.

It doesn't matter so much here, because there are at most four teams, so it's like an N+1 issue where N≤4. But my current practice (until I find a better way—I'm open to suggestions) is to just call `use_team_code_names` outside the loop, and choose inside the loop which name to use, violating DRY if necessary. (In this case, if we decide to decode the names in the team point standings emails below, then DRY is a moot point.)

tabbycat/tournaments/utils.py
tienne-B: Yes, it is only sent to members of the team.

tabbycat/results/utils.py
tienne-B: I see your point.

philipbelesky:
@czlee I'm happy with the new `matchup_codes` method if you are. Although we might want to make a note to see if it can replace some of the other existing cases where we are constructing team lists.

czlee:
I'm also happy with `matchup_codes`.

PR #814, opened 2018-09-09
Add classes for Email Message generators

tienne-B:
In an effort to reduce the amount of duplicated code, some inheritance is added to the execution of send_messages functions that are based on a queryset. This change will allow to mark those functions to be handled through Celery (background workers) in a later change.

### DIFF
--- a/config/requirements_heroku.txt
+++ b/config/requirements_heroku.txt
@@ -10,6 +10,7 @@ gunicorn==19.9.0                        # Primary server for wsgi
 
 # Cache
 django-redis==4.9.0                     # Use redis for cache (on heroku; local optional)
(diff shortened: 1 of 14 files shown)


philipbelesky:
The inheritance patterns here look good. Is this ready to be merged, or are you still working on it?

philipbelesky:
Ok cool, I'll try and find some time to look at the Celery branch also. No problem with Codeclimate — I never really give it more than a cursory glance. It's good to have it around but we don't really use it to vet PRs.

tienne-B:
Your initial Celery setup is included in this PR. Don't think there's much (if anything) different in that branch.

czlee:
Did we accidentally reverse the changes in #805 by merging this in?

PR #815, opened 2018-09-10
Re-factor email generator classes

tienne-B:
Having the classes take in a tournament object was un-necessary and might have been counter-productive for Celery-ifiation. Also removed the select_related from objects as that is just for QuerySets.

(Is the query count good in emails? The request history tab is broken)
I'm getting timeouts when trying Celery; given up a bit on that.


philipbelesky:
I'll make a note to test the query count at some point as part of a broader review of the email changes in 2.3

tienne-B:
Just because the preferences are possibly being checked with each email.

PR #829, opened 2018-09-17
Use different branch for request history toolbar

tienne-B:
The master branch of that was broken when jQuery was removed from the debug-toolbar.


philipbelesky:
Looking at your issue on their page the maintainers seem pretty responsive, would it be easier to just wait until the new version is released? Or is it too annoying to work around until then?

tienne-B:
Possibly, but I found it odd that the patch was placed in a new (and the sole non-main) branch, not `master` with indications to install the new branch. Made me wonder whether it was going to be merged...

PR #834, opened 2018-09-24
Remove deprecated page

tienne-B:
As indicated in a comment on the page: remove after Khao Manee.


tienne-B:
Interesting how there wasn't a merge commit, just you appearing as "committer."

philipbelesky:
Yea when merging on mobile it gave me the option to squash it, although I didn't realise that would change the apparent author. I'll probably stick with the standard merge commit in future unless you have a preference otherwise.

czlee:
Author and committer look fine to me, it says the author is @tienne-B?

PR #835, opened 2018-09-24
Fix public ballots page

tienne-B:
The object being sent was not the debate but the matchup, causing it to
fail.


tienne-B:
"**Actually** _actually_ fix public scoresheets"

PR #836, opened 2018-09-24
Add CheckInEventConsumer to self-checkin

### LINKED ISSUE #832: Use sockets for self checkin
Using `POST` messages for self checkin has no broadcast for the updated status, causing the change to not be reflected in the checkin views. This also has the reverse consequence that changes in status through the checkins sheet will not be reflected in the landing page (if an event happens while the participant has the page loaded). This is less important though. Don't think it is crucial but for the personal page, the events feed should be filtered to only include events relating to the subject.

tienne-B:
So that self-checkin events are live in the admin/assistant views, the self-checkins through private URLs have to send messages through the `CheckInEventConsumer` channel.

While this does not address having the private landing pages receive messages, that is much less important, especially as that event is caught with a "already checked-in" message and the fact that people wouldn't be staying on the page for long.

Closes #832.

### DIFF
--- a/tabbycat/checkins/views.py
+++ b/tabbycat/checkins/views.py
@@ -1,5 +1,7 @@
 import json
 
+from asgiref.sync import async_to_sync
+from channels.layers import get_channel_layer
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.views.generic.base import TemplateView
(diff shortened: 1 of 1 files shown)


philipbelesky:
Sorry for the delay in merging, looks great!

PR #837, opened 2018-09-25
Fix spacing in printable ballots

tienne-B:
There was no spacing between adjudicators and their info. Also made the point happen only at the end of the list, rather than after each adj and got the ampersand only with the last adj. Instead of an ampersand, writing out "and" could be less awkward, but that isn't done in this commit.


philipbelesky:
Well spotted; thanks!

PR #839, opened 2018-09-26
Notification documentation

tienne-B:
New page was added to supplant the help texts in the options. Feel free to suggest better formatting in the documentation.

More details in the commit messages.

### DIFF
--- a/docs/features/notifications.rst
+++ b/docs/features/notifications.rst
@@ -14,31 +14,45 @@ All emails have the ``{{ USER }}`` and ``{{ TOURN }}`` variables to indicate who
 .. list-table::
   :header-rows: 1
 
-  * - Event
-    - Description
+  * - Email content
+    - Description and when sent
     - Variables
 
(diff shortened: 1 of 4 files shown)


czlee:
@tienne-B Thanks! Added a little bit, let me know what you think?

tienne-B:
The formatting stuff I didn't like was the fact that we are on 3 columns, not enough space for the variables to have some room for readability. I like the additions though, but must warn that the lines don't break. If they did, the descriptions would be under the title as the list takes up lots of vertical space which is unused in the other column(s).

czlee:
Oh! They break for me:
image

This isn't a browser thing?

tienne-B:
They break for me now; must have been that I did something wrong when I tried. Also there is less dead space than what I thought there would have been. What do you think of having 2 columns?

czlee:
I actually thought about reducing to two columns and assumed you had a reason for not doing so!

PR #840, opened 2018-09-26
Add QR codes to private URL sheets

### LINKED ISSUE #775: Include QR code in private URL distribution pages
It is a bit odd to hand people a physical item directing them to the internet, but typing a random string is not really user-friendly. QR codes that directs to the URL could be added to the URL sheets for ease of use. The type of barcode that is in use can't encode as long a string, but a QR code is recognized to be scanned to lead to a website and seems to be well supported by phones.

This may require another dependency to generate these codes; `jsBarcode` can't generate them.

tienne-B:
It is awkward to type in a coded URL. To make it easier, QR codes allow for a URL to be encoded in a 2D barcode, redirecting to a URL when scanned with a cell-phone. The QR codes are generated in the client with qrcode, a new python dependency.

As this is done server-side, it will load slower, but PDFs are generated quicker.

Fixes #775.

On another note, thinking about dropping the HTML interface of printable pages in favour of server-side conversion to PDF, making printables automatically download. That would avoid people needing to fiddle with settings or wait for Chrome (also allows other browsers to work as well)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Change Log
 ------------------
 *Release date: ?*
 
+- When printing the per-participant private URLs they now feature QR codes in addition to the URL. Thanks to Étienne Beaulé for contributing this feature!
 - Split up the Django settings files. Note that this means if you are upgrading a local install of Tabbycat to this version you will need to:
     - Copy `tabbycat/settings/local.example` to become `local.py` (and fill in your original database details).
     - Optional: repeat the same copying procedure for `development.example` and set the `LOCAL_DEVELOPMENT` environmental variable to `True` if you would like to use the settings designed to aid local development.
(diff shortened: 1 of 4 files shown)


philipbelesky:
Hey thanks for the PR, I'll get around to checking it out properly in the next day or two. Just a quick question: do you have a sense of how long the server side rendering process takes? Would it be greater than say 25 seconds for a 300-400 team tournament? (the heroku request timeout is 30 seconds)

tienne-B:
Don't have any dataset that large, but it takes 5-6 seconds for 256 people. The end-goal is to have the file directly as a PDF, which could be generated async and easily served once requested.

The goal of bypassing html is for better browser support and to rid ourselves of the printing instructions by having the document ready for printing. Haven't checked it yet, but the vue templates may have to be migrated back to plain HTML for easy printing.

philipbelesky:
This looks great, thanks!

tienne-B:
So the performance was good?

philipbelesky:
Yea, I tried it with a pretty large data set and it only took a couple of seconds. Extrapolating that data set to WUDC-size it should still be well under 15seconds; if not we can either split the page up into teams vs adjudicators or (at worst) shift the QR code generation to a front end library perhaps.

PR #843, opened 2018-09-27
Cast descriptors to string if empty

### LINKED ISSUE #842: Adjudicator popovers fail to generate with no descriptors
The affected code is the 'construct_popover' method (line 543 of utils/tables.py) - When an adjudicator has no descriptors (eg. if they're a solo chair which is unaffiliated with an institution), the descriptors array isn't converted to a string (line 556) and remains an array, which the resultant 'append' method (lines 558-561) can't handle.

tienne-B:
When the descriptors array is empty, it returns false, which keeps it as an array, causing an error in the adjudicators table.

Fixes #842.

### DIFF
--- a/tabbycat/utils/tables.py
+++ b/tabbycat/utils/tables.py
@@ -552,8 +552,7 @@ class TabbycatTableBuilder(BaseTableBuilder):
                     descriptors.append("<span class='text-danger'>in minority</span>")
                 text = a['adj'].name
 
-                if descriptors:
-                    descriptors = "(%s)" % (", ".join(descriptors))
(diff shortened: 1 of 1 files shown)


tienne-B:
@viranweerasekera: If it is blocking a tournament (as the [FB post](https://www.facebook.com/groups/tabbycat.debate/permalink/1085140874987202/) seems to show), you could hot-patch the change.

viranweerasekera:
All good @tienne-B, in the end I advised them to just add an institution to all of the adjudicators, and it's worked for them!

PR #849, opened 2018-09-29
Use channels to send notifications

tienne-B:
Email generators will now only create tuples of (context, user) to be made into emails to be sent off. In this new setup, the action of requesting emails is sent to an asynchronous channel which then calls a generator. This should reduce the number of calls to the database, and make emails non-blocking.

The side-effect is that success/failure messages cannot be sent as easily, and has not been attempted here. Rather, messages notifying that messages *are queued* have been made instead.

This has the same goal as #740.


tienne-B:
Sadly, the consumer does not seem to activate on the `POST` events, and I can't figure out why. It would be helpful to point it out to me :)

czlee:
I'm sure I'm missing something but I'm confused about exactly what the underlying problem is? How many database hits does it make currently when you cause emails to be sent?

czlee:
Oh, I see. This is actually already covered in `Tournament`:

https://github.com/TabbycatDebate/tabbycat/blob/90c70a34b6c5d0b59b6ac8a40c490ca4c774a1a0/tabbycat/tournaments/models.py#L66-L75

So long as you pass the same `Tournament` object into each `TournamentEmailMessage`, it should just consult that `self._prefs` dict. Would that do the trick?

tienne-B:
Didn't know about it! Then there shouldn't be a performance hit here.

Are you opposed to using Channels for notifications?

czlee:
I think I'd just need to be convinced there was a good reason to? So I'm opposed to using channels in cases where there's no reason to use channels, because they're a scarce resource (I think, @philipbelesky will correct me if I'm wrong).

philipbelesky:
@czlee they're not an especially scarce resource — e.g. for HWS maximum client utilisation has been 2  out of 20 over the last 24-hour period. Previously there was an upstream bug that meant clients were not being closed properly which was what was previously causing redis limits to be reached.

In terms of scope it is pretty standard practice for potentially long-running tasks to be moved to an asynchronous worker process; especially in a heroku context where we have the 30 second timeout issue. Remember that channels isn't just a way to run websockets, its also a general purpose  asynchronous layer. There is further discussion of this in #740, but essentially I'm leaning towards using channels, rather than something like celery, to run async tasks given our needs are relatively simple, it doesn't require additional dependencies, and the channels layer works happily in a heroku and local context.

philipbelesky:
Ok have pushed up a working version and will merge in a second. Notes:

- Definitely lead you down the wrong track duplicating the existing pattern, as per the doc above using workers requires a different setup in routing; unlike other channels these are url/protocol-less so are just called by a string identifier
- It also turns out that the communication with the worker doesn't operate with the `InMemoryChannelLayer` so for local development purposes this will require the `USE_REDIS_LOCALLY` flag (and a running copy of redis server). Because local installs probably don't send emails anyway this isn't a problem, but if we are implementing core functionality (i.e. allocations) using a worker then we either need to specify redis as an install dependency (not ideal) or maintain both a worker and non-worker code path to that functionality and switch based on the channel layer backend. 
- As a result of this (and future planned use of async tasks) we now require a worker to run in order to handle those processes; i.e. `dj runworker notifications`. I'll add this on to the `serve` script and the heroku deployment tasks, but this will require a worker to be scaled when upgrading existing instances.

philipbelesky:
(Also just a note that I'm closing as I've manually merged this into `develop` because I was not keen on the web editor)

tienne-B:
Including beb18d1 ?

tienne-B:
Made #863 for that.

PR #850, opened 2018-09-29
Make private URL into links in tables

tienne-B:
To not have to copy-paste URLs from the admin view to the browser, now they are hyperlinks. Also reduced code duplication when building the tables.

Thought the participants' names could be the link, or to only have the "key" part of the URL, but some tournament might rely on copy-pasting the urls, which would be made more difficult otherwise.

### DIFF
--- a/tabbycat/privateurls/views.py
+++ b/tabbycat/privateurls/views.py
@@ -66,7 +66,7 @@ class RandomisedUrlsView(RandomisedUrlsMixin, VueTableTemplateView):
 
         def build_link(person):
             if person.url_key is None:
-                return {'text': _("no key"), 'class': 'text-warning'}
+                return ''
             path = reverse_tournament('privateurls-person-index', self.tournament,
                 kwargs={'url_key': person.url_key})
             return {'text': "🔗", 'link': request.build_absolute_uri(path)}
--- a/tabbycat/checkins/views.py
+++ b/tabbycat/checkins/views.py
@@ -1,5 +1,7 @@
 import json
 
+from asgiref.sync import async_to_sync
+from channels.layers import get_channel_layer
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.views.generic.base import TemplateView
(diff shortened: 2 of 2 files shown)


tienne-B:
Might just be how I use web-pages, but I find links easy enough to copy, right-click and copy. It's better than dragging over, as the formatting and other information could also be picked up as well. I'll see if there's another way, like a "hidden column" or something.

czlee:
How about

ht​tps://my-fun-tournament.herokuapp.com/funfunfun/privateurls/8yqm4wef/ ([link](https://my-fun-tournament.herokuapp.com/funfunfun/privateurls/8yqm4wef/))

tienne-B:
Don't like the emoji?

czlee:
Oh, sorry, I didn't see the updated commit before I wrote that! (I also haven't been trying it live, just reading code.) Emoji works! Though since `utils/tables.py` is quite long I try only to put things in there if they are (or are likely to be in the near future) used in multiple places. If it's a special-purpose column, I either construct the column inline, or delegate it to a special subclass of `TabbycatTableBuilder`, see for example `draw/tables.py`.

tienne-B:
Avoiding embedding HTML in the python. The table builder only likes links to cover the entire cell. But the emoji works in a new column as it's all aligned (unlike using the same cell) and doesn't take much space (problem with long names). Constructing the column as a method of `RandomisedUrlsView` as that is the only place I'd imagine it would be used.

tabbycat/privateurls/views.py
```diff
@@ -66,7 +66,7 @@ def build_url(person):
 
         def build_link(person):
             if person.url_key is None:
-                return {'text': _("no key"), 'class': 'text-warning'}
+                return ''
```
tienne-B: Sadly emojis can't be superimposed; I was thinking an "X" over the link. Guess I forgot the text was there.

tienne-B:
Odd, where's the "-B" from my username in the branch name? :)

tabbycat/privateurls/views.py
czlee: Oh, my preference would still be no text 🙂

czlee:
ha, the remote on my computer skips it because my fingers are too lazy to use capital letters 😛

PR #851, opened 2018-09-30
Don't send ballot receipts on elimination rounds

tienne-B:
Ballot emails are score-oriented, which is not compatible with elimination-type ballots.

Closes BACKEND-ZJ

### DIFF
--- a/tabbycat/results/mixins.py
+++ b/tabbycat/results/mixins.py
@@ -4,6 +4,7 @@ from django.contrib import messages
 from django.utils.translation import gettext as _
 
 from notifications.utils import ballots_email_generator
+from tournaments.models import Round
 from utils.misc import get_ip_address
 
 from .models import Submission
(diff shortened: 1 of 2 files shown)


philipbelesky:
Thanks!

czlee:
Wait, this should only apply to BP, right? Two-team forms just have normal ballots?

tienne-B:
You tell me... Does two-team **elimination** rounds still use scores?

This does not disable email notifications for BP; it disables elimination rounds.

czlee:
Yeah, two-team elimination rounds still use scores. Only BP elimination rounds don't do scores.

We do one day intend to support two-team winner-only ballots, for both preliminary rounds and elimination rounds. But that's for another day.

philipbelesky:
Sorry that totally skipped my mind when merging. I've amended in 5c9aba1e12b069490fced9aeec3fd2ef1dd07421

PR #863, opened 2018-10-01
Get tournament based on debate/round

tienne-B:
The notifications worker looks for a tournament_id, which might not be
present. This change looks for other object ids that can be used to get
the tournament (and round if applicable).


tienne-B:
As we are returning a boolean, we did not have to have `if` statements. Changed the commit to condense the method.

PR #865, opened 2018-10-01
Make flake8 happy

tienne-B:
Fixed places that were being flagged by flake8. Previous PR was from wrong branch.


philipbelesky:
Just as an fyi a [git hook for flake8](http://flake8.pycqa.org/en/latest/user/using-hooks.html) that will prevent commits which trigger it

tienne-B:
You've made the necessary changes.

PR #867, opened 2018-10-02
Implement team information emails

### LINKED ISSUE #791: Email notifications: add speaker to team assignment
Add the ability to email out speakers noting what their team is. It might be good to think about how this could be used in conjunction with team codes.

tienne-B:
Team information emails can be used to double-check make sure teams'
registrations are correct by sending them what was recorded in Tabbycat.
These notifications have information such as the members of the team,
the team name, and their category(ies).

Closes #791.

### DIFF
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -123,9 +123,11 @@ def standings_email_generator(url, round_id):
     teams = round.active_teams.prefetch_related('speaker_set')
     populate_win_counts(teams)
 
-    context = {'TOURN': str(tournament)}
-
-    context['URL'] = url if tournament.pref('public_team_standings') else ""
+    context = {
+        'TOURN': str(tournament),
+        'ROUND': round.name,
+        'URL': url if tournament.pref('public_team_standings') else ""
+    }
 
     for team in teams:
         context_team = context.copy()
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -171,3 +171,31 @@ def motion_release_email_generator(round_id):
             emails.append((context, speaker))
 
     return emails
+
+
+def team_speaker_email_generator(tournament_id):
+    emails = []
+    tournament = Tournament.objects.get(id=tournament_id)
+
+    for team in tournament.team_set.all().prefetch_related('speaker_set', 'break_categories').select_related('division', 'institution'):
+        context = {
+            'TOURN': str(tournament),
+            'SHORT': team.short_name,
+            'LONG': team.long_name,
+            'CODE': team.code_name,
+            'DIVISION': team.division.name,
+            'BREAK': _(", ").join([breakq.name for breakq in team.break_categories.all()]),
+            'SPEAKERS': _(", ").join([p.name for p in team.speaker_set.all()]),
+            'INSTITUTION': str(team.institution),
+            'EMOJI': team.emoji
+        }
+
+        for speaker in team.speakers:
+            if speaker.email is None:
+                continue
+
+            context['USER'] = speaker.name
+
+            emails.append((context, speaker))
+
+    return emails
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -190,7 +190,7 @@ def team_speaker_email_generator(tournament_id):
             'SHORT': team.short_name,
             'LONG': team.long_name,
             'CODE': team.code_name,
-            'DIVISION': team.division.name if not None else "",
+            'DIVISION': team.division.name if team.division is not None else "",
             'BREAK': _(", ").join([breakq.name for breakq in team.break_categories.all()]),
             'SPEAKERS': _(", ").join([p.name for p in team.speaker_set.all()]),
             'INSTITUTION': str(team.institution),
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -46,10 +46,11 @@ def adjudicator_assignment_email_generator(round_id):
             if adj.email is None:
                 continue
 
-            context['USER'] = adj.name
-            context['POSITION'] = adj_position_names[pos]
+            context_user = context.copy()
+            context_user['USER'] = adj.name
+            context_user['POSITION'] = adj_position_names[pos]
 
-            emails.append((context, adj))
+            emails.append((context_user, adj))
 
     return emails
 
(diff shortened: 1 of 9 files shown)


tabbycat/notifications/utils.py
```diff
@@ -171,3 +171,31 @@ def _create_motion_list():
...
+    emails = []
+    tournament = Tournament.objects.get(id=tournament_id)
+
+    for team in tournament.team_set.all().prefetch_related('speaker_set', 'break_categories').select_related('division', 'institution'):
+        context = {
+            'TOURN': str(tournament),
+            'SHORT': team.short_name,
+            'LONG': team.long_name,
+            'CODE': team.code_name,
+            'DIVISION': team.division.name,
```
czlee: This line crashes if `team.division` is `None`.

tabbycat/notifications/utils.py
```diff
@@ -171,3 +171,31 @@ def _create_motion_list():
...
+            'EMOJI': team.emoji
+        }
+
+        for speaker in team.speakers:
+            if speaker.email is None:
+                continue
+
+            context['USER'] = speaker.name
+
+            emails.append((context, speaker))
```
czlee: This line adds a reference to the dict `context`, not a copy (in Python, dicts are always passed by reference). Because you're modifying the same `context` once for each speaker, the `emails` list will have the last speaker in each team, multiple times, because you have the same dict for each speaker in a team.

The dict is different for different teams, because line 181 assigns the whole dictionary. It's just the same for all speakers on a single team.

The method `context.copy()` might help.

czlee:
This is very helpful, thanks! I was just thinking about writing such a feature at HWS IV 🙂

tabbycat/notifications/utils.py
tienne-B: Wouldn't I have seen that it doesn't work with the other email generator functions that use the same pattern (like `motion_release_email_generator` or `standings_email_generator`)?

czlee: When I call `standings_email_generator` from the shell, this is exactly what happens:
``` python
In [2]: standings_email_generator("http://helloworld.org/", 25)
Out[2]:
[({'TOURN': 'emailstest',
   'URL': 'http://helloworld.org/',
   'POINTS': '0',
   'TEAM': 'Team Bravo',
   'USER': 'Andreas Grunewald'},
  <Speaker: Zhu Tsao>),
 ({'TOURN': 'emailstest',
   'URL': 'http://helloworld.org/',
   'POINTS': '0',
   'TEAM': 'Team Bravo',
   'USER': 'Andreas Grunewald'},
  <Speaker: Deodata Trevisani>),
 ({'TOURN': 'emailstest',
   'URL': 'http://helloworld.org/',
   'POINTS': '0',
   'TEAM': 'Team Bravo',
   'USER': 'Andreas Grunewald'},
  <Speaker: Frida Padovesi>),
 ({'TOURN': 'emailstest',
   'URL': 'http://helloworld.org/',
   'POINTS': '0',
   'TEAM': 'Team Bravo',
   'USER': 'Andreas Grunewald'},
  <Speaker: Andreas Grunewald>)]
```

I don't seem to be able to get emails working at all, though—I think something's broken for me with the new channels system. So I couldn't tell you what's in the emails. (They don't get logged, either, so it's like the consumer's not picking up on the message.)

tienne-B: Aie! Guess I have more to fix then! Also, make sure a Redis server is running, and redis is enabled. Then the emails should work (other than this one apparently)

tabbycat/notifications/utils.py
```diff
@@ -126,16 +128,18 @@ def standings_email_generator(url, round_id):
...
+        context_team = context.copy()
+        context_team['POINTS'] = str(team.points_count)
+        context_team['TEAM'] = team.short_name
 
         for speaker in team.speaker_set.all():
             if speaker.email is None:
                 continue
 
-            context['USER'] = speaker.name
+            context_user = context.copy()
```
czlee: did you intend to copy `context_team` here?

tienne-B: Good catch.

tabbycat/notifications/utils.py
```diff
@@ -166,8 +170,38 @@ def _create_motion_list():
...
+    emails = []
+    tournament = Tournament.objects.get(id=tournament_id)
+
+    for team in tournament.team_set.all().prefetch_related('speaker_set', 'break_categories').select_related('division', 'institution'):
+        context = {
+            'TOURN': str(tournament),
+            'SHORT': team.short_name,
+            'LONG': team.long_name,
+            'CODE': team.code_name,
+            'DIVISION': team.division.name if not None else "",
```
czlee: Yeah, this still crashes, because `not None` is always `True` 😛

czlee:
The solution to this isn't obvious, but it worries me a little that the default email text includes code names, regardless of whether code names for the tournament have been enabled. This might throw off some unsuspecting tab director, who might send email registrations only to then have to field hundreds of questions about whether this tournament is using code names.

A tab director who knows about how this works would, of course, know to edit the preferences ahead of time to remove the code name from the email text (and add whatever other details they might want). But there's nothing anywhere reminding you that this is probably a good idea, much less preventing you from emailing code names to everyone when you didn't want to.

With most things I'd still merge on the understanding that we should resolve this before the next release, but with this I feel that even I'd be caught out sometimes. A "quick fix" might just be to make the code name part conditional on the code name setting, but this doesn't play nice with preferences. Another quick fix might be to have an alert somewhere, if the template includes `CODE_NAME` with code names disabled, or doesn't include `CODE_NAME` with code names enabled, just before sending the emails?

czlee:
I'd have put that commit on the develop branch, but it seemed more sensible to avoid merge conflicts knowing that this PR would be coming in soon 🙂

(Apparently there was confusion at some tournament with some people only opening team points emails much later, and thinking it related to a silent round, not the preceding round when it was actually sent.)

tienne-B:
Not including the round as a variable in that email was an oversight; should have been there, especially that it is dependent on the round. Also removed the code name from the default; hoping that tab directors see the list of available variables they may use.

czlee:
Hmm, yeah, that'll do for now. It's easier to send more information than to have to make some broadcast asking people to ignore information, and there's at least some chance (though not guaranteed) that good tab directors will ask themselves about code names before sending the email.

We should make a note to have a better system for warning people before they commit to something potentially stupid, though 😉

Issue #541, opened 2018-01-15
If a team ironed last round, check if it's ironing this round


czlee:
As a redundancy check, it's good to check explicitly whether a team who ironed last round also ironed this round. Ideas for achieving this, we should not necessarily do all of them:

- [X] Present a list of teams who ironed last round somewhere in the UI, linked to from an alert.
- [x] Present an alert on the ballot entry page for their next debate, reminding user to check if it was an iron.
- [x] Write an unmissable warning on the paper ballot asking the chair to mark explicitly whether the team is ironing this round.

philipbelesky:
1. Any ideas for where this would be? I guess it could be a button or alert on the results page (perhaps for admin's only) that brings up a modal with more details?
2. Definitely; this seems like the critical use case
3. This should be possible; ideally by substituting an existing UI element so the layout is consistent.

czlee:
On (1), I have about as much of an idea as you do—probably don't want inline, so either a modal or a page.

tienne-B:
I'll take it. I think the easiest solution for (1) is to indicate it in the admin results table (like with the "user-x" icon). Wondering if 2 icons can co-exist, one for the win status, the other as an indicator of this.

I already have a prototype for display in the debate table like for adjudicator chair or dissent, using the "magnet" emoji: 🧲. Didn't really know which would be the most appropriate (maybe 👤 or ⚙️ but they don't really stand out) Suggestions (or for using icons)?

(2, 3) are in the works. For the purposes here, if in a 3-person team format it count as iron-manning to only have 2 people.

philipbelesky:
Sorry, the problem I meant to state with (1) is that because the ballots vanish from the table when entered the record of those prior-ironed-teams also goes with them. So given that there is no longer one single list of that information. Happy to have the icons show that info inline but ideally it would also be augmented by having a consolidated list elsewhere on that ballots overview page to aid manual review of the entered data; perhaps behind a modal popup triggered by a button.

tienne-B:
I don't understand the "ballots vanish[ing]" problem... Check the indications I have implemented.

czlee:
We used to have ballots vanish from the table in the assistant view (but not the admin view) once they were confirmed, but we ditched this back in 7c3eac09dd48ad75abd138a1399d932501f51ccb, is that what @philipbelesky is referring to?

philipbelesky:
[...]

PR #868, opened 2018-10-03
Add iron-manning indicators to ballots table

tienne-B:
This is a start on #541. The two commits add annotations to `debate_set_with_prefetches` to find all the teams (or `debateteams`) who have iron-manned in either the current round or the precedent one. More info in the commit messages.


tabbycat/utils/tables.py
```diff
@@ -893,7 +906,7 @@ def add_team_results_columns(self, teams, rounds):
             header = {'key': 'r%d' % round_seq, 'title': round.abbreviation}
             self.add_column(header, results)
 
-    def add_debate_results_columns(self, debates):
+    def add_debate_results_columns(self, debates, iron=False):
```
czlee: Does this argument get used anywhere? (Was it meant to be used in line 922?)

tabbycat/utils/tables.py
```diff
@@ -906,7 +919,8 @@ def add_debate_results_columns(self, debates):
                 team = debate.get_team(side)
 
                 subtext = None if (all_sides_confirmed or not debate.sides_confirmed) else side_abbrs[side]
-                cell = self._team_cell(team, show_emoji=False, subtext=subtext)
+                cell = self._team_cell(team, show_emoji=False, subtext=subtext, show_iron=True,
+                    iron=debateteam.iron, iron_prev=debateteam.iron_prev)
```
czlee: The `iron` and `iron_prev` attributes aren't guaranteed to exist here, because there's no guarantee that it was called with `debate_set_with_prefetches(iron=True)`. The attributes are called even if not used, because they're evaluated before being passed to `self._team_cell`. It might be tempting to guarantee that the attribute exists, but because there's nothing about _any_ `DebateTeam` that can guarantee that it exists (and because this is not a core function of `DebateTeam`, and because you don't know who will call `add_debate_results_columns()` in the future), it's better for this code not to expect it to exist. For example, we could use `getattr(debateteam, 'iron', 0)`.

Sadly, this isn't just hypothetical—the adjudicator record and public results pages now both crash with `AttributeError: 'DebateTeam' object has no attribute 'iron'`.

czlee: On a different note, since this annotation is only used in one place (here), it might make more sense to add to `cell` from this function, rather than pass more arguments to `self._team_cell()`? (This part is up for discussion.)

tabbycat/tournaments/models.py
```diff
@@ -499,13 +499,19 @@ def debate_set_with_prefetches(self, filter_kwargs=None, ordering=('venue__name'
...
                 debateteam_prefetch_queryset = debateteam_prefetch_queryset.select_related('team__institution')
             if speakers:
                 debateteam_prefetch_queryset = debateteam_prefetch_queryset.prefetch_related(
                     Prefetch('team__speaker_set', queryset=Speaker.objects.order_by('name')))
+            if iron:
+                debateteam_prefetch_queryset = debateteam_prefetch_queryset.annotate(
+                    iron=Count('speakerscore__ghost', filter=Q(speakerscore__ghost=True))).annotate(
+                    iron_prev=Count('team__debateteam__speakerscore__ghost',
+                        filter=Q(team__debateteam__speakerscore__ghost=True) & Q(team__debateteam__debate__round=self.prev)))
+
```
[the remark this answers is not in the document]
czlee: Oh, derp, I forgot that `distinct=True` with `speakerscore__ghost` would then count the number of distinct `ghost` values, which of course can then never be greater than one. So we should have changed the count expression too. My bad, will fix.

philipbelesky:
I've pushed up two changes. One tweaks the display of the inline icons to make them slightly more obvious and the other adds the modal list describe above. Let me know what you think!

Also, @czlee, I was at a loss for a button-length bit of text to describe 'iron person' for the button, however the title of the modal (hopefully) displays a jargon-agnostic explanation of what the modal is displaying. Does that work?

tabbycat/utils/tables.py
```diff
@@ -915,6 +915,24 @@ def add_debate_results_columns(self, debates):
                 else:
                     cell = self._result_cell_class_four(debateteam.points, cell)
 
+                if iron and (debateteam.iron > 0 or debateteam.iron_prev > 0):
+                    cell['text'] = "🗣️" + cell['text']
+
+                    popover_text = []
+                    if debateteam.iron > 0 and debateteam.iron_prev > 0:
+                        popover_text = _("Team iron-manned this round and the last.")
+                        warning_level = "text-info"
```
tienne-B: Should the fact that they ironed last round take precedence (and thus get upgraded to `text-warning` level?

tabbycat/utils/tables.py
```diff
@@ -915,6 +915,24 @@ def add_debate_results_columns(self, debates):
                 else:
                     cell = self._result_cell_class_four(debateteam.points, cell)
 
+                if iron and (debateteam.iron > 0 or debateteam.iron_prev > 0):
+                    cell['text'] = "🗣️" + cell['text']
```
tienne-B: I might want to change the emoji to be a magnet once it is widely supported. The magnet has some red colouring which is good to gain attention, and it has a connotation to iron. While "iron" may be regional, a person speaking is debating in a nutshell, and does not convey this sense of missing a team member (no emoji adequately represents it universally...)

philipbelesky: Sounds good, the magnet to signify iron is nice. If the button text stays referencing 'iron' it would be good to tie them together.

We could also potentially use the internal terminology 'ghost' which would then mean 👻 could be dropped in.

tienne-B:
Just added my patches for points 2 and 3. I like the little changes @philipbelesky has made and then modal view. @czlee: Hope I have satisfied your review.

philipbelesky:
Thanks! Both look great, I made a few small changes to the wording. I'll leave this open for CZ to check the text if he has the time otherwise it should be good to merge.

tabbycat/utils/tables.py
czlee: A bit late, but for what it's worth, I concur with @philipbelesky.

PR #878, opened 2018-10-18
Add check for round; closes BACKEND-10G

tienne-B:


philipbelesky:
Just a heads up that I'm assigning to @czlee to this one as he knows this area of the code better than I

tienne-B:
Pinged you as you replied to Omar beforehand.

czlee:
Oooh, actually, this isn't ideal—that's the line that filters out reply speeches, so if there are reply speeches then it risks including reply speeches in that average. The better thing to do would be to find a different way to grab the tournament (not via round), though I think there might be no easy way to do it with the current architecture. It might require us to add `tournament` as a keyword in the constructor and then pass it in, which is mildly annoying.

Also, I just noticed in this code that it should probably filter for `debateteam__speakerscore__ghost=False` as well.

Issue #848, opened 2018-09-29
Use SendGrid API to detect email bounces, deliveries etc.


czlee:
This is probably a nice-to-have but it would be a nice extension to our notifications system if we could pull information from SendGrid about which emails bounced, which ones were delivered, etc., since SendGrid has richer information about this.

(At the moment, our sent message records just record messages that were sent to SendGrid, not anything about whether SendGrid managed to send them.)

tienne-B:
For reference:

https://sendgrid.com/docs/API_Reference/Event_Webhook/getting_started_event_webhook.html
https://sendgrid.com/docs/API_Reference/Web_API_v3/bounces.html
https://sendgrid.com/docs/API_Reference/Web_API_v3/invalid_emails.html

Don't think emails can be fetched, but we can check through these.

tienne-B:
I'm making some progress here (currently on a branch attached to 2d0281c93cd50c8a90b8cff76904b8ae1a9a474d). Figuring out how to present this info.

tienne-B:
Little status update.

# Model Migrations
There will now be three tables in the `notifications` app.

## New Schema
All these changes have migrations that will prevent any data-loss.

### `SentMessageRecord`
Field | Type | Notes
-----|-----|-----
`id` | Primary Key | _Unchanged_ (sadly)
`message_id` | Email | Explained below
`recipient` | Foreign Key | -> `Person`; _Unchanged_
`method` | Enumeration | _Unchanged_
`email` | Email | _Unchanged_
`context` | JSON | _Unchanged_
`message` | Text | Contains full email (with headers)
`notification` | Foreign Key | -> `BulkNotification`

Each email message has its own unique identifier found in the email header as `Message-ID`. This ID is in the form of an email address and so should be held in an email field. While this would be a good primary key, to maintain compatibility with old records (before the headers were also included in the `message` field), it is not.

### `BulkNotification`
Field | Type | Notes
-----|-----|-----
`id` | Primary Key | Automatically generated
`event` | Enumeration | From `SentMessageRecord`; Unchanged
`timestamp` | Date-Time | Current Time; from `SentMessageRecord`; Unchanged
`tournament` | Foreign Key | -> `Tournament`; from `SentMessageRecord`
`round` | Foreign Key | -> `Round`; from `SentMessageRecord`

[...]

Issue #792, opened 2018-09-02
Email draw notifications: Send email to teams on round release


tienne-B:
Continued from #671. The teams should get a notification with the venue (& round), their position (if applicable), opponent(s), and adjudicators for the match.

tienne-B:
Ready; depends on #913 and #915.

PR #879, opened 2018-10-19
Create views to select people to send emails

### LINKED ISSUE #800: Use Summernote Widget for Email bodies
This could allow for simple HTML formatting, which could be useful. The main thing would be to make the personalized information emphasized (**bold** or _italics_) to be able to glance the pertinent information.

Follows (6) of #653.

### LINKED ISSUE #874: Private URLs emails could be partitioned by role
Private URLs may not be relevant for speakers, only for judges in a tournament. Sending everyone a code would then be wasteful. Maybe with the preferences to enable private URL functions emails are sent only to relevant parties.

### LINKED ISSUE #875: Email notifications: Check recipients before bulk sending
Currently, email notifications which are sent in bulk (_e.g._ private URLs, registration information, team points) are sent immediately upon pressing a button. This can be a bit nerve-racking to click, especially when you're trusting the software to correctly "fill in the gaps" with emails that weren't previously sent.

One way to alleviate this was in the old private URL emails system, which had you click through a page where it shows whom it's about to email first. That page didn't allow you to change anything, but at least you knew before you sent them who would receive them.

Now that we have a more comprehensive notifications system (thanks, @tienne-B!), it's probably worth the investment in a more complete, generic solution to providing users with more certainty about what emails they're sending, before it happens. What I have in mind is the following:

- In any bulk email function, you first click into a page that lists potential recipients in a table.
- The table lists everyone eligible to receive the email (all speakers, or all adjudicators, or both).
- The first column of the table is a checkbox column; checked recipients (only) will be emailed.
- The checkboxes are prefilled using the same filter as the status quo: those who have already received an email with the same information as the one they would be about to be sent, are initially unchecked; those who haven't already been emailed or whose information has changed, would be initially checked.
- The table also lists their name, email address, some relevant information (_e.g._ their URL key), and a "status" column indicating whether they've received no such email, or how many emails they've received so far and whether one of them is the exact same one they'd be about to be sent, or whether their email address has changed since they were last sent the email, or whether their information has changed since they were last sent the email.
- Those recipients without an email address _are_ listed, but in faded text and but no checkbox is rendered for them (or the checkbox is disabled).

In terms of implementation:
- Since this idea would be reused with all bulk notifications, there should be some base class and base template from which the specific pages inherit.
- The form would have to be dynamically generated, because it's one checkbox per person in the list.
- The checkboxes don't change a database state, so no need for channels, they should just be posted when the user wants to confirm sending emails.

tienne-B:
As a start to #875, these commits create a template with a list of participants which can be selected/deselected to get sent a custom email, as can be provided at the bottom of the list. These classes are extensible to handle templated messages soon.

The change also adds a `send_to` list when activating the consumer. This list contains the IDs of people who are to be sent messages. This list is then checked in the generators.

I'd appreciate some help in making the table left of the message form (have two columns) and where to link to this new view.

Closes #800.
Closes #874.
Closes #875.


tienne-B:
The main table could also be used to get the status of already sent emails/their status with the participant's email. That would solve the interface aspect of #848.

tienne-B:
This will mean that places like draw generation will need a separate button to send messages. Another thing is what to do with email events that may require many templates. For example, an adjudicator or team draw email would not be the same (or with registration emails). Ideas for what to do in those cases? Wouldn't like having 3 buttons with "Release the Draw"

philipbelesky:
Hey, did you still need a pointer for "making the table left of the message form (have two columns)"? I've had a play with the branch and don't see quite what this might be (maybe its been fixed already?).

As for the location of the new pages, for the case of draw releases it could fit pretty nicely in between the draw display and draw printing printing links under the `Display Draw` area? We could also then potentially deprecate the preference for them and just have sending them always be an explicit action. The motion email release could also live there and be similarly explicit. Alternately they could live under the Release Draw/Release Motion blocks and perhaps indicate their state (i.e. whether things have already been sent). I don't mind if each block gains another couple of items. 

Thoughts? This would mean that those email types would no longer ben an automatic action which I guess would nominally slow things down. But I think most tab directors would probably appreciate the higher level of control/certainty.

philipbelesky:
Hey thanks for the updates! A couple of notes:

- Given Email Motion can send to both speakers and adjudicators I think it probably shouldn't have the subtitle "(to active teams)" as this implies that it can only do the former. It seems like it would be better if either there were two views (one for each type) or there was a way within the table to quickly select/deselect each type (sort of like the availability tables do). What do you think would be better? The same consideration I guess applies to the draw email, although depends on #792
- Emailing motions/draw don't seem to be hooked up to the worker (but private URLs are). Can you reproduce this?
- Would it be possible to extend `default_send` to deselect people for whom there is already a `sentmessagerecord` for this email type? Ideally there would also be a column in the table that shows whether an email has been sent.
- I made a few minor wording changes in e33c3be9763b195e2ac47bd13954cf7d66ccd1db

czlee:
Hey, this looks fantastic so far! Thanks so much for taking this on, @tienne-B!

I have one structural suggestion: Rather than using `self.event_type` to select between templates, can I suggest using subclasses for the different email types? You already have an excellent class inheritance system in place now, I'm just suggesting going one further, and making `EmailAdjudicatorDrawView`, `EmailTeamPointsView`, `EmailPrivateUrlsView`, `EmailParticipantInformationView`, `EmailMotionsView` (or rename as you think fit) subclasses of `TournamentTemplateEmailCreateView` and `RoundTemplateEmailCreateView`. Then, rather than parsing event types from the URL, they'd each just have their own `path()` entries in `urls.py`.

This might be counterintuitive, in that the functionality is mostly common and differentiable using just an event name. But hopefully the following will give an idea of why I think more subclassing would help:
- The app that "knows" the relevant things to send is actually the individual apps: `draw` for adjudicator assignments, `standings` for team points, `motions` for motions, `privateurls` for private URLs and `participants` for registration. It makes more sense (I'd argue) to think of it as a base class in `notifications` providing a service (base class) to those apps, that they can then customize (as a subclass) for their own purposes. Then `EmailPrivateUrlsView` shows up in `privateurls/views.py` and `privateurls/urls.py`, _etc._ This way, the relevant information about what we're sending is in the apps that govern that information.
- It would save having to use `if`…`else` statements to choose between recipient lists, as we do in `RoundTemplateEmailCreateView`, and instead just have each subclass specify what it wants, so you can read a single event type's subclass and get a feel for what it's doing (rather than parsing out its own `if` blocks).
- I can see that it's convenient to have `self.event_type + "_email_subject"` as a convenience shortcut, but this sort of pattern is hard to discover for readers of code (you can't `git grep` the preference to see where it's used), and constrains flexibility in choosing names. So even here, I'd suggest making the preference name a class attribute, `subject_preference_name = 'team_points_email_subject'`.
- `get_page_subtitle()` then wouldn't need to be overridden, because you can just set the `page_subtitle` class attribute of each subclass; ditto for `get_success_url()` (I think).
- If we ever have a weird case that is an "odd one out", it's easier to subclass and override, while keeping consistent with the existing pattern.

Let me know whether that made sense!

tienne-B:
It makes sense! Even better, the change removed other dicts in `consumers.py` :)
However, subclassing under the other apps seem to be inconsistent with apps such as `printing`.

Currently trying to get selection/deselection buttons (and debug the async hook)

tabbycat/templates/tables/CheckCell.vue
```diff
@@ -1,12 +1,13 @@
...
+  <td :class="cellData.class ? cellData.class : null" >
 
-    <span v-if="cellData['sort']" hidden>
+    <span v-if="cellData.sort" hidden>
       {{ cellData.checked }}
     </span>
     <div class="form-check">
-      <input type="checkbox" class="form-check-input  position-static"
+      <input type="checkbox" class="form-check-input position-static"
+             :name="cellData.name" :value="cellData.value"
```
tienne-B: Why doesn't this work?

tienne-B:
All the mass emails work now. However email receipts for voting formats do not (consensus ones do...)
There is the same problem with b183f0265f512edb8471d0b4e652c8743406b40d applied or not.

philipbelesky:
Hmm yea, the checkbox construct is designed save 'checks' in an async manner that may not apply easily here as you want to check status to stay local until submission. We may need to abstract that part of the checking out to a Mixin as well as to use slots/template inheritance to break the assumption that `CheckboxTablesContainer` is only used in availabilities (if we go down that route).

How are the checkboxes currently wired up (or are they not?). I wonder if rather than using a Vue table we could present a form (perhaps styled in a table-like manner) with the checkbox fields. Any filtering buttons could then use jQuery to set the HTML element directly, perhaps based on a class property distinguishing between adjs/teams.

tienne-B:
The checkboxes are currently wired-up, using participants' IDs to as the `value` passed to the form. I think it just took many passes of building Vue for the change to be recognized. Also the vue table is encapsulated in the HTML form to connect the checkboxes.

These buttons should be specified in the subclasses (just as the "role" column) as useful filters may differ between events.

philipbelesky:
Ah yes, adding to the sub-view would make more sense; I forgot that is added across all of those tables.

By "not all adjudicators are checked by default" what I mean is if you load up the view in question, some participants in the list are checked and others are not; I think in this case speakers are default-checked and adjudicators are not-checked. Without being able to see who is an adj vs who is a speaker the checks appear random.

On the vue front, are you using `USE_WEBPACK_SERVER` to have it rebuild automatically with `npm run serve1? Another consideration is that your browser may be caching the local files which should be disable-able in the development inspector.

tienne-B:
What view(s) are you talking about? Mind if I overwrite your last commit?

tienne-B:
Think you're talking about motion release. Mind if I overwrite your last commit?

philipbelesky:
Yes motion release was the one (I now have the code in front of me). Feel free to overwrite!

philipbelesky:
Hey thanks for the ongoing iteration on this, just a quick heads up that I probably wont have time to review things properly until the weekend.

tienne-B:
No problem; got other changes in development as well! Anyways as this PR covers _three_(!) issues, it does take a thorough check :)

philipbelesky:
Hey I took this for a spin and it all looks great! I did notice the speaker/adj buttons were not working which seemed to be an issue with passing `mixed_participants` up the inheritance hierarchy. I've also added a few quick methods to the checkbox table to prevent it trying to save information (this was causing console errors) and to hide the autosave counter. 

I'll give you a chance to check the last commit, otherwise let me know and I'll merge it in. Feel free to also add a commit updating the CHANGELOG if you'd like to add that entry.

tienne-B:
Looks good :)

philipbelesky:
🎊merged🎊, thanks again!

PR #880, opened 2018-10-21
Include only preliminary scores in speaks overview

tienne-B:
In the standings overview, the top scores would have the elimination
rounds included. Or, they do not receive scores so they become None
which deplaces real scores. This commit filters the scores to only
include preliminary rounds.

### DIFF
--- a/tabbycat/standings/views.py
+++ b/tabbycat/standings/views.py
@@ -67,6 +67,7 @@ class StandingsIndexView(AdministratorMixin, RoundMixin, TemplateView):
             'debate_team__team__institution'
         )
         if self.tournament.pref('teams_in_debate') == 'bp':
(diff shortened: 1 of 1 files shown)


philipbelesky:
Should this filter be up at line 61 so that it will affect both BP and non-BP formats?

tienne-B:
No, because non-BP formats have scores in elimination rounds, and so could still be in "highest/_lowest_ scores"

philipbelesky:
Ah, of course, should have thought of that!

tienne-B:
As this is a bug which affects the master, should this be included in the next hot-fix?

philipbelesky:
That makes sense, I've picked it into a `2.2.6` branch. It probably should have gone into the one just released also, that was an oversight.

PR #882, opened 2018-10-23
Fix email forms for ballot receipts

tienne-B:
Ballot receipts must look different between BP and 2-team formats as the latter have a boolean win/loss instead of points. Points are now fixed, as the ranking is the inverse of points.

The Win/Loss is also fixed in the ballots email for two-team formats, resolving the problem with these emails not sending.

### DIFF
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -91,11 +91,18 @@ def ballots_email_generator(debate_id):
         ballot = ""
 
         for side, (side_name, pos_names) in zip(tournament.sides, side_and_position_names(tournament)):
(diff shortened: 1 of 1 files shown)


tabbycat/notifications/utils.py
```diff
@@ -108,6 +112,7 @@ def _create_ballot(result, scoresheet):
         return ballot
 
     if isinstance(results, VotingDebateResult):
+        results.load_scoresheets()
```
tienne-B: So the problem seemed to be that the scoresheets were not being loaded. Any idea how to fix it?

philipbelesky:
Cheers!

PR #912, opened 2018-11-03
Enable motion emails for adjudicators

tienne-B:
It was looking for speakers in the selected list, disregarding adjudicators. This removes the check for team membership.


tienne-B:
Thought I had changed that last `speaker`

PR #914, opened 2018-11-04
Add attribute check for venue in adj draw email

tienne-B:
As the venue can be null, a check was needed. Closes BACKEND-118.


philipbelesky:
Thanks! I'm pretty sure `attr` parameter in `hasattr` needs to be a string; I've added a further commit to change that (only noticed after merging). Given BACKEND-118 was reported on 2.2.X the file in the `VENUE` context is actually in `draw/utils.py` so I've added a further commit to backport the fix.

PR #915, opened 2018-11-06
Format templated emails as HTML

tienne-B:
Now the defaults for templated emails use HTML, including parameters such as `{{ SCORES }}`. To keep compatibility, instead of stripping HTML tags, the emails are converted into Markdown, retaining some formatting in plain text.

~Some current problems:~
* ~Marking HTML variables (such as `{{ SCORES }}` as safe. Otherwise the emails fail~
* ~Ballot email setting doesn't want to change widgets to Summernote for some reason~
* ~Email templates may oddly be "cached"~


tienne-B:
Fixed some of my errors.

PR #916, opened 2018-11-08
Fix viewbox for QR codes

tienne-B:
While the size of the codes are of 37px, QR codes have a buffer of 4px. This caused the bottom and right sides to be cut-off. To keep the buffer, a break tag was added.

### DIFF
--- a/tabbycat/printing/templates/randomised_url_sheets.html
+++ b/tabbycat/printing/templates/randomised_url_sheets.html
@@ -40,7 +40,7 @@
         {% endblocktrans %}
       </p>
       <p>{{ p.url }}</p>
-      <svg version="1.1" viewBox="0 0 37 37" width="200px" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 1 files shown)


philipbelesky:
In which browser is this? I've had a look at Safari and Chrome without any luck. It also seems like a problem a break tag might not be the best fix for — maybe a margin or a better positioning of the svg element.

tienne-B:
This is both in Safari and Chrome. `viewbox` behaviour seems consistent between browsers. Just to be clear, 
badqr
is cut-off.

I have pushed a new version that doesn't use the break tag.

philipbelesky:
Hmm, I was looking for slipping of the edges on both of those browsers and couldn't seem to find anything. But anyway, no problem merging this in, thanks for the PR!

philipbelesky:
I guess they just looked normal? The anchors in the corners seemed to have their borders unclipped. I wouldn't be surprised if there are some differences in how SVG is rendered even between pretty comparable setups.

Issue #811, opened 2018-09-09
Split out some Import Data pages


philipbelesky:
Problems:

- The conflicts pages in particular has a huge response time (due to rendering all the foreign key fields across all values) at large tournaments. It risks timing out or severely delaying other requests
- When the list of conflicts gets very long the space to add new items is way down at the bottom of the page and is potentially overlooked or at least annoying to get to

Proposal:

- Separate out the Edit and Add pages into distinct views for venue categories/constraints and conflicts
- The Add pages can render the full list of options for each addition (since there will be at max 10 `extra` options)
- For the Edit versions of the Conflicts pages the conflicter/conflictee fields shouldn't be big selects — they can just show the existing data inline without the option for editing (only deleting)

This does mean that those pages essentially become 'delete' pages rather than edit pages, but I'm not sure the use case of going in an editing a conflict to be a different person ever really happens?

tienne-B:
Would #590 be another possible solution to this?

philipbelesky:
Hmm, I'm not sure if it would work given the time is spent within the templating system. There might be a way to pass that off. However to the extent that #590 is implemented I'd prefer it to actually be used for actions that are asynchronous from the perspective of the user — i.e. for in-page actions (like allocations) or actions that are 'fire and forget' (like email). In this case splitting out the views makes sense (I think) from a UX perspective so pushing the render on to a task runner wouldn't help with that angle.

philipbelesky:
Yea, I wouldn't mind keeping them on the same page. Although I think the case specified of deleting and adding being made easier might need to be balanced against the general task of adding being less convenient by having to scroll all the way down to the end of a long list.

czlee:
Hmm, true, that's a fair point.

philipbelesky:
Sorry, should have clarified: the problem is not duplicate SQL queries but rather the render time, I assume due to the fact that each row contains two select elements with all possible values — e.g. all adjudicators and all teams in the case of adj-team conflicts. At a tournament with 100 adj-team conflicts rendering takes around 40 seconds for me (running locally). 

I don't see the SQL queries showing either. So perhaps the render time is masking the SQL loop; or whatever method the forms widget uses doesn't report them to Debug Toolbar but is still slow regardless.

tienne-B:
Ref #2578 that solves the first problem

PR #919, opened 2018-11-13
Add filter for institutions in constraints

tienne-B:
Not all institutions may be relevant to be included in the constraints form-sets. When no one from an institution is present in a tournament, there is no need for the institution to be listed. This commit adds a property to the Tournament model to get the relevant institutions.

### DIFF
--- a/tabbycat/adjallocation/views.py
+++ b/tabbycat/adjallocation/views.py
@@ -14,7 +14,7 @@ from actionlog.models import ActionLogEntry
 from availability.utils import annotate_availability
 from breakqual.models import BreakCategory
 from draw.models import Debate
-from participants.models import Adjudicator, Region
+from participants.models import Adjudicator, Institution, Region
 from participants.prefetch import populate_feedback_scores
 from tournaments.models import Round
 from tournaments.mixins import DebateDragAndDropMixin, LegacyDrawForDragAndDropMixin, RoundMixin, TournamentMixin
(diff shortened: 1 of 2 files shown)


czlee:
I would prefer to list all institutions, actually, for (at least) two reasons:

1. Sometimes there are double-past institution conflicts affecting many people, even though no one from that institution is at the tournament now. For example, a European tournament might include five people who used to go to Harvard. Even though no one at the tournament is affiliated with Harvard _now_, it's still easiest to enter conflicts with an institution object for Harvard (than making each of the up to 10 personal conflicts that would be necessary).

2. Sometimes you might know that an institution will have teams later, but has nothing now. Then, when adding an adjudicator with a past affiliation to this institution, you might want to add that conflict in now, rather than relying on remembering to go back and add all these institution conflicts when you finally add its teams.

tienne-B:
So if I understand your point 1, if there are team-institution and adjudicator-institution constraints pointing to the same institution, there are implicit team-adjudicator constraints?

This doesn't disable creating conflicts with institutions which aren't on the list.

The problem I was trying to solve here was that when there are many tournaments on one site, a full list of institutions would make the list unwieldy.

czlee:
Yes, you understood point 1 correctly, and it's a pretty routine occurrence in the university circuit, so I don't consider it to be an edge case. It's also counterintuitive to a user who's left scratching their head about why an institution they added doesn't show up on the institutions list.

We should discuss what needs to happen to better support the needs of multi-tournament sites, but as things stand, I consider this change to be a regression in at least some common use cases. There are already known design flaws with this page (see #811), so I wonder if rethinking how this page works (not just in terms of #811) would help it address both use cases.

PR #925, opened 2018-11-18
Check if venue attribute is None, not if exists

tienne-B:
Little error; closes BACKEND-12M. Also made the venue use the display name.


tienne-B:
Also #923.

tienne-B:
Would you rather I just submit one PR and one of you cherry-pick, or two PRs?

philipbelesky:
Sorry do you mean two PRs because the code wont merge nicely into `develop`? It that's the case it would be appreciated but definitely not necessary!

Issue #429, opened 2017-02-18
Add venue rotation feature


czlee:
The idea is to (automatically) rotate venues so that the same teams aren't always far away in every round. This is probably nontrivial, some preliminary thoughts:

- The primary new information would presumably be some indication of distance attached to each venue. It might be tempting to overload this onto existing fields, for example:
   - We might use "priority" as an inverse of distance.
   - Instead of tying this to venues, we might be tempted to tie it to venue groups for convenience.

  My inclination is against both of these, for the following reasons:
   - Low-priority rooms might not always be far away, for example, they might be a nearby room that's just not very nice.
   - Not all venues in a group, for whatever reason, might be the same distance away. Also, if venues can have multiple groups, the idea of tying distance to venue groups just makes no sense.

  So I think we'd just want to add a new field `distance` to `Venue`.

- `distance` should be a float, and it should be optional. Tabbycat should show a useful error if venue rotation is enabled, and it tries to allocate venues, but any venue lacks a distance.

- We'd have to get this to work with venue constraints somehow. Venue constraints are strict priority, so presumably constraints would always win over rotation.

- We'd have to know venue history, but that's easy enough, because the information's already in the draw.

The natural thing to do would be to allocate the nearest rooms to those debates with teams that have had the farthest history, choosing randomly between those that have the same distance. But this might have odd effects: It might cause some teams to ricochet between very far rooms and very near rooms, in some circumstances, rather than allowing them a chance at "moderate" rooms at some stage.

philipbelesky:
This seems like its being made overly complicated. Adding distances on a per-Venue basis is a lot of data entry work that I imagine almost no tournaments would use. Requiring this limits a feature that is most likely strictly better than random allocation to a handful of tournaments rather than making it a broad and universal improvement over random allocation.

I think a much simpler solution that would deliver the vast majority of the benefit would be:

- Constraints always win over rotation
- Assume VenueGroups/Categories are a good enough proxy for distance. For each team prefer finding a Venue in VenueGroup/Category that it has not been before, or (failing that) a VenueGroup/Category it has been in the least.
- Within each assigned VenueGroup/Category check and ensure the actual Venue itself is not a repeat of an earlier allocation.

[...]

PR #926, opened 2018-11-18
Split venue constraints page(s)

tienne-B:
As subclasses of the adjudicator conflict views, the venue constraints views have been split in the same fashion and listed in the simple importer page. New action log events have been created to specify what kind of constraint have been added.

### DIFF
--- a/tabbycat/actionlog/models.py
+++ b/tabbycat/actionlog/models.py
@@ -39,7 +39,11 @@ class ActionLogEntry(models.Model):
     ACTION_TYPE_VENUES_SAVE                       = 've.save'
     ACTION_TYPE_VENUES_AUTOALLOCATE               = 've.auto'
     ACTION_TYPE_VENUE_CATEGORIES_EDIT             = 've.ca.edit'
-    ACTION_TYPE_VENUE_CONSTRAINTS_EDIT            = 've.co.edit'
+    ACTION_TYPE_VENUE_CONSTRAINTS_EDIT            = 've.co.edit' # obsolete
+    ACTION_TYPE_VENUE_CONSTRAINTS_ADJ_EDIT        = 've.cj.edit'
+    ACTION_TYPE_VENUE_CONSTRAINTS_TEAM_EDIT       = 've.ct.edit'
+    ACTION_TYPE_VENUE_CONSTRAINTS_INST_EDIT       = 've.ci.edit'
+    ACTION_TYPE_VENUE_CONSTRAINTS_DIV_EDIT        = 've.cd.edit'
     ACTION_TYPE_MATCHUP_SAVE                      = 'mu.save'
     ACTION_TYPE_SIDES_SAVE                        = 'ms.save'
     ACTION_TYPE_DRAW_RELEASE                      = 'dr.rele'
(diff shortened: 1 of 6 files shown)


tienne-B:
I'm placing my commits that implements #429 in here as I based the work on the same branch, and PRs can't depend on other PRs. This lets you look and comment at the new changes that I am making (and steer me on the right path if I'm diverging).

Should be able to merge the first commit separately though by fetching the branch and then merging the first commit (then pushing to GH). A bit convoluted though...

tienne-B:
@czlee: FYI, the rows are the debates, columns are the venues (opposite of your comment) ;)
@philipbelesky: Might ask for some help with integrating these auto-allocators in the interface :)

philipbelesky:
Just catching up on some PRs that were slated for 2.4, is this ready for review?

tienne-B:
Yes, it's ready for review (2.4)

philipbelesky:
Would it be possible to add unit tests to this? For major new features like this, which I don't expect to ever be using myself in a 'real' scenario, I think we need to increasingly rely on testing to ensure we can support/release things in the long term.

tienne-B:
It would be possible to add some tests to things such as cost calculations (although it would take quite a bit of setup). Don't think it would be necessary for the actual allocation as it uses the same logic as adjudicator allocations (and in my view relatively straightforward). Was that what you had in mind?

philipbelesky:
Did you still want to review this @czlee? I'm happy to do so, although would just be testing the UX and bugs. Before I do so it would be helpful if you could handle the merging of `develop` into this @tienne-B

tienne-B:
> Venue allocation methods have been added which returns optimal venue assignments based on the given constraints. In addition, new methods allow for venue rotation so that teams are not always in distant venues, or so that they stay in the same area.
CHANGELOG entry - removed to avoid more conflicts

tabbycat/venues/allocator.py
```diff
@@ -168,7 +196,139 @@ def allocate_unconstrained_venues(self, debates):
...
+        for t in teams.all():
+            prev_team_vc[t] = []
+            for dt in t.debateteam_set.all():
+                prev_team_vc[t].extend(dt.debate.venue.venuecategory_set.all())
+
+        self.prev_venue_categories = {}
+        for debate in self.debates:
+            self.prev_venue_categories[debate] = []
+
+            # Ajudicators are excepted from this constraint
```
czlee, line 259: typo 🙂

tabbycat/venues/allocator.py
```diff
@@ -168,7 +196,139 @@ def allocate_unconstrained_venues(self, debates):
...
+            for dt in debate.debateteam_set.all():
+                self.prev_venue_categories[debate].extend(prev_team_vc[dt.team])
```
czlee, line 261: Is "previous value" functionality unused in the non-rotation Hungarian allocators? It might make the code a bit easier to understand if this is spun out to an intermediate base class, so like:

```
                                                                RotationVenueAllocator
                                                              /
HungarianVenueAllocator  <-- BaseHistoryHungarianVenueAllocator
                                                              \
                                                                StationaryVenueAllocator
```

(I realise this makes the class structure more complicated, but it did confuse me for a bit.)

Also, I try to name classes `Base` if and only if they're abstract base classes (not to be used directly); if it can be used directly, it's fine to not call it `Base` even if lots of things derive from it.

tabbycat/venues/allocator.py
```diff
@@ -168,7 +196,139 @@ def allocate_unconstrained_venues(self, debates):
...
+            Prefetch('debateadjudicator_set', queryset=DebateAdjudicator.objects.select_related('adjudicator', 'adjudicator__institution')),
+            Prefetch('debateteam_set', queryset=DebateTeam.objects.select_related('team', 'team__institution')),
+        )
+
+    def allocate(self):
+        self.prev_venues = self.get_previous_venue_categories()
+        self.constraints = self.collect_dict_constraints()
+
+        self.max_venue_score = self.venues[0].priority
+        self.max_constraint_priority = list(self.constraints.values())[0][0].priority
```
czlee, line 233: This line crashes with an `IndexError` if `self.constraints` is empty.

(This is the only "change requested".)

czlee (changes requested):
Sorry about the long delay. Looks great! I picked up a crash when the constraints list is empty, but when the constraints list is not empty it looks good. The requested change is just to fix that crash.

Not necessary for an MVP, but I think it would be nice if the user had some way to verify that the rotation options are working as intended. With constraints we do this by listing met and unmet constraints in the rightmost column. This might not work as well with history, since I guess in the standard case we'd expect teams to cycle back to the category, and it's not really a critical thing so might feel like clutter. That said, the only other idea I have is a dedicated page, like the "Draw Details" page, which tbh I think would be way overkill for this kind of feature (so don't actually do it).

I'm not too concerned about unit tests for the actual algorithm. Of course it'd be nice, but probably not one of my priorities as far as unit tests ago.

PR #927, opened 2018-11-18
Use template array in response_class()

tienne-B:
As the TemplateResponse uses an array, it might be necessary for the custom response (for errors in scoresheet views).

Closes [BACKEND-12P](https://sentry.io/tabbycat-ig/backend/issues/768062785/).


tienne-B:
Candidate for `hotfix/2.2.8`

PR #928, opened 2018-11-18
Fix Back link in legacy allocation views

tienne-B:
The "Back" link was broken somehow, but the legacy views are still useful.


philipbelesky:
Cheers!

PR #930, opened 2018-11-24
Split the "In this Round" card

### LINKED ISSUE #891: Private URL landing pages should also contain the information in participant records
Most important is the "In This Round" box, but it wouldn't hurt to have past results there as well. (The "About this participant" box probably isn't so useful.)

At the moment, the "home" for a participant on the tab website is both the participant's record page _and_ the private URL landing page. The participant's record page should stay where it is, because it's also useful for the public. But it'd be easier for participants to make the landing page a one-stop shop for the subject of the page.

The implementation should probably involve some sort of template inclusions and view mixins to avoid code duplication.

tienne-B:
This commit splits the components of the "In this Round" card from team and adjudicator pages into separate files to reduce code duplication. In addition, a new parameter for the grammatical gender was added between the third and second person. Fixed some i18n omissions too.

The box is now on private URL landing pages. It uses the second person (motivation for the new parameter). With the information showing when the draw is released, the page is useful (active) even if no actions are available.

As it is private to an individual, it is unaffected by whether the public record pages are enabled (but it waits for the draw to be released :) )

Closes #891.


czlee:
Is it grammatical gender or grammatical person we're talking about? The name `grammatical_gender` appears a couple of times in the current changes, but it's mostly `grammatical_person`.

I haven't tried out your code live yet, so without having seen it in action: I agree that there's at least a prima facie case for the "about the participant" box being useful for the reasons you say, so I'm happy for it to be on the page somewhere, maybe just deemphasized a bit since people won't be looking at it much relative to other content on the page.

I'm open to links being in the table, though you want to remember that the use case for most people is (I believe) clicking through on their phone, so we really want the ballot/feedback forms to be easy to hit without having to scroll around—that is, a large button near the top of the page. Not mutually exclusive, of course. It's hard to give a firm opinion without seeing it first. Also, @philipbelesky  tends to be better at page layout questions than me 🙂

tienne-B:
Oups, little typo :) Fixed.
I'll wait for Philip for layout, and for de-emphasis (I'm not good at design either) Was thinking to put the buttons over the cards.

philipbelesky:
Hey thanks again for adding this and sorry for the delay. It's great to have the extra information here. 

At present I made a mild tweak to the way conflicts are displayed so that they are inline rather than in bullet lists. I would also like to reduce the height of that registration box further, as its a little awkward when a round has not been released, or when on mobile. However I think that can be shelved for later.

philipbelesky:
Sure, no problem. The options are a little cryptic to me but I will try and remember to not debase. Also let me know if you ever need a force push override for the history.

PR #953, opened 2018-12-12
Remove comparative notes on shadow draws

tienne-B:
Minor documentation update.

With preset panels the behaviour is similar between the tab software compared in the document.


philipbelesky:
Great, thanks!

PR #954, opened 2018-12-16
Fixed ballot adjudicator formating

nickdbush:
Printed ballots with two adjudicators print with ```Adjudicating with & Adjudicator Name (Type, Institution)``` under the ballot title. Printed ballots with three adjudicators print with ```Adjudicating with Adjudicator Name (Type, Institution); & Adjudicator Name  (Type, Institution)``` under the ballot title.

Proposing that the first case becomes ```Adjudicating with Adjudicator Name (Type, Institution)``` and the second case becomes ```Adjudicating with Adjudicator Name (Type, Institution) & Adjudicator Name  (Type, Institution)```

### DIFF
--- a/tabbycat/printing/templates/PrintableDebateInfo.vue
+++ b/tabbycat/printing/templates/PrintableDebateInfo.vue
@@ -10,8 +10,9 @@
           <span v-if="ballot.debateAdjudicators && ballot.debateAdjudicators.length > 1">
             Adjudicating with
             <span v-for="(da, i) in panellistsExcludingSelf">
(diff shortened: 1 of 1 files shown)


tienne-B:
Looks good! I'm not that opinionated on the lack of the Oxford semi-colon :)

philipbelesky:
Hey @nickdbush thanks for the PR! Let me know if you need a hand with any future ones or had any questions about the project.

This change looks great; happy to merge. Those ballot templates are slightly hellish in the mix between vue logic and templating so thanks for digging into it!

PR #957, opened 2018-12-27
Allow private URL-based view of draft ballots

tienne-B:
For transparency, adjudicators' ballots should remain visible to them at all times. This change makes the scoresheet view work when the ballot is still in a draft state (instead of having to wait until the round is marked as completed)


tienne-B:
Ready for merge. The first commit is the bug-fix (or else the private URL landing page won't work after round 1). The next two are improvements on the private url scoresheet page. If those require more discussion, could the first one be merged separately? Didn't want to make 2 PRs as they would conflict.

philipbelesky:
Hey, thanks for adding this in. While this is good to have, it seems like it could be made to be much more valuable if adjs had a way to see their past ballots (perhaps from the table on their private URL page). Otherwise I'm not sure how anyone (including admins) would actually be able to find the correct URL for an individual ballot — its only ever linked to from the public results page iirc.

tienne-B:
That was already implemented...

philipbelesky:
Sorry, I missed that part because I assumed it would show when the ballot was still in a draft state (as per this issue). Instead it seems to only show the link when the ballot is in a confirmed state, and even then the link only actually shows the ballot itself once all the round's results are released.

This seems contrary to the intent of this PR as described in the first message. Was that an oversight, or is there a reason to hide the ballot this way until both the ballot and round is confirmed?

philipbelesky:
To clarify the above edits: a bug in the results scoresheet made me misunderstand things. I still think theres a question of whether the ballot should be shown when it is a draft/unconfirmed state though. Moreover, even if the answer to that question is that it shouldn't be shown, ideally there would be some sort of feedback in the private URL table that indicates if/when it will be shown — otherwise it will likely seems very mysterious to the end-user. 

I'm happy to add those capabilities in if you'd prefer. What was the thinking behind not showing unconfirmed ballots?

tabbycat/utils/tables.py
```diff
@@ -868,10 +869,10 @@ def add_debate_ballot_link_column(self, debates, show_ballot=False):
         elif self.private_url:
             ballot_links_data = []
             for debate in debates:
-                if not debate.confirmed_ballot:
-                    ballot_links_data.append("")
+                if not debate.ballotsubmission_set.all().exists():
+                    ballot_links_data.append(_("No ballot"))
                 elif self.tournament.pref('teams_in_debate') == 'bp' and debate.round.is_break_round:
-                    ballot_links_data.append("")
+                    ballot_links_data.append(_("Elimination"))
```
tienne-B: Didn't really know what to use here...

tienne-B:
I don't really remember why I limited it to confirmed ballots (but round uncompleted). Possibly because I had in mind if the non-latest version of a ballot (draft status) was being examined, the incorrect one would be presented. Changed to include drafts as I can't really think of where that would happen.

Added messages for if the ballot is unavailable or cannot be presented (elimination ballots - 1f9fde6).

philipbelesky:
Ok, np, that's still a big gain over the current lack of ability to see any of them. I think it's worth merging and getting some feedback before deciding whether the filter needs to be more liberal. Thanks for the PR!

PR #979, opened 2019-01-07
Show read-only check-in status if self-checkin disabled

tienne-B:


philipbelesky:
Hey, thanks for the addition. In cases where checkins are used I think this will be super useful, particularly as I think we will need to move the public checkins status page to be static rather than live-updating.

However, the check-in system is generally only used for the largest tournaments and I worry the "You are not checked-in" message would confusing (or at least superfluous) when the check-in system is not being used. I'm not sure what conditional would avoid this — perhaps an initial check for *any* check-ins within the window (from people; not ballots or venues) before showing the message?

tienne-B:
Good point. Done.

philipbelesky:
Thanks, merged with a CHANGELOG credit. Let me know if I missed anything to credit regarding private URLs. The CHANGELOG as a whole also needs to be audited/updated and is pretty far behind `develop` — will get to this soon.

PR #984, opened 2019-01-12
Don't select people with duplicates by default

tienne-B:
This commit adds filters for participants for who to select by default to exclude participants who have already been sent a notification of the same type in the same tournament/round.

### DIFF
--- a/tabbycat/notifications/views.py
+++ b/tabbycat/notifications/views.py
@@ -275,12 +275,20 @@ class TemplateEmailCreateView(BaseSelectPeopleEmailView):
 
 class TournamentTemplateEmailCreateView(TemplateEmailCreateView):
 
+    def get_default_send_queryset(self):
+        return super().get_default_send_queryset().exclude(
+            sentmessagerecord__notification__event=self.event, sentmessagerecord__notification__tournament=self.tournament)
+
     def get_extra(self):
         extra = {'tournament_id': self.tournament.id}
         return extra
 
 
 class RoundTemplateEmailCreateView(TemplateEmailCreateView, RoundMixin):
+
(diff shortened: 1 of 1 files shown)


philipbelesky:
Hmm, what happens here if I want to send out a second set of emails to update information that was incorrect earlier? I.E. if I forgot to allocate venues before emailing the draw? 

This seems like it might be a bit of a 'magic' behaviour that could be confusing as compared to the lesser annoyance of just receiving two emails.

tienne-B:
Good point. The idea is that for most situations where re-sending messages is warranted, it would only be for a subset of the potential recipients, of who would probably be remembered. Having everyone (or most) selected would not be wanted (but I do see your situation, and am disappointed that JSON fields can't be compared)

philipbelesky:
Ah right, I think that makes sense and the tables make it pretty clear who each email is going to. If there is a need to provide new data its probably either a case of sending to a very small number of people or sending to *everyone*. The table accommodates either option pretty well.

tienne-B:
Has there been a change in the selection table code? It now shows "No changes" and not the selection options (works in the old branch though)

philipbelesky:
Sorry, which table should I be looking at? And by selection code do you mean the buttons for All/None that toggle the checkboxes?

tienne-B:
The people selection table (any of them; private URLs email as example)
Yes, the buttons for All/None. Also the button indicating changes. Odd regression...

philipbelesky:
Yup, that's the problem. See 135cc2beeaec94f0eabee30e35991972efb3f05f for the solution (it was caused by a template inheritance issue). Also as an FYI if you use Chrome for development the Vue.js devtools extension will make debugging a lot easier.

PR #989, opened 2019-01-12
Fix selection buttons in participant email table

tienne-B:
This commit makes the email navigation items into a transclude-type
template, rather than hierarchical, as the pages use different base
templates.

Partially reverts 82778855db9bc6e597eba7aafc59bf4317f6d180 (discussed in #984), supersedes 135cc2beeaec94f0eabee30e35991972efb3f05f.


tienne-B:
You beat me to it! What about having the nav-bar on the page?

philipbelesky:
Ah true. Hmm the template structure for the navbar would probably need to be reworked then so that it is an included item rather than an inherited block. Would you like to add that?

tienne-B:
This patch does that.

PR #992, opened 2019-01-13
Remove atomic-type of notifications migration

tienne-B:
After discussion in #913. Splitting the migration would be harder for the same thing.
Closes [BACKEND-16T](https://sentry.io/tabbycat-ig/backend/issues/841911615/).

The migration comported both a schema and data migration, which caused
an error when filling in message-ids. This commit corrects the behaviour
by specifying the migration as non-atomic.

The commit also makes the BulkNotification data migration check for
successive events, and not timestamps. It also removes the Regex to look
for Message-IDs in favour of Python's email module.


tienne-B:
Sentry issue: [BACKEND-16T](https://sentry.io/tabbycat-ig/backend/issues/841911615/?referrer=github_plugin)

tabbycat/notifications/migrations/0007_migrate_data.py
```diff
@@ -0,0 +1,48 @@
...
+
+    for m in SentMessageRecord.objects.all():
+        if m.event is not current_event:
+            current_bulk += 1
+            current_event = m.event
+
+            BulkNotification.objects.create(event=m.event, timestamp=m.timestamp,
+                                            round=m.round, tournament=m.tournament)
+        m.notification_id = current_bulk
+        m.save()
```
czlee: Hmm, it's probably better not to assume that IDs will start at 1 and be ordered sequentially—in most cases they will be, but there are probably edge cases where you end up having to start at a higher number or whatever. Since `.create()` returns the new object, could just look up the ID of the object you just created?

Also, does this loop assume that all `SentMessageRecord` instances relating to the same event are next to each other in the list? Would we want to use `SentMessageRecord.objects.order_by(...)` to guarantee this? Also, `[itertools.groupby()](https://docs.python.org/3/library/itertools.html#itertools.groupby)` might be useful 🙂

tabbycat/notifications/migrations/0007_migrate_data.py
```diff
@@ -0,0 +1,48 @@
...
+
+    current_event = None
+    current_bulk = 0
+
+    for m in SentMessageRecord.objects.all():
+        if m.event is not current_event:
+            current_bulk += 1
+            current_event = m.event
+
+            BulkNotification.objects.create(event=m.event, timestamp=m.timestamp,
```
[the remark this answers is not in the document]
tienne-B: I'll set auto_now in the third migration.

tabbycat/notifications/migrations/0007_migrate_data.py
tienne-B: I haven't really looked into itertools before. Also yes, it does assume that, but I'll correct it.

tienne-B: Grouped by timestamp then event. For ballot receipts, made a further conditional. Didn't use itertools.

czlee:
@tienne-B Mind having a look at my additions and letting me know if you have any objections?

tabbycat/notifications/migrations/0007_migrate_data.py
```diff
@@ -2,32 +2,33 @@
...
-import re
+import operator
 
 from django.db import migrations
 
 
 def create_bulk_notifications(apps, schema_editor):
-    SentMessageRecord = apps.get_model("notifications", "SentMessageRecord")
-    BulkNotification = apps.get_model("notifications", "BulkNotification")
+    SentMessageRecord = apps.get_model("notifications", "SentMessageRecord")  # noqa: N806
```
tienne-B: Aren't the migrations excluded from flake8?

czlee: umm... yeah. I think it might be my local configuration I need to fix.

czlee: They're only really excluded for the long lines in the automatic migrations though; in principle I'd like data migrations to be compliant.

tabbycat/notifications/views.py
```diff
@@ -117,8 +117,11 @@ def get_tables(self):
 class EmailEventWebhookView(TournamentMixin, View):
 
     def post(self, request, *args, **kwargs):
-        if kwargs['key'] is not self.tournament.pref('email_hook_key'):
-            return HttpResponse(status=404) # 404: Not Found
+        if not self.tournament.pref('email_hook_key'):
+            return HttpResponse(status=403) # 403: Forbidden
+
+        if kwargs['key'] != self.tournament.pref('email_hook_key'):
```
czlee: @tienne-B heads-up, `!=` and `is not` aren't the same thing, and if you intend to compare values (rather than identities), you should use `!=` rather than `is not`. I needed to change this in order to get the webhook to be recognized by Tabbycat—I'm curious that you didn't find similarly?

philipbelesky:
All the above changes from @czlee seem to have been resolved and are easily verified, so going to merge this in.

PR #994, opened 2019-01-14
Add 4-team cap option to AIDA2016 Australs break model

viranweerasekera:


tabbycat/breakqual/aida.py
```diff
@@ -130,8 +130,16 @@ def reinsert_capped_teams(self):
 
 
 @register
-class Aida2016AustralsBreakGenerator(BaseAida2016BreakGenerator):
-    key = "aida-2016-australs"
+class Aida2016Australs3CapBreakGenerator(BaseAida2016BreakGenerator):
+    key = "aida-2016-australs-3cap"
```
tienne-B: `institution_cap = 3` here?

viranweerasekera: The base AIDA generator has the `institution_cap` set at 3 by default, so I only included an override variable in the new option (so the 3-team one should still inherit `institution_cap` from the base)

czlee:
Funnily enough, the institution cap actually used to be configurable when I first implemented it, and I removed it to simplify things because it was never actually customized in practice and thinking about break rules as well-defined rules made more sense. 🤷‍♀️

tabbycat/breakqual/aida.py
czlee: Sadly, changing this key actually breaks the data in the database: After migration, the value stored in the database will still be `aida-2016-australs`, and Tabbycat will no longer know what that means.

Options:
(1) Leave this as `aida-2016-australs`, or
(2) Write a data migration (as well as a schema migration) to convert this value.

I don't have a strong preference between these.

viranweerasekera:
Yeah, part of the new AIDA rules are a fun sliding scale of how many teams are spared from the cap (though for it to expand beyond 4 teams it would require a tournament of more than 120 teams, which seems unlikely)

tabbycat/breakqual/aida.py
```diff
@@ -130,8 +130,16 @@ def reinsert_capped_teams(self):
...
     def reinsert_capped_teams(self):
         self._reinsert_capped_teams(self.capped_teams)
+
+@register
+class Aida2016Australs4CapBreakGenerator(BaseAida2016BreakGenerator):
+    key = "aida-2016-australs-4cap"
+    institution_cap = 4
+    
+    def reinsert_capped_teams(self):
+        self._reinsert_capped_teams(self.capped_teams)
```
czlee: Since `reinsert_capped_teams` is exactly the same, it's probably better practice to create a new class in the hierarchy, say, `BaseAida2016AustralsBreakGenerator`, from which `Aida2016Australs3CapBreakGenerator` and `Aida2016Australs4CapBreakGenerator` both derive. 🙂

tabbycat/breakqual/aida.py
viranweerasekera: Ahh, got it - Given the old option is still the default for Australs tournaments I would suggest (1); will update!

viranweerasekera:
@czlee I've restored the original `aida-2016-australs` registry key for the default 3-cap option and have modified the class hierachy, let me know what you think :)

viranweerasekera:
So after some tinkering I've updated the AIDA 2019 break generator to change the cap dynamically based on break size - In light of that I'll close this request and open a new one with those details!

PR #997, opened 2019-01-14
AIDA 2019 Dynamic Institutional Cap

viranweerasekera:
Adds a break generator which computes the institutional cap based on the number of teams in the break (in line with the constitutional amendment passed by AIDA in 2018).


philipbelesky:
I'm going to merge this in and assume that @viranweerasekera has done the testing required to verify the method.

PR #998, opened 2019-01-17
Correct possible "No object" exception for private URL checkin

tienne-B:


philipbelesky:
Thanks for the fix! Let me know if this is linked to a sentry issue / should be pulled into hotfix.

tienne-B:
No, it isn't.

PR #999, opened 2019-01-18
Make additions/fixes to changelog

tienne-B:
- Added line about team draw emails
- Append text about skipping duplicates
- Fixed monospace text


philipbelesky:
Hey, thanks for the updates. What was the need for double backticks around inline code blocks? I haven't had any problem with using single ` which seem to match the [github markdown spec](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting).

tienne-B:
The problem is that we are not using Markdown, we’re using reStructuredText of which uses the double back-tick.

PR #1000, opened 2019-01-18
Add Unicode 11 emoji

viranweerasekera:


tienne-B:
Quadruple digits now 🎉

tienne-B:
Few quick comments:
* A migration has to be made as the list is used as the choices of the teams' code/emoji field.
* The space before the comma (after the boolean) is for alignment, so there shouldn't be a space with `False`s.

philipbelesky:
Hey thanks for adding these in. One consideration, however, is unciode adoption rates. Last time I checked Unicode versions are tied to browser or OS updates. Particularly on Android we can't assume that a majority of users are running the latest release, which means that there is a reasonable chance the emoji symbols won't render properly. If anything we've probably been too aggressive in the past with bumping the Unicode set. I'm not sure about what exact % should be the threshold though.

czlee:
Also, we have an emoji spreadsheet to make managing names and whether it's included in the random set easier, I just added both of you to it.

Issue #1003, opened 2019-01-22
Implement elimination ballots for two-team formats


tienne-B:
In two-team formats' elimination rounds, scores may not be attributed, similarly to BP elimination rounds. While BP a checkbox ballot form, a similar form should be added for other formats.

Consensus ballots would be simpler, as most of the BP elimination ballot logic could be re-used (it could also be a solution to #527. Voting ballots are trickier, and could require something akin to a new `TeamScoreByAdj` table to hold who voted how.

The form required could also be integrated with scores for #643.

czlee:
More generally, you can get entire tournaments where (even in inrounds) judges only give results, not speaker scores, and it's long been a low-priority idea to complete refactoring of the scoresheets system to allow for this. Just a heads-up, though—while result.py and scoresheet.py themselves are (iirc) refactored to support new ballot types easily, other parts of the code are not, which is why I've been holding off doing this, tied/low-point wins, and BP grand final results 🙂

I'm in principle fine with adding a `TeamScoreByAdj` model if it's necessary to record this kind of thing. If I remember right, it is—one thing though, we need to be clear in our minds which model fields are primary sources of information, and which are just calculated from other fields.

tienne-B:
Honestly, I still don't really understand `scoresheet.py` :( That's part of why I did not bother with it when adding `.as_dicts()` in the associated PR. Here, it might be to change `DeclaredWinnerMixin` to take a list of winning sides, and its use in two-team formats to raise an error if both are selected. There are classes for tied/low-point wins though, and so it does seem refactored.

I wonder what parts of the code do not readily support these alternative ballots?

To reduce the `TeamScoreByAdj` table, we could remove `margin` and `score` as they are based on the speaker scores of which could be compared by adjudicator separately. However, should the table be populated with per-adjudicator scoresheets?

tienne-B:
It's been a while, but I figured out `scoresheet.py`!

> I wonder what parts of the code [would] not readily support these alternative ballots?
@czlee: Wondering on this question. #1004 up to 1f9fde6 works well in my testing and live (or I don't remember where it fails). So the changes work well with elimination ballots. Any other considerations for formats requiring `TeamScoreByAdj`?

Should `TeamScoreByAdj` be populated for all voting formats, or just for formats requiring it? I'm leaning towards the former, but that would increase the number of rows used up.

czlee:
> It's been a while, but I figured out `scoresheet.py`!

[...]

Issue #643, opened 2018-06-22
Allow score draws in ballots


tienne-B:
Tabbycat should allow for inconsequential ballots (_i.e._ if not all judges ballots are draws or the panel is split) to have the same total scores for both sides (thus being a draw). It is most useful for tournaments that solely use speaker scores to determine standings. However, even if having a Win/Loss is preferable, there could be an option for who wins, like on the paper ballot, only taken into consideration in the event of a tie. The ballot margins could be 0, provided the debate margin is not.

PR #1004, opened 2019-01-22
Create elimination ballot form for non-BP formats

### LINKED ISSUE #527: Write framework for managing different result types
Currently the GUI provides no way to nominate a single team winning, as they do in final elimination rounds.

tienne-B:
This commit creates subclasses of elimination ballots so that a similar format to BP is used in two-team formats' elimination rounds, using the same data structure as for BP, which is limited to consensus ballots.

It also fixes a link bug in the old ballot-set form view.

Starts #643 & #1003
Fixes #527


tienne-B:
Should this new ballot form be as a new preference, as I assume there are some formats which use scored elimination rounds (or it would have already been implemented)?

philipbelesky:
I'll leave the code review to CZ, but jump in to say this should definitely be a preference. It might even be worth adding an alert to elimination ballots in two-team formats that highlights the existence of this preference. 

I don't know of any formats that don't explicitly ask for scored ballots. However it is pretty common not to care about scored ballots in finals rounds and so they often end up with 'fake scores'. This would provide an easy way around that.

The one problem that this then raises is: what happens if people switch preference over the course of the elimination rounds? I.E. if they loose the ballots for one elimination round and so set the preference to unscored then, but later revert to scored, will that cause issues? What if they change the setting within a round? We should definitely both test and account for those issues.

tienne-B:
There shouldn't be much effect if the preferences change mid-way. While the ballot views would show TIE without speakers or scores, draw generation would still work.

tienne-B:
A reply to https://github.com/TabbycatDebate/tabbycat/issues/1003#issuecomment-496633728 would be appreciated for the next (and current) steps in this PR.

tabbycat/results/forms.py
```diff
@@ -813,18 +813,19 @@ def scoresheets(self):
             yield sheet_dict
 
 
-class BPEliminationResultForm(BaseResultForm):
+class BaseEliminationResultForm(BaseResultForm):
```
czlee: [see overall comments]

tabbycat/results/forms.py
```diff
@@ -837,15 +838,31 @@ def clean(self):
...
         result.save()
+
+
+class BPEliminationResultForm(BaseEliminationResultForm):
+
+    result_type = BPEliminationDebateResult
+    num_advancing = 2
+
+
+class SingleBallotEliminationResultForm(BaseEliminationResultForm):
```
czlee: [see overall comments]

tabbycat/results/templates/ballot/ballot_set.html
```diff
@@ -1,5 +1,5 @@
-{% if pref.teams_in_debate == 'bp' and debate.round.is_break_round %}
-  {% include "ballot/bp_elimination_ballot.html" %}
+{% if pref.elim_ballot_nom and debate.round.is_break_round %}
+  {% include "ballot/elimination_ballot.html" %}
```
czlee: What happens if `pref.teams_in_debate == 'bp'` but `pref.elim_ballot_nom == False`? Does it show the standard BP ballot?

tabbycat/results/templates/public_ballot_set.html
```diff
@@ -72,29 +72,11 @@ <h4 class="card-title">
...
-                {% else %}
-                  {% blocktrans trimmed with name=team.team.short_name side=team.side %}
-                    <em>Total for {{ name }} ({{ side }})</em>
-                  {% endblocktrans %}
-                {% endif %}
-                <span class="badge badge-secondary float-right">
-                  {{ team.total }}
-                </span>
-              </li>
+              {% if pref.elim_ballot_nom and debate.round.is_break_round %}
```
czlee: ditto re what happens if `pref.teams_in_debate == 'bp' and not pref.elim_ballot_nom`.

tabbycat/results/views.py
```diff
@@ -269,9 +269,13 @@ def get_all_ballotsubs(self):
 
     def get_form_class(self):
         if self.tournament.pref('teams_in_debate') == 'bp' and self.debate.round.is_break_round:
+            if self.debate.round.next is None:
+                return SingleBallotEliminationResultForm
             return BPEliminationResultForm
         elif self.debate.round.ballots_per_debate == 'per-adj':
             return PerAdjudicatorBallotSetForm
+        elif self.debate.round.is_break_round and self.tournament.pref('elim_ballot_nom'):
```
czlee: I think there's an implicit `self.tournament.pref('teams_in_debate') == 'two'` in here (because the BP case is covered in the first `if` block) but it's better to be explicit if that's the intention.

tabbycat/utils/tables.py
```diff
@@ -870,8 +870,6 @@ def add_debate_ballot_link_column(self, debates, show_ballot=False):
             for debate in debates:
                 if not debate.confirmed_ballot:
                     ballot_links_data.append("")
-                elif self.tournament.pref('teams_in_debate') == 'bp' and debate.round.is_break_round:
-                    ballot_links_data.append("")
```
czlee: I'm also curious about why we feel the need to permit a scoresheet view for result-only consensus ballots, given that the only thing to know about them is the final result, which is already shown in the results table.

tabbycat/results/result.py
```diff
@@ -886,6 +927,40 @@ def advancing_teams(self):
...
+
+
+class BPEliminationDebateResult(BaseNominationDebateResult):
+    """For BP elimination rounds, where result is decided by consensus to advance two teams."""
+
+    is_voting = False
+    num_advancing = 2
+
+
+class ConsensusEliminationDebateResult(BaseNominationDebateResult):
```
czlee: [see overall comments]

czlee:
> I don't know of any formats that don't explicitly ask for scored ballots.

fwiw, a convenor of a Wellington high school league once asked me whether there was a way not to bother with speaker scores because their judges only choose the winner and loser, so there's at least one.

tienne-B:
This refactor is getting hard!

Thinking of a new pattern for DebateResults, Scoresheets, Forms, and Ballot Views.

This could be the correspondance between `DebateResult`, `Scoresheet`, and `ResultForm`.
* `BaseDebateResult`
  * `ResultOnlyScoresheet`
    * `(ByAdj/)WinnerSelectionBallotSetForm` (multiple-select)
* `DebateResultWithSpeakers`
  * `HighPointWinsRequiredScoresheet`
    * `(ByAdj)SpeakersBallotSetForm`
  * `TiedPointWinsAllowedScoresheet`
    * `(ByAdj)WinnerSelectionSpeakersBallotSetForm` (drop-down)
  * `LowPointWinsAllowedScoresheet`
    * `(ByAdj)WinnerSelectionSpeakersBallotSetForm`
* `VotingDebateResult`
* `VotingDebateResultWithSpeakers`
* `ConsensusDebateResult`
  * +`BPEliminationScoresheet`
    * `WinnerSelectionBallotSetForm`
* `ConsensusDebateResultWithSpeakers`
  * +`BPScoresheet`
    * `SpeakersBallotSetForm`

Each form has a `per-adj` subclass, but thoughts on this general structure? Only scoresheets have a 2,4-team differentiation, but is restricted to certain forms/results.

tienne-B:
Superseded by #1180.

PR #1008, opened 2019-01-28
Convert raw SQL to Django ORM

tienne-B:
This commit converts some of the uses of raw SQL into annotations in the linking of adjudicator speaker scores to speakers, counting the number of debateteams on a particular side, and in side count.

There is one last instance of using `RawSQL`: checking for team history in draw tables. That one is harder to convert.

### DIFF
--- a/tabbycat/participants/utils.py
+++ b/tabbycat/participants/utils.py
@@ -1,4 +1,4 @@
-from django.db.models.expressions import RawSQL
+from django.db.models import Count, Q
 
 from tournaments.models import Round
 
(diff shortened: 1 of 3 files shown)


tienne-B:
For checking for teams that have already seen each-other in a previous round, it'd be neat if it was side-independent, and so could be used in BP.

philipbelesky:
Hey, thanks for adding the change. SQL is not my area of expertise but I trust the conversion has been tested/evaluated vs the old method.

PR #1076, opened 2019-04-24
Start developing Debate XML import/export

tienne-B:
This commit starts creating an interface for the import/export and implementation of the Debate XML specification. Here, a side-wide view for importing an XML file is created, with a consumer link for asynchronous processing.

_More to come_


tabbycat/importer/templates/archive_importer.html
```diff
@@ -0,0 +1,36 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block head-title %}{% endblock %}
+{% block page-title %}{% trans "Import tournament archive" context "page title" %}{% endblock %}
+{% block nav %}{% endblock %}
```
tienne-B: Wants to get the tournament slug, even though the nav is empty (similarly to `test-email.html`)

tienne-B:
I'm a bit concerned I'm putting the cart before the horse here, as the schema has not been reviewed or discussed: https://github.com/TabbycatDebate/dta-spec/pull/4. @czlee, @philipbelesky: Mind looking it over (even just cursorily)? I should get back to this in a week or so.

philipbelesky:
Happy to look over it, although might need to wait a few days.

philipbelesky:
I've mostly commented over in the DTA repo, but just was curious about the desire to write an importer and whether you were planning to do that before/after/simultaneous with the exporter? It's certainly an interesting idea. Initially I pictured the point of DTA as enabling much more of an archival-storage/ data-vis purpose, but I guess having a Tabbycat importer means that even if the spec doesn't take off more broadly we would have a robust way of mitigating against data loss and/or combining tournaments into a single instance.

tienne-B:
An importer would be useful for me as a way to collate various tournaments previously on individual sites to a central site, of which is needed by SUCDI for end-of-year prizes. Having the importer would also be useful as TC already has some data visualisations and analyses (and result views) that would make the archive format more accessible.

I am creating the importer simultaneously with the exporter as I think having the importer gives a visible meaning for the exporter (although that isn't the main point); aiding with adoption.

tienne-B:
So exporting seems to be working well but with N+1 SQL inefficiencies that I'm finding hard to resolve in the `.add_rounds()` method. What prefetches on `round_set` should I do so that `DebateResult` does not fetch from the database for each debate?

In order to expose the debug toolbar to view the SQL queries, I add a typo in another method.

philipbelesky:
Hmm, I think to avoid `DebateResult` set being fetched the debates need to be gathered without any reference to their current status (and obviously the winners involved). Are the N+1 issues severe, or just inelegant? Given the scope of the exporter I wonder if it might be easier to accept some of them.

tienne-B:
Sorry, I don't understand on the points of "fetching without any reference of current status", nor what would be the threshold between inelegance and being severe. On the first, is it an inability to prefetch a table in various ways (_e.g._ from `speakerscore`: `debateteam__team__speaker` and `speaker`)?

tienne-B:
Importing a tournament seems to (normally) have fewer queries as there are a lot of `bulk_create()`s. However, I have not used them for institutions as they are shared between tournaments on an instance, nor used for speakers or adjudicators due to the one-to-one key relation with `Person`.

I'll get the importer to set preferences for the tournament with what it encounters when parsing.

Should I use `DebateResult`? I think it would be more efficient if I use the results querysets directly.

tabbycat/importer/templates/archive_importer.html
tienne-B: Resolved. Had to set `user_groups` in the view.

tienne-B:
django/django#10998 would make the code neater by making `_add_foreign_key()` obsolete, as all the objects would already have their IDs.

tienne-B:
@czlee: From #1199, should `.bulk_create()` be avoided here in favour of one-by-one?

czlee:
Umm… yeah. Unless it's painfully slow, like, more than a minute or so. But otherwise I think I'd rather display a progress bar and have it more robust to future changes in `.save()` functions. Otherwise we'd have no choice but to violate DRY (or promise never to use `.save()`, `pre_save` and `post_save`).

I'm open to discussion on this, though.

tienne-B:
Understood. I'll prepare a patch for performance comparison (after #1180 as it would improve result importation). Having a progress bar would be quite difficult as import does not use consumers.

czlee:
Hmm, should the import use consumers? Normally this is best practice for anything that takes a nontrivial amount of time?

(It's fine if you don't want to do this for a first implementation! Just thinking out loud.)

tienne-B:
This PR depends on #1180 for the use of `DebateResult` for XML. The refactor should be merged before this PR, as merging this will indirectly merge the other.

PR #1079, opened 2019-04-30
Improvements to webhook-id

tienne-B:
Fixes BACKEND-1F8.

### DIFF
--- a/tabbycat/notifications/views.py
+++ b/tabbycat/notifications/views.py
@@ -125,14 +125,16 @@ class EmailEventWebhookView(TournamentMixin, View):
 
         data = json.loads(request.body)
 
-        records = SentMessage.objects.filter(hook_id__in=[obj['hook-id'] for obj in data if 'hook-id' in obj])
+        # Ignore all objects without a Tabbycat-specified hook ID
+        data = [obj for obj in data if 'hook-id' in obj and obj['hook-id'] is not None]
+        records = SentMessage.objects.filter(hook_id__in=[obj['hook-id'] for obj in data])
         record_lookup = {smr.hook_id: smr.id for smr in records}
         statuses = []
 
         for obj in data:
             dt = datetime.fromtimestamp(obj['timestamp'])
             timestamp = timezone.make_aware(dt, timezone.utc)
-            email_id = record_lookup.get(obj.get('hook-id'))
+            email_id = record_lookup.get(obj['hook-id'], None)
             if email_id is None:
                 continue
             statuses.append(EmailStatus(email_id=email_id, timestamp=timestamp, event=obj['event'], data=obj))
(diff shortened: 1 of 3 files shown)


philipbelesky:
Hey thanks for submitting the PR. From following the other thread I assume this is good to go, but will leave it to @czlee to check quickly.

czlee:
It still happens on line 135 😕

https://github.com/TabbycatDebate/tabbycat/blob/9706c7326256e5cd5fffbe478ace5040ff8e1b38/tabbycat/notifications/views.py#L132-L138

tienne-B:
Just needed one more layer of `.get()`

tabbycat/notifications/views.py
```diff
@@ -125,14 +125,16 @@ def post(self, request, *args, **kwargs):
 
         data = json.loads(request.body)
 
-        records = SentMessage.objects.filter(hook_id__in=[obj['hook-id'] for obj in data if 'hook-id' in obj])
+        # Ignore all objects without a Tabbycat-specified hook ID
+        data = [obj for obj in data if 'hook-id' in obj and obj['hook-id'] is not None]
```
tienne-B: The hook-id is never None, as it is there only when specified. `and obj['hook-id'] is not None` is not needed.

tabbycat/notifications/views.py
```diff
@@ -125,14 +125,16 @@ def post(self, request, *args, **kwargs):
...
+        records = SentMessage.objects.filter(hook_id__in=[obj['hook-id'] for obj in data])
         record_lookup = {smr.hook_id: smr.id for smr in records}
         statuses = []
 
         for obj in data:
             dt = datetime.fromtimestamp(obj['timestamp'])
             timestamp = timezone.make_aware(dt, timezone.utc)
-            email_id = record_lookup.get(obj.get('hook-id'))
+            email_id = record_lookup.get(obj['hook-id'], None)
             if email_id is None:
```
tienne-B: Wouldn't need this check then, right?

tabbycat/notifications/views.py
czlee: I thought about it, and I realized we don't actually know this. It's coming from an external system, so you can't make any assumptions about the input (_i.e._, you should always check). It's technically an open URL, so I could even write a short script on my computer to POST to the webhook address with `{'hook-id': null}`, and then you'd have a `obj['hook-id']` that's `None`.

In this case it's not really a security risk, per se, because the consequence is pretty minor and the request can't modify important data anyway. But this is the kind of thing that I think about when interfacing with external systems.

tabbycat/notifications/views.py
czlee: Not the inner one! But still need the outer one: it could be a hook ID that wasn't in the database (so it'll still be in `data`, but won't be in `record_lookup`).

PR #1088, opened 2019-05-11
Fix typos in djangojs gettext

tienne-B:
"Contrainer" -> "Container" in filenames.

@philipbelesky: Should the typo in the translated files be corrected as well? Submitted as hotfix?


tienne-B:
You can selectively merge the first commit if the typo shouldn't be corrected in the other files.

philipbelesky:
Ha, I had no idea how that typo became so pervasive. I've forgotten whether those strings are comments or play a role as referencing, but seems best to merge both at the same time.

tienne-B:
Why'd you create a branch reverting this PR? (`revert-1088-i18n-js`)

philipbelesky:
Ah, that was a mis-click. I thought I had cancelled the request before it finished, but looks like it that was not the case (it didn't give any indication of what had happened though).

PR #1090, opened 2019-05-12
 Implement pull-up filter by draw strength

### LINKED ISSUE #1025: Implement pull-up based on opponent strength
Somewhat related to #484, as speaks could be used as well for this.

There could be a method to pull-up the team(s) with the lowest opponent strength (summed over the previous rounds) of the lower bracket. This could try to equalize the draw strength of each team, which could be seen as fairer.

tienne-B:
This commit implements a filter for pull-up eligibility as having the lowest draw strength of one's bracket. This allows to compensate for if a team had easier match-ups in previous rounds (even if power-paired) and so had an advantage.

By draw strength in this context, the average of team scores (by speaker score) from previous rounds is used. The average is used instead of the sum to prevent teams who had been absent a round from always being pulled up. Scores are used instead of wins as brackets are already based on wins.

Also included is a clarification in the standings doc page about draw strength.

Closes #1025.


philipbelesky:
Hey thanks for adding this, it would be super interesting to try out. I might leave the draw code for @czlee to review if he has the time.

Also a quick ping to @viranweerasekera who might be interested in trying this out.

viranweerasekera:
Ooh cool, I was just playing with something similar for the new AIDA minimising-repeat-pullups-with-limits rule - Just clarifying; these filters don't affect intermediate brackets, right?

tienne-B:
I have not taken intermediate brackets into account, but it shouldn't be affected as this is a pull-up restriction (analogously to "least pulled up")

viranweerasekera:
Got it, I suspected as much (the AIDA rule I referred to applies a restriction for pulling up into an intermediate bracket, but as far as I understood those two things are distinct concepts in here!)

tienne-B:
Thinking that `npullups` should also be a metric annotator to be consistent with the other metrics used for power-pairing. Will try it in a couple of days.

tienne-B:
This PR is ready

philipbelesky:
@czlee were you able to look at this?

czlee:
Not yet, sorry. It might be another few weeks before I get a chance, but I'll make a note to prioritize this when I next work on Tabbycat.

czlee:
…and I just realized this will now conflict with #1150. I forgot about this when compiling today's to-do list, I've now added it so I'll remember to do it first the next time I open my Tabbycat workspace.

tabbycat/draw/generator/powerpair.py
```diff
@@ -79,6 +81,9 @@ def __init__(self, *args, **kwargs):
         if self.options["pullup_restriction"] == "least_to_date":
             self.check_teams_for_attribute("npullups", checkfunc=lambda x: isinstance(x, int))
 
+        if self.options["pullup_restriction"] == "easy_draw":
+            self.check_teams_for_attribute("drawstrength", checkfunc=lambda x: isinstance(x, float))
```
czlee: When generating the draw, I get the error: "24 out of 24 teams don't have a 'drawstrength' attribute." Is this meant to be "`draw_strength_score`" instead of `"drawstrength"`?

tabbycat/draw/manager.py
```diff
@@ -187,11 +186,17 @@ def get_relevant_options(self):
...
-        generator = TeamStandingsGenerator(metrics, ('rank', 'subrank'), tiebreak="random")
+        extra_metrics = []
+
+        if self.round.tournament.pref('draw_pullup_restriction') == 'least_to_date':
+            extra_metrics.append('npullups')
+
+        if self.round.tournament.pref('draw_pullup_restriction') == 'easy_draw':
+            extra_metrics.append('draw_strength_score')
+
+        generator = TeamStandingsGenerator(metrics, ('rank', 'subrank'), extra_metrics=extra_metrics, tiebreak="random")
```
czlee: Actually, even when I try the fewest-pullups restriction, I get: "24 out of 24 teams don't have a 'npullups' attribute."

The standings generator doesn't add metrics as attributes, they'll still be in `standing.metrics['npullups']`. Did we want to add the attributes in lines 203-206 below?

tienne-B:
@czlee: Overwrote your rebasing; didn't notice until I tried to update the PR (and it wouldn't be necessary until #1150).

I didn't really think of tournaments publicizing the metric, but rather that it is a metric to describe teams more abstractly.

czlee:
Oh, the rebase was just to bring it onto develop, I hadn't resolved conflicts with #1150, dw.

I guess you just added a new feature by accident then 😉

tabbycat/options/preferences.py
```diff
@@ -242,6 +242,7 @@ class DrawPullupRestriction(ChoicePreference):
     choices = (
         ('none', _("No restriction")),
         ('least_to_date', _("Choose from teams who have been pulled up the fewest times so far")),
+        ('easy_draw', _("Choose from teams with a low draw strength from previous rounds"))
```
czlee: Should this say "lowest" rather than just "low"? It's not really clear what "low" means (bottom decile? bottom quartile? bottom half?). I think we're really choosing among those teams that have the minimum draw strength in the bracket, right?

(Also, since this is done by speaks, in most cases this will be a unique team, right?)

docs/features/draw-generation.rst
```diff
@@ -57,6 +57,7 @@ Options are set in the **Configuration** page as described in :ref:`starting a t
     - Whether and how to restrict pullups
     - - No restriction
       - Choose from teams who have been pulled up the fewest times so far
+      - Choose from teams with a low draw strength from previous rounds
```
czlee: [see comment in preferences.py]

tabbycat/draw/manager.py
```diff
@@ -184,17 +183,25 @@ def get_relevant_options(self):
...
 
         metrics = self.round.tournament.pref('team_standings_precedence')
-        generator = TeamStandingsGenerator(metrics, ('rank', 'subrank'), tiebreak="random")
+        extra_metrics = []
+
+        if self.round.tournament.pref('draw_pullup_restriction') == 'least_to_date':
+            extra_metrics.append('npullups')
+
+        if self.round.tournament.pref('draw_pullup_restriction') == 'easy_draw':
+            extra_metrics.append('draw_strength_score')
```
czlee: Also, if either of these two metrics are _in_ the team standings precedence, then the draw fails to generate for adding a metric twice.

I'm open to dropping this restriction on standings generation though, I can't quite remember why it's there other than to catch user error quickly. (Some metrics mean something different if repeated, namely who-beat-whom.)

czlee:
A more general remark: If you think about it, a more general version of this idea would be: "Choose from among the teams that are the lowest according to metric **A**".

Also, correct me if I'm wrong, but I believe if metric **A** happens to be the second metric in the team standings precedence and there are exactly two metrics in it, then this is **exactly** equivalent to "Pull up from bottom". (If there's a third metric, then it's roughly equivalent, but not exactly.) So the idea's mainly useful when the metric you want to use to pull up teams is different from the second metric in the team standings precedence.

I'm pointing this out because I want to be clear about what this idea does and why users might be interested in it, both for ourselves and as explained in the UI/docs.

tienne-B:
I'll defer to you on how to deal with the draw error with duplicated metrics. (Also, instead of `extra_metrics`, how about `pullup_metric`?)

Good generalization, and you are correct about the situation where it is equivalent to pullup-from-bottom. On the flipside, pull from _position_ would not apply with draw strength as there would normally be just one team caught in the filter. The number of pullups or draw strength is not normally the second metric (would just be extras I suppose) as they aren't really measures of team performance.

czlee:
^ that force-push is just a rebase onto develop

czlee:
^ force-push to apply a fixup to ff731755a82e2930e037710d3adbb32ddf86fc64, in which there was a mistake: https://travis-ci.org/TabbycatDebate/tabbycat/builds/564638228

czlee:
@tienne-B Please take a look at let me know if you object to anything?

tienne-B:
It all looks good!

PR #1092, opened 2019-05-22
Use half-in checkin style when team not complete

### LINKED ISSUE #1091: Team partial check-in indicator doesn't work with 2+ speaker-teams
The team's partial check-in indicator (where the box on the 'check-ins' page has green and grey stripes) displays when only 1 speaker in a team is checked-in, but then turns grey again when more than 1 speaker is checked-in but the entire team is not checked-in.

tienne-B:
Team checkin entities had striped green bars when only one speaker (out of two) were checked-in. This commit adds the style when not all speakers are, for the case where 2/3 are.

Fixes #1091.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,7 +28,8 @@ Change Log
     - A adjudicator/team conflicts with an adjudicator if *any* institution appears as an institutional conflict for both parties
 - When printing scoresheets you can now edit the motions display just on that printing page. This allows you to use placeholder motions in Tabbycat (in order to prevent leaks) while still producing ballots with the correct motions
 - Info-slides can now be split into paragraphs
-- Instiututional caps in breaks can be based on the number of teams in the break. Thanks to Viran Weerasekera for this feature!
+- Check-in pages now differentiate between teams with 1 and 2 checked-in people in two-team formats
+- Institutional caps in breaks can be based on the number of teams in the break. Thanks to Viran Weerasekera for this feature!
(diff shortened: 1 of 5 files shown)


philipbelesky:
Hey, thanks for adding this in. I've made one slightly tweak — will discuss this in the original issue to loop Viran in and potentially get more feedback.

CHANGELOG.rst
```diff
@@ -28,7 +28,8 @@ Change Log
     - A adjudicator/team conflicts with an adjudicator if *any* institution appears as an institutional conflict for both parties
 - When printing scoresheets you can now edit the motions display just on that printing page. This allows you to use placeholder motions in Tabbycat (in order to prevent leaks) while still producing ballots with the correct motions
 - Info-slides can now be split into paragraphs
-- Instiututional caps in breaks can be based on the number of teams in the break. Thanks to Viran Weerasekera for this feature!
+- Check-in pages now differentiate between teams with 1 and 2 checked-in people in two-team formats
```
tienne-B: Not just two-team formats!

philipbelesky: Oh, are there four-team formats with 3+ speakers per team?

philipbelesky:
@tienne-B hmm, in the formats with 3 speakers that I'm familiar with there is usually an informal or formal limit that says a speaker can deliver one iron-person speech. Is there a particular format where this assumption doesn't hold?

CHANGELOG.rst
tienne-B: The two comments were unrelated. For the changelog, I'm noting that there isn't an entry about the different styles in checkins.

tienne-B:
I do not have an example for formats with three speakers, but there are formats where team members may be substituted and things like that, and so the case where the team size is 4 or greater and there is partial checkins, one of the half-styles should be used (like a default)

PR #1094, opened 2019-05-24
Fix typo in djangojs.po

tienne-B:
The 'BallotsCell.vue' template in results has cell pluralized.


philipbelesky:
Good catch, thanks!

PR #1095, opened 2019-05-28
Mark 'scoresheet_list.html' for translation

tienne-B:
Some messages were forgotten.


philipbelesky:
Great, thanks for adding these!

PR #1089, opened 2019-05-12
Mark messages defined in Vue files for translation

tienne-B:
Check commit messages.

(There should also be a better way to handle these messages! ;) )

Closes #744.
~~Continues #1088 (therefore must be dealt with before this PR).~~ (Done)

PR #1107, opened 2019-06-08
Revert "Merge pull request #1089 from tienne-B/i18n/744"

tienne-B:
Merge commit 74cdb11ed8f48b9ad5d987a6e70a43a072e5ebb1.

Gettext does not work well with Vue.js, particularly with text parameters.


philipbelesky:
Hmm, I don't think it is worth reverting these wholesale, as many of the translations don't need string substitution. 

In terms of doing this within Vue, if I recall correctly we aren't really doing this through Vue at all, we just map the [Django JS translation catalogue](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#using-the-javascript-translation-catalog) to functions that are then [made available to Vue](https://github.com/TabbycatDebate/tabbycat/blob/4c73b1d8ada828cf50f361b6d336d50dbebc907e/tabbycat/templates/js-bundles/main.js#L223).

So those functions might provide what we need. Sorry that I didn't bring them to your attention earlier. I'll try and have a look at the diff and adapt that ones that use string substitution, but any help would be much appreciated.

tienne-B:
Being handled by using the `interpolate` JS function with `gettext`

PR #1111, opened 2019-06-10
Add blank init file to Django root

tienne-B:
This file was removed in 28bf81d01c when Celery was removed. However, it is necessary for the path to the apps be found, such as for tests. This commit re-adds the file, but blank.


czlee:
Oh, I did this already, 51b1f19b474534f6b2a5e56a35514d0ca48a196d

PR #1118, opened 2019-06-18
Specify no relevant options for manual draw

tienne-B:
The manual draw-type did not override `get_relevant_options()` and so included some options when in two-team formats, causing a crash. This commit makes the draw manager return an empty list in that case.

@philipbelesky: Neither "Edit Sides/Matchups" (legacy and new interfaces) work when no teams are allocated, but fail differently. The legacy interface includes the draggable draw table but no teams, while its the opposite in the new interface (teams available but no table).


philipbelesky:
Great, thanks for the fix. The Edit Sides UI is definitely on my TODO list — its one of the things current blocking 2.3 so I hope to get to it soon.

PR #1119, opened 2019-06-22
Adjallocation: Fix zero-division exception

tienne-B:
When normalizing adjudicator scores, a division by zero error can occur if the range is of 0.

### DIFF
--- a/tabbycat/adjallocation/allocators/hungarian.py
+++ b/tabbycat/adjallocation/allocators/hungarian.py
@@ -40,7 +40,10 @@ class BaseHungarianAllocator(BaseAdjudicatorAllocator):
 
         for adj in adjudicators:
             adj._weighted_score = adj.weighted_score(self.feedback_weight)  # used in min_voting_score filter
-            adj._normalized_score = (adj._weighted_score - score_min) / score_range * 5  # to 0-5 range
+            try:
+                adj._normalized_score = (adj._weighted_score - score_min) / score_range * 5  # to 0-5 range
+            except ZeroDivisionError:
+                adj._normalized_score = 0.0
 
         ntoolarge = [adj._normalized_score > 5.0 for adj in adjudicators].count(True)
         if ntoolarge > 0:


philipbelesky:
Great, thanks for following up the report!

czlee:
Thanks, @tienne-B! This is definitely a case that I overlooked.

Thought: Users never see normalized scores, it's just an algorithmic convenience so that "matching" debate importances (always on a 1-5 scale) and adjudicator scores makes sense. Moreover, when something is outside the range, it won't force it within the range, it'll just keep going with out-of-range scores, come what may. (It won't crash.) Then this fix would have the effect of nullifying any effect of adjudicator scores whenever the minimum and maximum adj scores are the same, so adj allocations will be random (except for conflict avoidance). But this behavior occurs only when the range is exactly zero, not even when the range is (say) 0.01, and there's no obvious way the user has to detect what it's doing or why.

Here are some ideas:
1. Warn users somewhere appropriate if `min_score == max_score` (in combination with any other idea, including what this PR did).
2. Fall back to `adj._weighted_score` rather than `0.0`. That way adj scores still do _something_ in the allocation. But it's still a singular case: the behavior at 0.0 still wouldn't be a limiting case of very small adj score ranges.
3. Implement the limiting case by forcing all importances to 0.0. I'm not a fan of this approach though, I think it introduces too many code paths for a pathological case.
4. Refuse to run the allocation if `min_score == max_score`, because there's no way to do it sensibly. But then, some users might want exactly the behavior this PR implements: maybe they _want_ adj scores to be ignored and allocations to be consequentially random, and configure `min_score == max_score` in order to enforce it.

Thoughts?

tienne-B:
As changing the range to be 0 requires a conscious decision, and having a range of 0 intuitively means that it shouldn't be taken into account, this is the correct behaviour. Scores should be ignored in this case. A warning could be added to the allocations page.

czlee:
I don't think that's intuitive, but nothing seems intuitive to me, so this is definitely defensible 😊 The warning in #1122 should do the trick.

PR #1124, opened 2019-06-23
Add strings to prohibited tournament slugs

tienne-B:
For the notifications test page, summernote, and XML import, forecasting its use. Also removed the participant x-tournament string from the list.


tienne-B:
BTW, are shifts part of the divisions setup, can't find any mention of the feature.

philipbelesky:
Send me a PM if you'd like and I can send you the documentation for them. They were very much a one-off feature for the divisions client and I don't think anyone else uses them. At this point I think we should either plan to remove that feature (along with the rest of divisions) or at least reimplement it to properly work across all formats (although arguably that would be feature bloat).

PR #1139, opened 2019-07-01
Revert standings email template to use "wins"

tienne-B:
Instead of "points," which is more applicable in BP. It was changed incompletely in 270230c8aa6c1973f04d6d436f681478652ddfa1. The subject length change is kept but expanded to the BP preset.

A hotfix patch could be provided to include the missing `ROUND` variable. The variable is present in `develop` with the refactored notification system.


philipbelesky:
Sorry I didn't realise there was a separate class it switches to for BP messages, my bad.

PR #1140, opened 2019-07-01
Un-mark error strings for translation

tienne-B:
A couple strings that were marked for translation never surface to users as they are passed into an exception. Having them marked sends the translated content to Sentry. (_e.g._ BACKEND-1BK)

### DIFF
--- a/tabbycat/adjallocation/views.py
+++ b/tabbycat/adjallocation/views.py
@@ -248,11 +248,11 @@ class LegacyCreateAutoAllocation(LogActionMixin, LegacyAdjudicatorAllocationMixi
         round = self.round
         self.log_action()
         if round.draw_status == Round.STATUS_RELEASED:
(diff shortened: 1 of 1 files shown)


czlee:
I think this does surface to users, messages in a `BadJsonRequestError` are passed through a JSON response back to the front-end, which in turn displays it in a danger alert:
https://github.com/TabbycatDebate/tabbycat/blob/d9dce218027866a474b52c70de04faa47985ad6e/tabbycat/utils/views.py#L87-L97

tienne-B:
Might have been mistaken; similar translatable strings are in `adjallocation/consumers.py:58` and `venues/consumers.py:18`. Thought to remove the translatability here to be in line with `venues/views.py:83`

czlee:
Yeah, it looks to me like venues/views.py:83 is the one that's wrong. I wouldn't bother with this though, they're legacy views due to be deprecated once @philipbelesky's satisfied they're no longer needed.

PR #1145, opened 2019-07-12
Translation: Traduit plus de français

tienne-B:
Try to avoid having bad merge conflicts with the Crowdin initiative; these are strings I have been sitting on.

Fun fact, "translation" is a false friend in french, which means to move something (such as a saint's bone) from one place to another.


philipbelesky:
Yea, I'm wasn't planning on merging any of the Crowdin ones in until I get a handle on whether they are merging nicely. We should probably run a general update to the `.po` files before then anyway — I don't think that's been done for awhile.

PR #1149, opened 2019-07-21
Permit empty standings precedence and warn user accordingly

### LINKED ISSUE #1108: Draw generator fails when no metrics are provided
I hit an internal server error while trying to load this URL:
https://v-potiguar.herokuapp.com/vpotiguar2019/admin/draw/round/4/create/

Please describe what you were doing when the error happened, along with any other useful information:

Whenever try to generate a draw to round 4 on my tournment, this error displays:
error

czlee:
As discussed in #1108.
ping @ArthurPaiva if you're interested (ignore this if not)

Closes #1108

### DIFF
--- a/tabbycat/availability/templates/availability_index.html
+++ b/tabbycat/availability/templates/availability_index.html
@@ -194,6 +194,20 @@
     </div>
   {% endif %}
 
+  {% if pref.team_standings_precedence|length == 0 %}
+    {% tournamenturl 'options-tournament-section' section='standings' as standings_config_url %}
+    {% blocktrans trimmed asvar message %}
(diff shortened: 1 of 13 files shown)


philipbelesky:
Looks good! I wondered if it was worth adding a warning to the options page itself, but I figure its probably enough to have it pre-draw and in the standings pages which should surface things on the critical path anyway.

czlee:
@philipbelesky I'd prefer that this be kept off the release branch, unless there's a need for this to be in 2.3? Although on the surface it's minor, it changes a lot of things, so I'd like the code to be put through its paces first.

PR #1150, opened 2019-07-21
Rewrite draw strength annotator to query for opponents directly

### LINKED ISSUE #1071: Draw strength doesn't work for BP
The current method of computing draw strength relies on `populate_opponents`, which assumes that each debate-team only has one opponent. It'd be nice if it could be updated to grab all opponents. Since the existing code doesn't use the opponent for anything other than its team ID, there should be a single (complicated) query that can grab this directly (then look it up in `points_queryset` as defined in the existing code, as it does currently).

https://github.com/TabbycatDebate/tabbycat/blob/fbdabf141d9e39c049fb3234159296abba287437/tabbycat/standings/teams.py#L214-L231

Possible gotcha: If A faces B twice, then they should count for each other twice. In other words, don't use `.distinct()`, and the queryset should be by debate-team, not by team.

I think it's not ideal to change `populate_opponents` itself, since that's often used in contexts specific to two-team formats, and it'd be annoying to have to always deal with a list when most of the time it's just one team.

czlee:
Previously, this relied on the populate_opponents() method in draw/prefetch.py, which is designed to produce a single opponent per DebateTeam given to it (mainly for use in two-team formats). This doesn't work in BP (see #1071). Furthermore, draw strength doesn't require information on individual debate teams, it just requires a list of all opponents seen by a team, which can be queried more directly.

Fixes #1071

### DIFF
--- a/tabbycat/standings/metrics.py
+++ b/tabbycat/standings/metrics.py
@@ -100,7 +100,7 @@ class QuerySetMetricAnnotator(BaseMetricAnnotator):
     def get_annotated_queryset(self, queryset, column_name, round=None):
         """Returns a QuerySet annotated with the metric given."""
         annotation = self.get_annotation(round=round)
-        logger.info("Annotation in %s: %s", self.__class__.__name__, str(annotation.as_sql))
+        logger.info("Annotation in %s: %s", self.__class__.__name__, str(annotation))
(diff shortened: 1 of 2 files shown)


czlee:
Pinging @mpopst, if you're interested

tienne-B:
Don't worry about the conflict with #1090; quite minor

PR #1151, opened 2019-07-25
Admin Nav: 'Room' -> 'Venue'

tienne-B:


czlee:
This needn't affect this PR, but this seems like a good place to mention that I've actually thought about changing English terminology wholesale from "venue" to "room", because "room" is less ambiguous ("venue" can be the entire host complex) and it would bring it in line with other languages.

Don't do this now though, I think we want to do it all in one go only after we decide to do it, and I don't want to do it for LaPerm.

tienne-B:
Just to note another potential change in this vein: "test score" https://github.com/TabbycatDebate/tabbycat/pull/762#issuecomment-413969389

PR #1154, opened 2019-07-28
Interpolate remaining i18n Vue strings

tienne-B:
Use the tct method to combine interpolation and gettext for overlooked strings in Vue files.


philipbelesky:
Great, thanks!

Issue #1117, opened 2019-06-18
Implement non division-based round-robin draw generator


tienne-B:
From https://www.facebook.com/groups/1681761898801915/permalink/2214068098904623/

The `round-robin` draw-type is reliant on divisions to work, but could be used for regular tournaments. This generalisation should be quite easy, as teams need not be excluded, nor separate brackets created for this case.

However, a history check should be added to prevent repeat pairings.

Issue #517, opened 2017-10-29
Make adjudicator allocations hidden until released


czlee:
This is minor, but: If you want to allocate sides while the adj core is still allocating panels, and show the side updates in GA, it'll also show the work in progress on adj allocation, essentially live (every time you refersh the page). It'd be a small but nice improvement to be able to hide them.

Now that I think about it, I'm picking that there are probably other cases where you want to release team pairings without releasing adjudicator allocations.

I wouldn't do this if it's very hard, but maybe a system similar to draw and motion release might be useful? But we don't want to break the standard use case when they're released simultaneously, so there might be some UI questions.

philipbelesky:
Which page is being used to show the side updates? The public draw page; or the admin one? 

Potentially we could add a button on the front end to hide the adjudicators column, but I wouldn't be super enthusiastic about adding another step to the critical path for a niche use case.

czlee:
Umm, I think I was using the admin one, but anything that's up in GA you want to be available publicly anyway?

I don't think this should be on the critical path (in fact, I think it should not be), they should probably be released (or just "visible") by default and maybe we have a button on the adj allocation page to hide it or something.

philipbelesky:
Yea, or if its only desirable for admin use we could add another "Draw Display By..." option that excludes the adjudicators column by default. That's probably safest and simplest as there's no need for a button to hide things post-load.

czlee:
Hmm, maybe. I was thinking it would also be nice to be able to show outround pairings without judge allocations, as soon as it's obvious what they would be.

philipbelesky:
If it's just for out rounds it could almost be a brackets-style diagram of the elimination stages? But that might be a bit overkill.

czlee:
I think a bracket diagram would be cool, but different. If your objective is to figure out which venue you want to go to and who's in which position, bracket diagrams are a little harder to parse. (If your objective is to see when teams will first meet, bracket diagrams are great.)

0zlw:
Hi all, I had a go with working trying out the "hide adjudicators" feature in v2.4.0A over the weekend and generally found the fix didn't satisfy how we were hoping to using Tabbycat. 

I think it's somewhat common for tournaments to want to release the draw ahead of adjudicator allocations and then use prep time to complete adjudicator allocations. This is a particularly useful tool when trying to run smaller tournaments to time (i.e. non majors) or where judge allocations and conflicts are less important.

[...]

Issue #459, opened 2017-05-10
Distribute 'league' options into the wider system


philipbelesky:
So just thinking about how to manage WADL in the future (after accidentally updating them) and I think its worth either abolishing the 'league' section of settings, and/or either creating a dedicated fork or an environment variable that hides them from the mainstream version.

That is to say features that are WADL-related but have broader utility should be more fully supported and redistributed into the traditional settings areas (ie sharing venues; adjudicator notes; cross draw pages) whereas the settings and/or views that are highly specific to WADL (allocation confirmations, the printable room sheets, etc) should be more explicitly cordoned off both in the UI and potentially in the code base (this could mean a dedicated branch, or perhaps confining them into a specific django app). 

Obviously there are a lot of grey areas, such as the concept of divisions, that need to be decided on. In the case of divisions I think they need to be either cordoned off, or more explicitly integrated by way of having test cases and ensuring they work with the range of draw options rather than assuming the standard round-robin system. 

The ETA for this would be sometime much later in the year. It might also be worth be done in conjunction with making a more consolidated push (on my behalf) to seeing if other large-scale public leagues are interested in using the software.

philipbelesky:
Basically redundant given #1155; can be re-opened if adding more specific features.

PR #1155, opened 2019-07-29
Deprecate and remove league/division code

### LINKED ISSUE #487: SecondElimination DrawGenerator doesn't account for forfeits
Currently forfeit results are not recorded as wins, so will prevent the round advancing as there are no results.

As with other issues under the Cygent milestone these can be held off on until we confirm if/when Cygnet is brought up to date with master.

### LINKED ISSUE #488: Divisions allocator probably needs to be reworked
The auto-allocator for Divisions should be broken given the many updates to Venue Categories. If Cygnet is brought up to date with the main branch this will need to be check/verified.

### LINKED ISSUE #489: AssignMotionsView isn’t filtering for only that tournament’s divisions
I believe that on master the AssignMotionsView show motions for divisions/motions from all tournament's rather than just the current one. If/when cygnet is brought up to date with master this should be checked along with the ModelAssignForm implementation.

### LINKED ISSUE #490: Better prevent regressions in Cygnet functionality
Cygnet's core functionality often breaks as a result of core upstream changes. If/when it merges back from master there should be basic unit or integration tests to better check against this.

tienne-B:
This commit removes all of the league/division features with database migrations to purge associated fields. The divisions app is still present for the deletion migration. Templates and views were also removed with references where unsupported. Certain features could be "re-incarnated" if wanted.

The main user of these features is still on Tabbycat 1.4.6 and these features have not really been maintained since. The WADL preset has been modified to not include newly-removed settings.

Divisions are groups of teams whose matchups are bound between them. They also share a particular venue category.

A note will be added to the changelog about this change, although not in this commit as the version is not decided.

The features that were removed were:

- Shared Adjudicators/Venues
  Adjudicators and venues must be related to a tournament, and cannot be shared between them.
- Duplicate Adjudicators
  Adjudicators cannot be associated with more than one debate per round.
- Person notes
  Person notes was an arbitrary TextField available for text from the tab about an individual; only implemented for adjudicators.
- Western Australian Debating League Break Criteria
  WADL breaks are based on the winners/losers of divisions.
  - "WADL division winners first" takes the winners of divisions then losers if unfilled.
  - "WADL division winners guaranteed" sorts by rank.
- Creation of Bye teams
  Bye teams are placeholder teams for when there is an uneven number of teams, and so one team steps out, to debate with another bye team later. Teams can still be marked as "bye" to exclude them from standings. Not sure if bye _debates_ were implemented.
- Round Robin draw generator (add feature: #1117)
  A round-robin draw is where each team (in a division in this case) eventually debates all others.
- Postponed debates
  Debates can no longer be postponed, but can still be marked as such, without effect. Postponing debates meant that rounds could be advanced without postponed debates' results.
- Adjudicator scheduling
  A method to mark an adjudicator as available only at certain times. Superseded with round availabilities.
- Debate scheduling
  Specific debates cannot be assigned a time; only rounds.
- Venue sheets
  Printed sheets for venues/venue categories. For divisions, shows the division/debate assigned to the venue(s).
- Multi-tournament draw pages
  Pages showing the draw between multiple tournaments have been removed. By institutions, venues, and teams
- Flagging motions
  Motions cannot be specifically flagged (as a field) as being contentious.
- Debate forfeits (moots #487)
  Debate-teams cannot be explicitly set as having forfeit a debate. The "2/1/0 points" metric has been removed consequently.
- Hidden adjudicators (see also: #517)
  Adjudicators will always be shown in public views of a draw.

Which features should be re-added is up for discussion, and some could do well without necessarily divisions.

The `Cygnet` milestone and branch seem to be about divisions, and so they could be deleted/closed as a result.

Main task #459
Closes #487
Closes #488
Closes #489
Closes #490


philipbelesky:
Great, thanks for figuring out and pruning all of these. I've started a release branch but might try and fix the Docker issues on `develop` before merging this in — given the scope of this I'd like to leave it for 2.4. I can then remove all the old versions of the Vue pages also.

tienne-B:
I've hidden some strings only used by divisions from Crowdin (not deleted)

philipbelesky:
Merged! Thanks again, always great to see that much code being culled.

PR #1167, opened 2019-08-02
Add pluralization for adj score < 0 warning

tienne-B:
In the hungarian adj allocator, there are warning messages if there are adjudicators with normalized scores greater than 5.0 or lower than 0.0. The message for greater than 5 is localizable, but the latter was not.

This patch adds pluralization and fixes the type on the key message.

Closes BACKEND-238


philipbelesky:
Great, thanks for the very timely fix!

PR #1168, opened 2019-08-02
Send version of Tabbycat to Sentry

tienne-B:
The release version was no longer being sent to Sentry on errors after the migration to the newer SDK. This is due to it searching for a `SENTRY_RELEASE` env variable. This change specifies the version in the initializer.

### DIFF
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -6,6 +6,8 @@ import sentry_sdk
 from sentry_sdk.integrations.logging import LoggingIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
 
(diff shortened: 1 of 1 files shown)


tabbycat/settings/heroku.py
```diff
@@ -124,6 +126,7 @@
             LoggingIntegration(event_level=logging.WARNING),
         ],
         send_default_pii=True,
+        release=TABBYCAT_VERSION
```
czlee: Being picky, can we put a comma after this? My convention is to put trailing commas when the closing bracket is on a new line, that way if you want to add a new argument later (as you have here) or rearrange lines, it doesn't affect the other lines.

PR #1169, opened 2019-08-03
CHANGELOG: Correct 'three-team' to 'four-team'

tienne-B:
Is it okay to retroactively correct a changelog entry after its release? Also, ~1 1/2 years later?


czlee:
lol, I don't have a problem with this. @philipbelesky?

philipbelesky:
The only constant is change(log)!

PR #1170, opened 2019-08-05
Remove win-psycopg requirement from docs

### LINKED ISSUE #423: Evaluate/test 'standard' Windows install
Once Python 3.6.1 is released, see if the win-psycopg2 issue is resolved, and update installation instructions accordingly.

tienne-B:
The psycopg-binary requirement lists as having a passing build on
Windows and the link to the Windows version doesn't seem to work.

Continuing the Python version bump from 223c98327cf7134c604393b600ac7f7f98211e5f.

Closes #423.

### DIFF
--- a/docs/install/windows.rst
+++ b/docs/install/windows.rst
@@ -186,15 +186,8 @@ e. Install Tabbycat's requirements.
   If you installed **32-bit Python**::
 
     > python -m pip install --upgrade pip
-    > easy_install http://www.stickpeople.com/projects/python/win-psycopg/2.6.2/psycopg2-2.6.2.win32-py3.6-pg9.6.1-release.exe
     > pip install -r .\config\requirements_core.txt
     > npm install
-
-  If you're using a version of **Python other than 3.6**, replace the URL in the
-  second line with the appropriate link from the
-  `win-psycopg page <http://www.stickpeople.com/projects/python/win-psycopg/>`_.
-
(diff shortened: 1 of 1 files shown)


czlee:
Has anyone tried doing a local installation on Windows using this method? I've actually been toying with the idea of just dropping support for Python for Windows altogether, and going all in on the Ubuntu on Windows system for Windows users, since that's the environment I actually use, and I don't really want to maintain a working Windows local installation on my computer so I can keep up with what's still working.

I'm not sure that Travis is enough evidence for me, I need someone to confirm that their Windows installation is working for them.

philipbelesky:
It would probably be tested as part of a 3.8 release. However, I'd be happy to drop support for that installation method. The subsystem seems equally easy/hard to setup but would have the benefit of simplifying the task of maintaining windows support.

czlee:
I guess we could ask in the Facebook group if anyone still uses a Windows local installation, and that we're looking to drop support if not? Since I use the subsystem all the time, it's a lot easier for me to be a contact point for people having difficulty with it.

tienne-B:
FB thread: https://www.facebook.com/groups/tabbycat.debate/permalink/1294891527345468/

philipbelesky:
I've tested the standard windows install instructions as-is without installing psycopg and there don't seem to be any problems connecting to postgres, so will merge this.

PR #1172, opened 2019-08-06
Notifications Status: Add cell value if recipient is null

tienne-B:
Recipients to a SentMessage can be null, which may cause a key error when trying to access their attributes. This happens in the email status view. A check is added for this case.

Closes BACKEND-24J

### DIFF
--- a/tabbycat/notifications/models.py
+++ b/tabbycat/notifications/models.py
@@ -40,7 +40,12 @@ class SentMessage(models.Model):
         verbose_name_plural = _("sent messages")
 
     def __str__(self):
-        return "%s: %s" % (self.recipient.name, self.notification.get_event_display())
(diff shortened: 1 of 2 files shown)


czlee:
In what situations does this practically show up? Emails always have a recipient when sent, so I'd expect this to show up when a participant gets delete from the database, right? Asking because it's a bit weird for a recipient of an email to be "not applicable", so using N/A as the label might be confusing.

czlee:
Thanks for putting this in, @tienne-B! I ended up doing it a bit differently (though roughly the same, just rephrased) in 86c719e986b5e6e0af55c9261b72facf5c30766e.

As a heads-up, the implementation in the PR will omit the recipient column altogether for emails where no status report has been received.

Issue #155, opened 2015-03-03
Drop requirement for teams to have institutions


czlee:
Nice-to-have, also probably a lot of work as I think current code base assumes all teams have institutions.

philipbelesky:
Just a note/thought on this: cross institutional teams in many cases still need conflicts pointing to their original teams. So if this ever is refactored, the option to have multiple Institutions is something to consider.

Alternately, `Institution` could be a property of each speaker, and each speaker could have multiple institutions (or an institutional history ala tabbie).

czlee:
Hmm, speaker institutions might not be a bad idea. We'd just have to make all institution fields optional, which would make for fun times writing any code that handles institutions.

czlee:
Even if we do speaker institutions I suppose if we're being strictly correct we should allow for more than one? Though in the status quo I guess we expect users to make all these things as conflicts. In fact, doesn't our code separate the two entirely, so that (primary) institutional conflicts only count if they're actually marked as `AdjudicatorInstitutionConflict` objects?

philipbelesky:
Mental note: this should extend to adjudicators also

philipbelesky:
@czlee how viable do you think this is to implement (the dropping institutions from adjs/teams part; not the individual speaker institutions part). While this is less of a strict issue with BP draws, it is still quite annoying in the context of IVs and things where institutions are not required and are a hassle for data entry.

If a structural solution is too challenging for now, I wonder if there could be a data entry convenience during the importer that, when toggled, automatically creates a new institution just for that team.

czlee:
I can't see any structural challenges. We'd just have to modify code that would crash or malfunction if an adjudicator's institution is `None`. One approach could be just to do it, and then fix pages that crash as we encounter them.

czlee:
Oh, actually, for teams I'd have to examine the uniqueness constraints, since I think uniqueness isn't checked on null values. And possibly the `short_name`/`long_name` construction code. But I'm sure it can be done, somehow.

czlee:
Non-exhaustive list of things that will break, that I propose we ban rather than modify:
- One-up-one-down. Since it just compares institutions, two teams both without institutions will look like they have the same institution.
- AIDA institution caps. Again, since it just compares institutions, all teams without institutions will look like they have the same institution.

I guess we'll need to add error messages or something like that.

philipbelesky:
[...]

PR #1176, opened 2019-08-09
Add checks for None institutions in up-down swap

tienne-B:
When swapping teams due to being from the same institution, unaffiliated teams (thus with None) were barred from meeting. This commit adds checks for this case.


czlee:
Relates to #155 and the note in documentation in docs/features/draw-generation.rst added in c67ade1db071c20b25740819c9c4627b3f95994b.

PR #1179, opened 2019-08-10
Results: Fix link to overview page

tienne-B:
The old ballot entry page did not have `for_admin` set and so the "Back" button redirected to the assistant page. This commit sets `for_admin` as `True` for all views using `AdministratorMixin`.


philipbelesky:
Great, thanks for the fix!

Issue #922, opened 2018-11-17
populate_results tries to populate scores from trainee scoresheets


czlee:
In the traceback on Sentry, the adjudicator in question is actually a trainee in this debate. So I think when `populate_results` populates the scoresheets, it (correctly) excludes the trainee adjudicator from the `DebateResult` object, but because the trainee's `SpeakerScoreByAdj` instances are still in the database, it tries to add those scores, which causes an error.

The tempting thing to do would be add `.exclude(debate_adjudicator__adjudicator__type=DebateAdjudicator.TYPE_TRAINEE)` to the `SpeakerScoreByAdj` queryset in `populate_results` (in `results/prefetch.py`. But this wouldn't change the underlying data inconsistency: trainee adjudicators aren't meant to have `SpeakerScoreByAdj` objects.

This then raises the question: How did the database get into this state? It's possible that the adjudicator was originally a panellist and therefore had a scoresheet, but then got demoted to trainee. Then, if the result was entered before the demotion, the scores from the original results would hang around the database. If the trainee had been removed, cascade deletions would have deleted the `SpeakerScoreByAdj`s, but because it was just a change of `DebateAdjudicator.type`, the `DebateAdjudicator` was never deleted.

It's still always possible to edit the database to get the database to this state, though, so two things should happen:
1) This error should be caught and should fail more gracefully. I think this isn't trivial: We don't really want to delete data without its being the direct and timely result of a deliberate user action, but at the same time if you discover inconsistent data, we should prompt the user to fix it.
2) When an adjudicator is demoted to trainee, associated `SpeakerScoreByAdj` objects should be deleted.

https://sentry.io/tabbycat-ig/backend/issues/766941290/

```
KeyError: <Adjudicator: [name redacted]>
(18 additional frame(s) were not displayed)
...
  File "utils/views.py", line 127, in get_tables
    return [self.get_table()]
  File "participants/views.py", line 228, in get_table
    populate_confirmed_ballots(debates, motions=True, results=True)
  File "results/prefetch.py", line 66, in populate_confirmed_ballots
    populate_results(confirmed_ballots)
  File "results/prefetch.py", line 155, in populate_results
    ssba.position, ssba.score)
  File "results/result.py", line 540, in set_score
    scoresheet = self.scoresheets[adjudicator]

KeyError: <Adjudicator: [name redacted]>
```

czlee:
[...]

PR #1180, opened 2019-08-10
Refactor DebateResult and associated classes

### LINKED ISSUE #527: Write framework for managing different result types

### LINKED ISSUE #643: Allow score draws in ballots

### LINKED ISSUE #1003: Implement elimination ballots for two-team formats

tienne-B:
This branch is the work in refactoring `DebateResult`, `Scoresheet`, and `ResultForm` for technical debt and allowing for new types of results. See commit messages for more details.

Supersedes #1004 

Goal:
Fixes #527 
Fixes #643 
Fixes #1003


tabbycat/results/scoresheet.py
```diff
@@ -72,93 +75,102 @@ def identical(self, other):
...
-class DeclaredWinnerMixin:
-    """Provides functionality for explicit declaration of winner.  Only makes
-    sense for two-team scoresheets."""
+class DeclaredWinnersMixin:
+    """Provides functionality for explicit declaration of winner(s)."""
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.declared_winner = None
+        self.declared_winners = []
```
tienne-B: Maybe it'd be best if it were already a set rather than a list.

tabbycat/options/preferences.py
```diff
@@ -461,6 +461,21 @@ class BallotsPerDebateElimination(ChoicePreference):
     default = 'per-adj'
 
 
+@tournament_preferences_registry.register
+class BallotMustConfirmWinner(ChoicePreference):
+    help_text = _("Whether adjudicator(s) must select the winning team in their ballot, and how it should be treated. Note: Not supported in BP.")
```
tienne-B: I didn't really know how to word this preference...

tienne-B:
Ready for review!

tabbycat/notifications/utils.py
```diff
@@ -20,7 +20,7 @@
 from options.utils import use_team_code_names
 from participants.models import Person
 from participants.prefetch import populate_win_counts
-from results.result import BaseConsensusDebateResultWithSpeakers, DebateResult, VotingDebateResult
+from results.result import ConsensusDebateResultWithSpeakers, DebateResult, DebateResultByAdjudicatorWithScores
```
czlee: This should be `ConsensusDebateResultWithScores`, right?

(Will review the rest later, but I noticed this so thought I'd check in on it quickly.)

tabbycat/results/result.py
```diff
@@ -519,58 +427,45 @@ def save(self):
...
 
-    def set_declared_winner(self, adjudicator, declared_winner):
-        """Not currently used, in place for future implementation of declared winners."""
-        try:
-            return self.scoresheets[adjudicator].set_declared_winner(declared_winner)
-        except AttributeError:
-            logger.exception("Tried to set declared winner, but scoresheet doesn't do declared winners. "
-                    "self.takes_declared_winners is %s", self.takes_declared_winners)
+    def set_winner(self, adjudicator, winners):
+        self.scoresheets[adjudicator].set_declared_winners(winners)
```
czlee: So, this is a bit confusing, mostly because of the whole BP vs two-team format thing. There are multiple ways to do it, none of which are really satisfying. But it'd be helpful to read if the singular were used whenever it should be a single thing, and the plural were used whenever it should be a list (even if the list contains one item).

On a broader note, because most use-cases have just one winner (with BP non-final elimination rounds being the only exception), I think it's acceptable to break ordinary DRY practice / not always be strict about abstracting "number of winners", in order to be more explicit in our code when exactly one winner is expected. One of the reasons the previous code used `advancing` rather than `winners` was to provide an explicit reminder about when code expected to find two teams. So in this particular case, since this class presupposes two-team ballots, maybe it's acceptable to abstract the "number of winner" mechanics away from `DebateResultByAdjudicator`, and have this class "convert" it to a list of one item before passing it on to the scoresheet?

tabbycat/results/scoresheet.py
```diff
@@ -39,13 +31,34 @@ def identical(self, other):
...
+
+    def set_declared_winners(self, winners):
+        pass
+
+    def add_declared_winner(self, winner):
+        pass
+
+    def get_declared_winners(self):
+        pass
+
```
czlee: I actually think I'd prefer that these classes explicitly fail when a caller tries to set or get a declared winner when it's not meant to, rather than fail silently 🙂

This will mean that at the form level, forms will have to take care to distinguish between when a declared winner is and isn't required, and not attempt to declare one if it's not (i.e., if it's collecting it only for form validation). I think I'm more okay with this than I am with providing a method that depending on circumstance sometimes silently does nothing. But there might be something you've noticed that I haven't considered, so let me know if you've found otherwise 🙂

tabbycat/results/result.py
tienne-B: True.

On the broader note, I would still prefer to keep the format in the `ByAdjudicator` for consistency between the various places that use the `DebateResult`.

czlee (changes requested):
Hello, this happens when I set it up on a standard Australs setup:

```
  File "/home/czlee/tabbycat/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
[...]
  File "/home/czlee/tabbycat/venv/lib/python3.8/site-packages/django/forms/forms.py", line 402, in _clean_form
    cleaned_data = self.clean()
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 514, in clean
    self.clean_scoresheet(cleaned_data)
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 831, in clean_scoresheet
    high_point_declared = max(side_totals, key=lambda key: side_totals[key]) == cleaned_data[self._fieldname_declared_winner()]
TypeError: _fieldname_declared_winner() missing 1 required positional argument: 'adj'
```

czlee:
Adding to review above:

I get this on standard Australs settings _except_ with high-point win verification enabled, regardless of who I select as the winner:
```
  File "/home/czlee/tabbycat/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
[...]
  File "/home/czlee/tabbycat/venv/lib/python3.8/site-packages/django/views/generic/edit.py", line 142, in post
    return self.form_valid(form)
  File "/home/czlee/tabbycat/tabbycat/results/views.py", line 309, in form_valid
    self.ballotsub = form.save()
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 186, in save
    self.save_ballot()
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 434, in save_ballot
    self.save_participant_fields(result)
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 599, in save_participant_fields
    self.populate_result_with_scores(result)
  File "/home/czlee/tabbycat/tabbycat/results/forms.py", line 861, in populate_result_with_scores
    result.set_winners(adj, set([declared_winner]))
  File "/home/czlee/tabbycat/tabbycat/results/result.py", line 456, in set_winners
    self.scoresheets[adjudicator].set_declared_winners(winners)
AttributeError: 'HighPointWinsRequiredScoresheet' object has no attribute 'set_declared_winners'
```

I get this template error on standard Australs settings _except_ with tied-point win verification enabled, and selecting the lower-scoring team to win (i.e. an invalid ballot):
```
KeyError at /australs24team/admin/results/debate/old/89/new/
'adj'

In template /home/czlee/tabbycat/tabbycat/templates/components/form-errors.html, error at line 7
  {% include "components/alert.html" with type="danger" extra=errors\|cut:'__all__' %}
```

[Not a showstopper:] I think we probably want to include "affirmative" and "negative" here, a lot of people find it easier to think of teams in terms of positions than in terms of team names. Either instead of, or as well as—I'm not sure which is better, but I think position names should be in there somewhere.
image

Also I would like to apologize again for taking so long to spin this up for a test run on my local system. I'm hoping that with the branch checked out and a test database for this branch set up, I can now do a quick sanity drive quickly next time to pick up any obvious things, before taking longer to do the thorough check.

czlee:
Can I also pull @philipbelesky in to offer guidance on how this would fit in with the old/new results entry interfaces? This obviously requires changes to the result entry form; should we be (evenutally) modifying both forms or just one of them?

philipbelesky:
I think just the old results entry form. I've come to the conclusion that if/when the new entry interface is finished it needs to post to an API endpoint rather than its current (flawed) approach of trying to wrap Vue around Django's forms. That change is pretty much a rewrite so I don't think the current state of the 'new' form needs to be kept up to date - if anything it could be removed.

tienne-B:
Fixed

czlee:
Forms look better now, thanks! Will go over code/internals a bit later.

Don't put this in this PR (doesn't even need to be this release), but I wonder if we can optionally highlight low-point/tied-point wins in the results table.

tabbycat/results/scoresheet.py
```diff
@@ -72,93 +73,102 @@ def identical(self, other):
...
     """Draws are not permitted; winning teams must have a higher total.
     This is the standard type of scoresheet in Asia and Oceania."""
 
-    def _get_winner(self):
+    def _get_winners(self):
         aff_total = self.get_total('aff')
         neg_total = self.get_total('neg')
         if aff_total > neg_total:
-            return 'aff'
+            return set(['aff'])
```
czlee, line 133: Not a showstopper but if you like, the syntax `{'aff'}` does the same thing 🙂

(So long as there's no colon, it's a set not a dict.)

tabbycat/results/scoresheet.py
```diff
@@ -39,13 +31,22 @@ def identical(self, other):
         """Base implementation. Does nothing."""
         return True
 
+    def winners(self):
+        """Returns ['aff'] is the affirmative team won, and ['neg'] if the negative
```
czlee, line 35: Just pointing out I think this should be changed to `{'aff'}` (not `['aff']`)

tabbycat/results/result.py
```diff
@@ -519,57 +438,45 @@ def save(self):
...
-            logger.exception("Tried to set score by adjudicator %s, but this adjudicator "
-                "doesn't have a scoresheet.", adjudicator)
-            return
-        try:
-            scoresheet.set_score(side, position, score)
-        except AttributeError:
-            logger.exception("Tried to set score, but scoresheet doesn't do scores. "
-                    "self.takes_scores is %s", self.takes_scores)
+    def get_winner(self, adjudicator):
+        return next(iter(self.scoresheets[adjudicator].winners()), None)
```
czlee, line 450: Is there an assumption here that there is at most one winner whenever this function is called? If so, it's probably better to include an `assert` statement that verifies that `.winners()` has at most one item, and then return that item using `.pop()`, rather than using `next(iter(...))`. That way if the condition fails, it fails fast, and also it'll be more explicit to read.

tabbycat/results/result.py
```diff
@@ -693,202 +589,376 @@ def as_dicts(self):
...
     # --------------------------------------------------------------------------
 
-    def init_blank_buffer(self):
-        super().init_blank_buffer()
-        # To be complete, this list must have two sides in it, but we don't
-        # enforce anything before checking completeness.
-        self.advancing = []
+    def speakerscore_field_score(self, side, position):
+        return self.scoresheet.get_score(side, position)
+    get_score = speakerscore_field_score
```
czlee, line 881: Very minor but maybe put an empty line above this line? Without the blank link it looks like a typo lol
(If flake8 complains about it we should reconfigure flake8)
(I think I find this a little different to similar patterns in `admin.py` files because this line defines a new function, as opposed to modifies an existing one. But as it is is fine if you would find a blank line more confusing.)

tabbycat/tournaments/views.py
```diff
@@ -84,12 +85,11 @@ def get_context_data(self, **kwargs):
                     'content_object', 'user').order_by('-timestamp')[:updates]
         kwargs["initialActions"] = json.dumps([a.serialize for a in actions])
 
-        subs = BallotSubmission.objects.filter(
-            debate__round__tournament=t, confirmed=True).prefetch_related(
-            'teamscore_set__debate_team',
-            'teamscore_set__debate_team__team').select_related(
-            'debate__round__tournament').order_by('-timestamp')[:updates]
-        subs = [bs.serialize_like_actionlog for bs in subs]
+        subs = t.current_round.debate_set.filter(
```
czlee, line 88: I'd give this variable a different name (like `debates`) to be clearer about what it is
(then still call the new variable `subs` in line 92)

czlee (approved):
A few minor things I noticed but looks good, will merge and if you want to touch anything up feel free to do so in a new commit pushed straight to develop.

tabbycat/results/result.py
tienne-B: No, there may be either 0 or 1 winners, and if there are no winners, `None` should be returned. `.pop()` couldn't be used as it would mutate the set.

czlee: I said _at most_ one winner 😉 as in, there can't be more than one, though there might be less than one. But good point re `pop()`—kinda dissatisfying that `next(iter(...))` is the cleanest way otherwise, but it is what it is! I still think you want to `assert` assumptions to get it to fail fast though, or at least get the logger to throw an error at Sentry (if you think crashing due to the state is worse than ploughing ahead silently).

czlee:
Just noting a regression in #922 (see reopening comment there).

PR #1182, opened 2019-08-12
Simulaterounds: Prevent last substantive speaker from being reply

tienne-B:
This commit reduces the choices of the reply speaker when using the `simulaterounds` command. The bug was caused by a mixup of the 0-indexed array and 1-indexed speaker positions.

### DIFF
--- a/tabbycat/results/dbutils.py
+++ b/tabbycat/results/dbutils.py
@@ -94,7 +94,7 @@ def add_result(debate, submitter_type, user, discarded=False, confirmed=False,
                 result.set_ghost(side, i, False)
 
             if t.reply_position is not None:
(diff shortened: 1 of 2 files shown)


philipbelesky:
Hey thanks for adding this in, the `reply-random` flag works well. I think removing the min/max arguments in favour of using the set preference makes sense, but I'll defer to @czlee to approve that part as he wrote the utility.

tienne-B:
`reply-random` already existed and the min/max flags were already unused in favour of using preference values. Not adding anything new here!

philipbelesky:
Ah right, I didn't catch that they were redundant already; will merge it in.

PR #1183, opened 2019-08-12
Simulaterounds: Add motion vetos to debates

tienne-B:
This commit adds motion vetos to the simulaterounds command, changing a choice between motions into a sample. A check has also been added for the sample size if there aren't enough motions.


philipbelesky:
Hey, thanks for noticing this an adding it in. It works well!

PR #1189, opened 2019-08-18
Changelog: Add section for M-Release

tienne-B:
Also moved the pull-up metrics features to that section.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,12 +2,17 @@
 Change Log
 ==========
 
-2.3.0 (LaPerm)
---------------
+2.4.0 (M)
+---------
 *Release date: ?*
 
 - Pull-ups can now be restricted to teams with the lowest draw strength (by speaker or team points) of their bracket.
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -2,12 +2,17 @@
 Change Log
 ==========
 
-2.3.0 (LaPerm)
---------------
+2.4.0 (M)
+---------
 *Release date: ?*
 
 - Pull-ups can now be restricted to teams with the lowest draw strength (by speaker or team points) of their bracket.
```
philipbelesky: Shouldn't these be in the 2.3.0 section? Also I hadn't noticed your earlier commit with the CHANGELOG, so have just merged that now into the released branch. Maybe its best to wait until the release branch is merged back into develop before adding a 2.4.0 section.

tienne-B: No, the merge commit was after the branching of the release.

tienne-B: And yes, we could wait for after the release. Names in mind for "M"?

PR #1191, opened 2019-08-19
Un-mark email templates for translation

tienne-B:
Emails that are sent through the person-selector view do not show information (neither the help_text nor verbose name). Therefore, such text would not need translation.

The text in English has been kept however for quick reference in the file.

These strings have been hidden in Crowdin.

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -1379,16 +1379,16 @@ class BallotEmailMessageBody(LongStringPreference):
 
 @tournament_preferences_registry.register
 class PointsEmailSubjectLine(StringPreference):
-    help_text = _("The subject line for emails sent to speakers with their team points.")
-    verbose_name = _("Team points subject line")
+    help_text = "The subject line for emails sent to speakers with their team points."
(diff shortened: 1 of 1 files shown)


czlee:
This makes the code confusing without any comments—if you were looking at it, you'd wonder why there's an inconsistency, and my first instinct would be to "correct" it again. If those fields aren't used anywhere, I believe they're not mandatory, so we could remove them altogether. If we want to retain them for completion or something, I'd rather that the code say why they're not marked as translatable. Even if a string is marked translatable but hidden on Crowdin, that fact should ideally be marked in source code.

The way I think about things like this is: one day, in the future, for reasons unknown to our present selves, either one of our future selves or a future developer might for whatever reason place the options somewhere that shows the help text and verbose name again. They shouldn't then be left to be scratching their heads about why it's not translating properly.

That said, do these strings show up translated in the Edit Database Area?

tienne-B:
I thought I had added a note to the comment about having no section.

I also forgot about the Edit Database views; they would appear translated.

Issue #1069, opened 2019-04-02
Check for cases where bulk_update are applicable 


philipbelesky:
I'll update the requirements to Django to 2.2 soonish, although feel free to do so as part of any commits attached to this issue. When that happens there is a new `bulk_update` [method available](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#bulk-update).

I feel like there definitely are a few cases where we've wanted this in the past; some of which had a notable performance impact. I'll try and look into those, but wanted to throw this up both as an FYI/reminder and incase anyone was interested to help.

czlee:
Just a note, I once had a go at this with `bulk_create()` (which has existed much longer) and the fact that it didn't call the `save()` function or trigger `pre_save` or `post_save` signals put me off, because I felt it was too risky (unless the existing performance inefficiency was prohibitive, which it wasn't).

I think `bulk_update()` is likely to be safer in most cases, but if the field in question is or might one day be touched by the `save()` method, or `pre_save` or `post_save` signals, we should probably err on the side of leaving it with slower but safer update methods. Or at least we should consider how to make sure the irregularity won't cause something to break in the future when some unsuspecting future developer updates the relevant code for something else.

PR #1192, opened 2019-08-24
Use bulk_update with private url key generation

tienne-B:
This commit replaces the exception-catching with one-at-a-time updating of People objects when adding URL keys with the new bulk_update method of Django 2.2.

There is a huge performance increase with this!

Applies to #1069

### DIFF
--- a/tabbycat/privateurls/utils.py
+++ b/tabbycat/privateurls/utils.py
@@ -2,7 +2,7 @@ import logging
 import string
 from secrets import SystemRandom
 
-from django.db import IntegrityError
+from participants.models import Person
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tabbycat/privateurls/utils.py
```diff
@@ -13,20 +15,21 @@ def generate_url_key(length=8):
     return ''.join(SystemRandom().choice(chars) for _ in range(length))
 
 
-def populate_url_keys(queryset, length=8, num_attempts=10):
+def populate_url_keys(people, length=8, num_attempts=10):
     """Populates the URL key field for every instance in the given QuerySet."""
-    for instance in queryset:
+    existing_keys = list(Person.objects.exclude(url_key__isnull=True).values_list('url_key', flat=True))
```
tienne-B: Not using a filter to exclude people of other tournaments, but it should still be fine.

czlee:
This is cool! Since we're already on this, do you want to try seeing if we can make this even easier to read, removing the nested loop? Basically if we generate all keys in one go (while checking for duplicates), _then_ assign them to `Person` instances, does that make the code easier to follow?

Double-check this, but I think `generate_url_key()` is only used in one place so you could even refactor that if you think it would help.

(Not obligatory, but the perfectionist in me wonders 😊)

czlee:
Hmm, it is possible, but I just tried it and it's actually a little fiddlier than what you have.

With the existing method I actually liked the separate `generate_url_key()` a little better, so I'll merge just that commit 🙂

PR #1193, opened 2019-08-24
Update pip dependencies

tienne-B:
This commit updates the dependency lists to reflect the current need.

These actions have been taken:
- Moved sentry-sdk to Heroku
  Sentry does not seem to be used locally, so is Heroku-specific.
- Removed specific commit requirement from django-dynamic-preferences
  As the commit in question is included in the version I used: 1.7.1
- Removed dj-database-url from core requirements
  As duplicated in Heroku requirements
- Removed sqlparse
  Unused as Django's queries are mostly used instead
- Change branch for django-debug-toolbar-request-history
  For development, as the master branch is incompatible with the debug-toolbar v2.
- (Removed transifex-client)

### DIFF
--- a/config/requirements_core.txt
+++ b/config/requirements_core.txt
@@ -3,7 +3,7 @@
 # Django
 Django==2.2.4
 django-appconf==1.0.3                   # Helper for handling app configs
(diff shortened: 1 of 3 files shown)


philipbelesky:
Great, thanks for cleaning these up.

PR #1194, opened 2019-08-24
Note having no institution on reg cards

tienne-B:
This change adds "Unaffiliated" to the institution item on participants registration cards if the participant (adjudicator or team/speaker) has none set, instead of having no value shown.


tienne-B:
I'd just go with "Unaffiliated." Short but would be clearer than "None"

PR #1195, opened 2019-08-24
Update PY compatibility note to >=3.6

tienne-B:
In deploy_heroku.py, it is said to be compatible with Python 3.4 and up, but was not since 223c98327cf7134c604393b600ac7f7f98211e5f (and the comment was not updated)

### DIFF
--- a/deploy_heroku.py
+++ b/deploy_heroku.py
@@ -11,7 +11,7 @@ import sys
 try:
     from django.core.management.utils import get_random_secret_key
 except ImportError:
-    from secrets import SystemRandom
+    from random import SystemRandom
 
     def get_random_secret_key():
         chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'


czlee:
Oh, I see. Actually, deploy_heroku.py isn't meant to be governed by the Python version we normally support (as you can see by the fact that it's meant to work on Python 2.7). So if it's only the `secrets` module that breaks Python 3.4 compatibility, I'd actually prefer to restore it. In principle, deploy_heroku.py is meant to be able to be run by anyone with a (reasonable) version of Python, without having to update their entire machine to be Tabbycat-compatible.

czlee:
In fact... on further inspection, `secrets.SystemRandom` is just a direct import from `random.SystemRandom`. So the simpler solution would be to drop the `secrets` import and import from `random` instead.
https://docs.python.org/3/library/secrets.html#secrets.SystemRandom
https://github.com/python/cpython/blob/3.7/Lib/secrets.py#L20

tienne-B:
Done.

PR #1214, opened 2019-09-06
i18n: Use humanize for cardinals

tienne-B:
This commit uses the django.contribs humanize module to get the ordinal form of numbers when used in apps.

Also, numeral forms of numbers have been removed from localization as number translation does not seem prevalent and is not supported by the rest of Django.


philipbelesky:
Great, this cuts down quite a bit of template code. Are those ordinals also translated nicely by humanise?

tienne-B:
Have a quick patch for French that I couldn't upload till now; internet problems.

PR #1215, opened 2019-09-06
jsi18n: Delete duplicate messages

tienne-B:
Causes failures in message compilation.

Quick warning to be careful not to include duplicate messages when synchronizing with Crowdin.


philipbelesky:
Noted, thanks for the fix!

tienne-B:
Euh, I modified the French file and not the English one in this patch. Will create another for the English source. (I test translations in French)

PR #1216, opened 2019-09-06
Options: Update dynamic preferences migration

tienne-B:
There is a new one.


philipbelesky:
Sorry, what was the context here? Is this due to the bump in the dynamic preferences library requiring a new migration?

tienne-B:
I think so

PR #1220, opened 2019-09-07
Raise 404 in public scoresheets when debate nonexistent

tienne-B:
Fixes BACKEND-29N

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.db import ProgrammingError
 from django.db.models import Count, Q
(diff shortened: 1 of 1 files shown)


czlee:
Hmm, actually, overriding `get_object()` in this class (as was done in 344544d79ef68fe82423f9420d60a424daa6566b, see #1006) overrides and repeats (in the DRY sense) a lot of the intelligence in `SingleObjectMixin`, see [Django's source code](https://github.com/django/django/blob/master/django/views/generic/detail.py#L20-L56) to get an idea. The better thing to do is to override the `get_queryset()` to add the `.select_related()` and `.prefetch_related()` methods, and to let `SingleObjectMixin` take care of the `.get()` part. Can we please modify accordingly?

PR #1222, opened 2019-09-07
i18n: Remove numbers from i18njs source

tienne-B:


tienne-B:
I'll see about doing this through Crowdin.

PR #1224, opened 2019-09-07
Public scoresheets: Change get_object for get_queryset

tienne-B:
Using the get_queryset() method is more appropriate and avoids some error checking. For that, the model of the private URL scoresheet view has been changed to Debate and the slug retrieval changed.

Touch-up to 344544d and 50a008468.

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.db import ProgrammingError
 from django.db.models import Count, Q
-from django.http import Http404, HttpResponseRedirect
+from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from django.utils import timezone
 from django.utils.translation import gettext as _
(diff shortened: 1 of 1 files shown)


czlee:
Just noting link to #1220

tabbycat/results/views.py
```diff
@@ -620,6 +614,11 @@ def get_context_data(self, **kwargs):
 
 class PrivateUrlBallotScoresheetView(RoundMixin, SingleObjectByRandomisedUrlMixin, BasePublicBallotScoresheetsView):
 
+    slug_url_kwarg = 'url_key'
+
+    def get_slug_field(self):
+        return 'debateadjudicator__adjudicator__url_key'
```
czlee: I think it works if you just set the `slug_field` property of the class, right? Since this doesn't need to be generated dynamically?

tienne-B: You're right

czlee:
Looks good! There are a couple of things I noticed that date back to #957, but I'll raise them in a new issue.

PR #1225, opened 2019-09-07
Reconcile emojis with spreadsheet

tienne-B:


philipbelesky:
Ah thanks, I was going to reconcile these at a later date but may as well do so now.

PR #1226, opened 2019-09-08
Fixup manageidentifiers command

tienne-B:
- The 'length' argument is obsolete and removed.
- The options are no longer made into an attribute
- create_identifiers() gets the model to create from.

### DIFF
--- a/tabbycat/checkins/management/commands/manageidentifiers.py
+++ b/tabbycat/checkins/management/commands/manageidentifiers.py
@@ -2,9 +2,9 @@ from argparse import ArgumentParser
 
 from django.db.models import Q
 
+from checkins.models import PersonIdentifier, VenueIdentifier
 from participants.models import Person
 from utils.management.base import TournamentCommand
-from venues.models import Venue
 
 from ...utils import create_identifiers, delete_identifiers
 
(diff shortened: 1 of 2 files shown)


tienne-B:
Fixed.

czlee:
Question, can I ask why you removed the "Skipping N existing identifiers" message?

tabbycat/checkins/management/commands/manageidentifiers.py
```diff
@@ -14,42 +14,41 @@ class Command(TournamentCommand):
...
-            model_query = model_query.filter(Q(adjudicator__tournament=tournament) | Q(speaker__team__tournament=tournament))
-
-        if options['subcommand'] == "generate" and options['overwrite'] or options['subcommand'] == "delete":
-            self.stdout.write("Deleting all check-in identifiers for {0:s}...".format(options['model']))
-            delete_identifiers(model_query)
+        if options['model'] == 'person':
+            queryset = Person.objects.filter(
+                Q(adjudicator__tournament=tournament) | Q(speaker__team__tournament=tournament)
+            )
+            identifier_model = PersonIdentifier
```
tienne-B: What's wrong with finding the related model through attributes?

czlee: Oh, it just seems unnecessary given that it's known and we only deal with two models, and it makes the code more opaque. Especially in management commands, I'd like the code to be straightforward.

I may not always have lived up to this standard in other code, for many reasons (e.g. maybe something else seemed like a good idea at the time but doesn't now, or maybe it _was_ a good idea at the time but other things changed, or maybe it's just a borderline case).

tienne-B:
Re-added; I removed it to avoid another `if`.

czlee:
Travis failure is the same as in 97b31a5a118b9ec84b755c425551530e9d83755b, so not due to this branch. Merging.

PR #1232, opened 2019-09-11
Prefer django.test over unittest

tienne-B:
This commit replaces the uses of `unittest` with `django.test` so that they run with the other tests.

### DIFF
--- a/tabbycat/draw/tests/test_bphungarian_parts.py
+++ b/tabbycat/draw/tests/test_bphungarian_parts.py
@@ -1,4 +1,4 @@
-import unittest
+from django.test import TestCase
 
 from ..generator.bphungarian import BPHungarianDrawGenerator
 from .utils import TestTeam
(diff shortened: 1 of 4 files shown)


czlee:
Are the tests not already being run? They seem to be included in the test suite in Travis:
https://travis-ci.org/TabbycatDebate/tabbycat/jobs/583395139

Here's Travis on the PR, it runs 269 tests, just like the develop branch:
https://travis-ci.org/TabbycatDebate/tabbycat/jobs/583717159

tienne-B:
I'd think that even though the tests are run in either case that we should prefer using the Django-specific testing capabilities as being better adapted to Django and for consistency.

czlee:
Hmm, I don't think I agree. If tests don't require the Django framework, I would rather be explicit about that fact, so that it's clearer what the tests do. These particular tests are pure logic tests, and have nothing to do with the database or anything Django.

tienne-B:
Makes sense.

PR #1233, opened 2019-09-11
Check for duplicate allocations in scoresheet view

### LINKED ISSUE #1229: Private URL view of ballot crashes when adjudicator assigned to multiple debates
If you assign an adjudicator to several debates in the same round, then try to load the `PrivateUrlBallotScoresheetView` for them, it throws `MultipleObjectsReturned` and shows a 500 page.

To reproduce: In the Edit Database area for a tournament with at least one round drawn, create a new debate adjudicator for an already-assigned adjudicator, assigning them (also) to another debate. Then load their private URL page, and go to their ballot for that round.

-------------

Note: I think you can't easily assign adjudicators to multiple debates in the allocation UI, but it's a valid database state and we account for it in other pages. For example: participant landing pages list all assigned debates:

https://github.com/TabbycatDebate/tabbycat/blob/9851b4e4842371a528c1a2c107d4a45ea2923a04/tabbycat/participants/templates/in_this_round.html#L22-L26

and the public new ballot set view shows an error page if more than one debate is assigned:

https://github.com/TabbycatDebate/tabbycat/blob/9851b4e4842371a528c1a2c107d4a45ea2923a04/tabbycat/results/views.py#L487-L493

(I probably wouldn't have been looking out for this edge case if I hadn't already had to deal with the two cases above from previous bug reports!)

Relates to #957 and #1224.

tienne-B:
This commit makes public scoresheet views check whether there are multiple adjudicator allocations in a round. This can happen in private URL views as it filters per adjudicator and round.

Fixes #1229.

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -556,6 +556,7 @@ class BasePublicBallotScoresheetsView(PublicTournamentPageMixin, SingleObjectFro
     public_page_preference = 'ballots_released'
     tournament_field_name = 'round__tournament'
     template_name = 'public_ballot_set.html'
+    error_template_name = 'public_ballot_set_error.html'
 
     def matchup_description(self):
         if use_team_code_names(self.tournament, False):
@@ -568,6 +569,16 @@ class BasePublicBallotScoresheetsView(PublicTournamentPageMixin, SingleObjectFro
             'round'
         ).prefetch_related('debateteam_set__team')
 
+    def response_error(self, error):
+        status, message = error
+        return self.response_class(
+            request=self.request,
+            template=[self.error_template_name],
+            context={'message': message},
+            using=self.template_engine,
+            status=status,
+        )
+
     def get(self, request, *args, **kwargs):
         try:
             self.object = self.get_object()
@@ -577,14 +588,7 @@ class BasePublicBallotScoresheetsView(PublicTournamentPageMixin, SingleObjectFro
             error = self.check_permissions()
 
         if error:
-            status, message = error
-            return self.response_class(
-                request=self.request,
-                template=['public_ballot_set_error.html'],
-                context={'message': message},
-                using=self.template_engine,
-                status=status,
-            )
+            return self.response_error(error)
 
         return super().get(self, request, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


tabbycat/results/views.py
```diff
@@ -569,9 +569,13 @@ def get_queryset(self):
         ).prefetch_related('debateteam_set__team')
 
     def get(self, request, *args, **kwargs):
-        self.object = self.get_object()
+        try:
+            self.object = self.get_object()
+        except self.model.MultipleObjectsReturned:
+            error = (500, _("It looks like you were assigned to two or more debates. Please contact a tab room official."))
```
czlee: Hmm, I'm not actually sure what the error code should be, but I feel like it should be some 4xx. Probably 404? It's not necessarily an error that the judge was assigned two or more debates (it might've been intentional), and even if it is, it's not a server error (the tab director made it like that). I guess 404 isn't necessarily semantically correct, but it's possibly the closest?

Open to discussion on this.

tienne-B: A "Not Found" error is not really appropriate as the object is found, just incongruous. The 400 series is for client error, which an unsupported adj state is not. Thus it would be in the 500 series, for which 500 is the general code (and there are no more specific codes for this situation)

tienne-B:
To avoid needing to raise an error, the URL could include the debate's ID, if that seems better...

czlee:
> To avoid needing to raise an error, the URL could include the debate's ID, if that seems better...

My original thinking was that if URLs are by adj, then it's just a single data entry URL all tournament, and all they have to do is refresh the page to see their new ballot next round. At least, that was the idea. Now that we have landing pages, this might be less critical. Also, it relies on there being a clean way back to the results submission page from the confirmation page. (It actually used to redirect to itself, but we changed that in https://github.com/TabbycatDebate/tabbycat/commit/04341f0e1a4881fcff2bbdd2c938584ee8e3a5e0#diff-b86fca23dd96338a362014a953a86c47R402-R408 because it was confusing.)

Because the main use case is still one debate per adjudicator per round, I'd still rather design for this case in principle, and treat multiple debates per adjudicator as an exception, if it would affect the user workflow.

My computer's dead and about to be sent in for repairs, so I won't be able to review PRs for a couple of weeks, sorry :sob:

tienne-B:
I agree; and so I hadn't changed this patch :) I am still for using the `500` code though.

tabbycat/results/views.py
czlee: Hmm, yeah, I see where you're coming from. I think I'm a little hesitant because normally I think of "500" as "omg wtf just happened", as opposed to a predictable error condition. But if you feel strongly about it I don't mind running with 500 for now; if it proves troublesome we can change it later.

czlee:
I realized when reviewing this that the error pages for private URLs should also be treated as private URL pages, and therefore need their own template. Since it wasn't part of the original bug, but just a consequence of fixing the bug that you notice this is necessary, I just went ahead and added it to the PR.

PR #1240, opened 2019-09-14
Fix logic inversion for draw display links

### LINKED ISSUE #1239: Cannot view draw links for draws other than current
To reproduce:

- Go to the admin draw display area for a completed round
- Click any of the display links
- Links go to current round rather than the specific round

I assume this was a result of moving to the completed round system?

tienne-B:
If the draw was not current (i.e. completed), it would only show "Display" links for current rounds.

Also removed the or statement for if there is only one active round, as that active round might not be the current one.

Fixes #1239.

### DIFF
--- a/tabbycat/draw/templates/draw_display_admin.html
+++ b/tabbycat/draw/templates/draw_display_admin.html
@@ -271,7 +271,7 @@
 
       {% trans "(for the briefing room)" as briefing_room_subtext %}
 
-      {% if not round.is_current or tournament.current_rounds|length == 1 %}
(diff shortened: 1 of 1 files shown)


philipbelesky:
Thanks for adding this in. I'm not sure the "All Current Rounds" links need to be shown outside of out-rounds but this addresses the issue at hand.

tienne-B:
Guess I should have based the PR on the release branch...

philipbelesky:
No worries, I've just cherry picked it across.

PR #1241, opened 2019-09-15
Show missed speeches message if limit is 0

tienne-B:
The filter for the number of speeches a speaker can miss and still be in the tab has a default of -1 to include all speakers. The choice of 0 (can't miss any debate) is a valid option, but was unaccounted for in the messages about the option.

This changes the comparisons to greater than *or equal to* 0.

### DIFF
--- a/tabbycat/standings/templates/speaker_standings.html
+++ b/tabbycat/standings/templates/speaker_standings.html
@@ -28,7 +28,7 @@
     {% endblocktrans %}
   {% endif %}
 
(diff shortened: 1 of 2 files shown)


tienne-B:
The other commit will fix the issue described on [Facebook](https://www.facebook.com/groups/tabbycat.debate/permalink/1316340805200540/). Both could be cherry-picked for the release.

philipbelesky:
Great, thanks for the fix!

PR #1246, opened 2019-09-29
Codeclimate: Remove non-existing file from config

tienne-B:
Why is Ruby and PHP types listed; they are not used here.
Also, what is `.coveragerc` for?


philipbelesky:
I suspect we may have just copied a default config over. 

I think `.coveragerc` is there as configuration for [this library](https://coverage.readthedocs.io/en/v4.5.x/index.html) which is then used by code climate to estimate test coverage.

Issue #988, opened 2019-01-12
Rethink 'Setup' menu configuration


philipbelesky:
The full hierarchy of the 'setup' menu is large and unsemantic:

- Configuration
    - .... all the categories
    - .... set current round
    - .... all the presets
- Import Data
    - .... all the options
- Participants
    - Participants List
    - Institutions List
    - Email Registration
    - Code Names
    - Speaker Categories
    - Speaker Eligibility
- Private URLs
    - Print
    - Email
    - Generate
- Emails
    - Status
    - Send Test
    - Send Custom
- Preformed Panels
    - Per-round panels
- Edit Database

Problems:

- 'Setup' is presumably something you do at the start of the tournament; that doesn't necessarily encompass preformed panels or emails
- There are many second layers within each sub-section; ideally these would be menu items on their own
- General messyness

Thoughts:

- It could be possible to re-organise the menu around the *type* of data with functionality grouped by type on sub page. I.E. a "Speakers" page that then lists private URLs, categories, import options, etc. 
- Edit Database could go into the very top-most item on the menu given its not tournament-specific
- There's a case for making preformed panels its own top level menu item.

This problem, at this stage, should probably be slated for the release after next. It could be done in conjunction with the configuration re-organisation and other new menu items such as adjudicator categories and a front-end for adding users.

PR #1247, opened 2019-09-30
Move Edit Database link to admin title dropdown

tienne-B:
The edit database interface is not tournament-specific and so it should not be under the 'Setup' menu. The title dropdown (with the list of tournaments) is best.

Also marked some strings in the file for translation and changed some blocktrans to trans if the block is un-needed.

How to call the "title dropdown" in the changelog?

Works on #988

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Change Log
 ---------
 *Release date: ?*
 
+- Moved the 'Edit Database' menu item to the dropdown with the logo in the admin navigation.
 - Pull-ups can now be restricted to teams with the lowest draw strength (by speaker or team points) of their bracket.
 - The number of pull-ups and draw strength by speaker score are now available as team standing metrics.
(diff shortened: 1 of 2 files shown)


czlee:
Putting the Edit Database link in the main dropdown works fine for me (if it's fine for @philipbelesky).

philipbelesky:
Happy with the shift, although it may cause a little confusion with users. Guess we can just note that during release. I can't do so right now, but it might be worth checking if we refer to the old location in the docs anywhere.

czlee:
We could have a transitional "where did this link go?" where the old item was during Maine Coon, then drop the annoyance after?

tienne-B:
@czlee: Unsure (for now) that there should be a "transition" for if more work is done on #988, there might not be as much sense.

czlee:
Also true, take my comment as assuming there is only one change, can rethink as the entire rethink is done :slightly_smiling_face:

PR #1258, opened 2019-10-02
Fix translation of warning message

### LINKED ISSUE #1257: Error message for BP voting ballots crashes with 500
Sentry Issue: [BACKEND-2BV](https://sentry.io/organizations/tabbycat/issues/1252961179/?referrer=github_integration)

```
AttributeError: 'tuple' object has no attribute 'replace'
(6 additional frame(s) were not displayed)
...
  File "django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
  File "django/views/generic/base.py", line 158, in get
    context = self.get_context_data(**kwargs)
  File "options/views.py", line 54, in get_context_data
    "configuration to use consensus ballots."))
  File "django/utils/translation/__init__.py", line 79, in gettext
    return _trans.gettext(message)
  File "django/utils/translation/trans_real.py", line 282, in gettext
    eol_message = message.replace('\r\n', '\n').replace('\r', '\n')
```

tienne-B:
It was mistakenly placed in a tuple.

Closes BACKEND-2BV
Closes #1257.

### DIFF
--- a/tabbycat/options/views.py
+++ b/tabbycat/options/views.py
@@ -40,13 +40,13 @@ class TournamentConfigIndexView(AdministratorMixin, TournamentMixin, TemplateVie
         if t.pref('teams_in_debate') == 'bp':
             if t.pref('ballots_per_debate_prelim') == 'per-adj' or \
                t.pref('ballots_per_debate_elim') == 'per-adj':
-                error = _(("Your draw rules specify four teams per-debate but ",
-                           "your ballot setting specifies that adjudicators ",
-                           "submit independent ballots. These settings ",
-                           "<strong>are not compatible and will cause results ",
-                           "entry to crash</strong>. You need to go back to ",
-                           "the Debate Rules settings and change your ",
-                           "configuration to use consensus ballots."))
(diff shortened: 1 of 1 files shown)


tabbycat/options/views.py
```diff
@@ -40,13 +40,13 @@ def get_context_data(self, **kwargs):
...
-                           "entry to crash</strong>. You need to go back to ",
-                           "the Debate Rules settings and change your ",
-                           "configuration to use consensus ballots."))
+                error = _("Your draw rules specify four teams per-debate but "
+                           "your ballot setting specifies that adjudicators "
+                           "submit independent ballots. These settings "
+                           "<strong>are not compatible and will cause results "
+                           "entry to crash</strong>. You need to go back to "
+                           "the Debate Rules settings and change your "
+                           "configuration to use consensus ballots.")
```
czlee: Can we either align this so that the quotation marks align, or so that the hanging lines are a multiple of four spaces (in this case I'd prefer the former but the latter is okay)?

philipbelesky:
Thanks, I've tested and merged into hotfix

PR #1265, opened 2019-10-05
Double-filter for tournament presence; closes #1263

### LINKED ISSUE #1263: Institutions list includes teams/institutions from other tournaments
Apparently, if you open the institutions list on a site with multiple tournaments, it filters the institutions by tournament, but still includes teams and adjudicators from all tournaments for that institution. For example, if University A has 3 teams in Tournament Alpha, and 4 teams in Tournament Beta; and University B has 5 teams in Tournament Alpha, but is not participating in Tournament Beta, then Tournament Alpha will show "University A: 7, University B: 5", and Tournament Beta will show "University A: 7".

I suspect it's to do with this filter:
https://github.com/TabbycatDebate/tabbycat/blob/1b1f11f05254b8f5aab62c9d60eb46b4eb13fd53/tabbycat/participants/views.py#L90-L96

Normally putting the `filter()` before the `annotate()` should do the trick:
https://docs.djangoproject.com/en/2.2/topics/db/aggregation/#order-of-annotate-and-filter-clauses

But I'm guessing in this case the `|` operator in the `filter()` clause messes this idea up?

I'll do this once I get a personal computer again, but if someone else wants to examine it in the meantime that's cool too. We should add a (very simple) unit test for this.

philipbelesky:
This seems to produce the correct results, but is a little ungainly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "scripts": {
     "build": "NODE_ENV='production' npm-run-all -p build-* cp-*",
-    "windows-build": "SET NODE_ENV='production' & npm-run-all -p build-* cp-*",
+    "windows-build": "SET NODE_ENV='production' & npm-run-all -p build-* cp-* && cpx \"tabbycat/static/vue/app.js\" \"tabbycat/static/vue/js/\"",
     "lint": "npm-run-all -p lint-*",
     "serve": "npm-run-all -p serve-*",
     "docs": "sphinx-autobuild docs docs/_build/html --port 7999",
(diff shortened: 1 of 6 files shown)


tienne-B:
Yeah, this was also the solution I was thinking of. I think it is the best solution.

czlee:
I'll look at this once I have a new computer, but it was the first thing that came to mind for me too. I agree that it's a bit clunky though, so will also try poking around to see if I can find something better :slightly_smiling_face:

philipbelesky:
Ok, I'll cherry pick this onto hotfix for now but leave the PR open (for merging into develop)

tienne-B:
Moot PR as the hot-fix branch has been merged onto `develop`.

czlee:
Oh, I would've preferred that this be left open (as @philipbelesky said in #1263) so that I remember to do it. But I'll go back to reopen #1263 instead.

PR #1266, opened 2019-10-06
Check for motions being released in adj debate table

tienne-B:
As reported during NAUDC 2019. The motions were visible in adjudicator record debate tables even if not officially released. A check has been added to whether the motion was released or all of them are.

Also marked the dash for translation.


philipbelesky:
Great, thanks for the fix!

PR #1268, opened 2019-10-09
Mark skipped strings as translatable

tienne-B:
In the import of preset tournaments and in the names/descriptions of option presets.

### DIFF
--- a/tabbycat/importer/views.py
+++ b/tabbycat/importer/views.py
@@ -225,9 +225,8 @@ class LoadDemoView(AdministratorMixin, PostOnlyRedirectView):
                 "<p>There were one or more errors creating the demo tournament. "
                 "Before retrying, please delete the existing demo tournament "
                 "<strong>and</strong> the institutions in the Edit Database Area.</p>"
-                "<p><i>Technical information: The errors are as follows:"
-                "<ul>") + "".join("<li>{}</li>".format(message) for message in e.itermessages()) + "</ul></i></p>"
-            ))
+                "<p><i>Technical information: The errors are as follows:</i></p>"
+            ) + "<ul><li><i>" + "</i></li><li><i>".join(e.itermessages()) + "</i></li></ul>"))
             logger.error("Error importing demo tournament: " + str(e))
             return redirect('tabbycat-index')
         else:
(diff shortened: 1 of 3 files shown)


tabbycat/importer/views.py
```diff
@@ -221,18 +221,18 @@ def post(self, request, *args, **kwargs):
...
                                     force=True, strict=False, encoding='utf-8')
         except TournamentDataImporterError as e:
-            messages.error(self.request, mark_safe(
+            messages.error(self.request, mark_safe(_(
                 "<p>There were one or more errors creating the demo tournament. "
                 "Before retrying, please delete the existing demo tournament "
                 "<strong>and</strong> the institutions in the Edit Database Area.</p>"
                 "<p><i>Technical information: The errors are as follows:"
-                "<ul>" + "".join("<li>{}</li>".format(message) for message in e.itermessages()) + "</ul></i></p>"
+                "<ul>") + "".join("<li>{}</li>".format(message) for message in e.itermessages()) + "</ul></i></p>"
```
czlee: I think it might be better for the end of the translatable string to go before the `<ul>` tag? I'm not sure what Crowdin does but some translation platforms try to be intelligent about matching HTML tags, and having just a single one could confuse both machines and humans. (Thoughts?)

tienne-B: I prefer to match. Crowdin might complain for unmatched tags but doesn't try to be smart about it, so it is less of a concern.

tabbycat/options/presets.py
```diff
@@ -1,5 +1,7 @@
 import logging
 
+from django.utils.translation import gettext as _
```
czlee: This should be `gettext_lazy`, otherwise all of the class attributes won't translate, because the functions for class attributes are evaluated when the class is loaded, not at runtime. (Let me know if you have questions about this.)

tabbycat/importer/views.py
czlee: It doesn't really, I think it just makes it look like there was a mistake. If it's important to indicate to translators, it should be in a context note. (But I think in this case it's clear from the English, so no need.)

PR #1270, opened 2019-10-09
Shorten language selector label

tienne-B:
The link to the language selector was long, and would get longer. This change sets the label to be "Languages" in English then in the interface language, instead of all options.

The existing translations could easily be placed in Crowdin.


czlee:
Looks great, thanks!

PR #1274, opened 2019-10-13
Mark checkboxes as not required

tienne-B:
Checkboxes are set as being required in Django, which means that a False value cannot be inputted. That is unintuitive and can lead to wrong answers as required to submit the feedback.

This change makes the checkbox unrequired, letting it be false if unchecked.

### DIFF
--- a/tabbycat/adjfeedback/forms.py
+++ b/tabbycat/adjfeedback/forms.py
@@ -132,7 +132,7 @@ class BaseFeedbackForm(forms.Form):
         if question.answer_type == question.ANSWER_TYPE_BOOLEAN_SELECT:
             field = BooleanSelectField()
         elif question.answer_type == question.ANSWER_TYPE_BOOLEAN_CHECKBOX:
-            field = forms.BooleanField()
+            field = forms.BooleanField(required=False)
         elif question.answer_type == question.ANSWER_TYPE_INTEGER_TEXTBOX:
             min_value = int(question.min_value) if question.min_value else None
             max_value = int(question.max_value) if question.max_value else None


philipbelesky:
Sorry, could you clarify under what circumstances the fields are being marked as required? I just modified one of the default checkbox questions to be required and the form field as-rendered doesn't have a required attribute, or enforce that I tick it before submitting the form. Is this a browser issue?

tienne-B:
For me, not ticking a checkbox can be seen as a response to the question, and so this passive answer is still an answer. As such, marking such a question as mandatory doesn't make sense for checkboxes.

philipbelesky:
Yes, I agree that the questions should not force an interaction. I'm just wondering what this code actually changes? From what I can tell django doesn't mark those fields as required (on the front end) in any case?

tienne-B:
For Django, requiring an answer to a checkbox requires it to be ticked (like for EULAs).

philipbelesky:
That doesn’t seem to be the case though, and is what I’m trying to understand. When I’ve set a checkbox feedback question to be required it doesn’t render with the required attribute and the form doesn’t enforce it to be ticked. 

Screen Shot 2019-10-19 at 5 17 58 pm

philipbelesky:
Oh, actually, it turns out that the admin version of the feedback form just never renders or enforces any of the required attributes. I'll merge this and make a new issue for that.

czlee:
The issue this is supposed to fix still exists (after da6ae86818ab8542346ca22fe5f9c9c5fc85a64b), because line 161 still sets the `required` attribute according to what is specified by the user, overriding whatever was previously set by the constructor. So correct me if I'm wrong, but as far as I can tell this PR didn't really do anything?

https://github.com/TabbycatDebate/tabbycat/blob/da6ae86818ab8542346ca22fe5f9c9c5fc85a64b/tabbycat/adjfeedback/forms.py#L159-L162

Also, I should say that I find the new behavior intended by this PR less intuitive than before. The asterisk is still applied to the question text, and it's not clear what this or the "required" field of `AdjudicatorFeedbackQuestion` even means. Normally if a checkbox is marked "required", I take that to mean that I'm required to check it. I realize that interpretation might not be natural to people without a web design background, but required checkboxes also show this tooltip when you hover over them:
image

If the issue is that tab directors will sometimes make checkbox questions required when they just intended to me "I want _an_ answer", then we should somehow be preventing them from making the checkbox required in the first place, with some sort of warning message. We can add a custom element to the Edit Database area, similar to the existing ones that warn users about the need to specify own-institution conflicts explicitly. We could even go a step further and have the model fail validation if a checkbox question is required, but I'd prefer to preserve the option for CA teams who really do want to add some sort of "I understand that…" mandatory checkbox to their forms.

czlee:
Actually, I think I'm confused about what was broken before. Line 161 always explicitly sets the `required` attribute of the form field:

https://github.com/TabbycatDebate/tabbycat/blob/da6ae86818ab8542346ca22fe5f9c9c5fc85a64b/tabbycat/adjfeedback/forms.py#L159-L162

I (now) assume that if a user _wants_ a checkbox to be required, it was intended that they still be able to do so. But if they don't, line 161 was setting the `required` attribute to match what `AdjudicatorFeedbackQuestion.required` specified, and as far as I could tell, that was working. What have I missed?

PR #1299, opened 2019-11-09
Rename "test score" to "base score"

tienne-B:
As mentioned in https://github.com/TabbycatDebate/tabbycat/pull/762, "test score" could be a misnomer for tournaments that do not use adjudication tests to determine first score. For that, the term "base score" is more general for those cases, yet would still be understandable for tournaments with testing.

Migrations of models and methods using the old name have also been made for consistency of the term.


czlee:
I haven't taken this for a test drive yet, but in principle the terminology change is fine with me.

philipbelesky:
I'm happy with the terminology change also. I had a quick test of this and it all looks good; but its worth keeping an eye out for any changes that are harder to notice.

PR #1300, opened 2019-11-09
Remove data sandboxes

tienne-B:
Obsolete as we use the Django ORM for queries/aggregations and what was being tested has been implemented. Sandboxes not updated in years.


czlee:
Ha, this is a trip down memory lane 😅

PR #1303, opened 2019-11-17
Reduce size of judge feedback drop-down

tienne-B:
The drop-down to choose an adjudicator to give feedback contains all available applicable adjudicators, with positions. This commit makes the text shorter, removing the "chair rolled." This is for clarity between chairs and panelists when using phones, and drop-down items are scinded.

### DIFF
--- a/tabbycat/adjfeedback/forms.py
+++ b/tabbycat/adjfeedback/forms.py
@@ -305,7 +305,7 @@ def make_feedback_form_class_for_team(source, tournament, submission_fields, con
             # feedback expected only on orallist
             display = _("%(name)s (%(round)s — chair gave oral)")
         else:
-            display = _("%(name)s (%(round)s — chair rolled, this panellist gave oral)")
+            display = _("%(name)s (%(round)s — panellist gave oral as chair rolled)")
 
         display %= {'name': adj.name, 'round': debate.round.name, 'adjpos': ADJUDICATOR_POSITION_NAMES[pos]}
         return (value, display)


philipbelesky:
Where is the truncation normally? Is the idea here that people are better placed to see "... chair rolled" rather than "... panellist gave"?

tienne-B:
The truncation is most often after "chair." The idea is that people should see the position of the adjudicator before extra indications; that knowing the position is more important than "why this adj?" (Because the chair rolled)

czlee:
Oh, the verboseness there is kind of necessary. People are bad at understanding in what circumstances they are supposed to give feedback on judges, so it's meant to leave no doubt that they do not give feedback on panellists unless they gave the oral. (Even then, it still sometimes fails.)

If it's truncated that's obviously bad, but the remedy needs to be something that doesn't involve just removing the "chair rolled" indication. Maybe reversing the order: "panellist gave oral because chair was rolled"?

I think in my ideal world, there would be a first dropdown for the round, then a second that asks "who gave the oral?" providing options from that round, but I was too lazy/busy to code the JavaScript for that at the time (and wouldn't necessarily encourage this now, since that would probably just serve to increase the technical debt currently in adj feedback).

tienne-B:
Okay, I re-ordered the lines to say "chair rolled" after the adj position.

philipbelesky:
I'm happy with this change if you are @czlee

czlee:
Hmm, I'm going to nitpick a little. The semicolon could be interpreted by some to mean "or", rather than "and" or "because".

tienne-B:
How about: "%(name)s (%(round)s — panellist gave oral as chair rolled)"
You had anything in mind?

czlee:
I had "because" in mind, but "as" does nearly the same thing with fewer characters. (Nearly, because it could technically mean "while".) I approve. 🙂

PR #1306, opened 2019-11-26
Add 'base score' migration

tienne-B:
As not only the field is renamed but the verbose name changed. Forgot in the previous migration. Should it be squashed with the precedent migration?

### DIFF
--- a/tabbycat/adjfeedback/migrations/0006_auto_20191109_1240.py
+++ b/tabbycat/adjfeedback/migrations/0006_auto_20191109_1240.py
@@ -6,7 +6,7 @@ from django.db import migrations
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('participants', '0013_auto_20191109_1240'),
+        ('participants', '0013_rename_test_score'),
         ('tournaments', '0006_auto_20191109_1240'),
         ('adjfeedback', '0005_auto_20180928_1441'),
     ]
--- a/tabbycat/participants/migrations/0013_rename_test_score.py
+++ b/tabbycat/participants/migrations/0013_rename_test_score.py
@@ -15,4 +15,9 @@ class Migration(migrations.Migration):
             old_name='test_score',
             new_name='base_score',
         ),
+        migrations.AlterField(
+            model_name='adjudicator',
+            name='base_score',
+            field=models.FloatField(default=0, verbose_name='base score'),
+        ),
     ]


czlee:
In theory we shouldn't need to squash it; Django docs say not to worry about having too many migrations until you hit several hundred and want to squash it down. But I don't mind squashing it, or even just manually squashing it (as in, just modifying `participants.0013`), so long as no one's affected by it (_i.e._, if everyone who might've applied `participants.0013` already is also a developer).

tienne-B:
Squashed.

philipbelesky:
Sorry, could you update the migration's dependencies? I think a subsequent merge has meant it is is trying to apply itself in the wrong order.

tienne-B:
All dependencies are already the most recent.

philipbelesky:
Are you sure? I just made a new db, and performed all the migrations on `develop`. Changing to the PR branch and then migrating produced the following:

```
  File "/Users/philip/Sites/tc_dev/manage.py", line 18, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/base.py", line 83, in wrapped
    res = handle_func(*args, **kwargs)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/core/management/commands/migrate.py", line 90, in handle
    executor.loader.check_consistent_history(connection)
  File "/Users/philip/Sites/tc_dev/venv/lib/python3.6/site-packages/django/db/migrations/loader.py", line 299, in check_consistent_history
    connection.alias,
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration adjfeedback.0006_auto_20191109_1240 is applied before its dependency participants.0013_rename_test_score on database 'default'.
```

philipbelesky:
Related, I think, running `makemigrations` on develop produces:

```
Migrations for 'participants':
  tabbycat/participants/migrations/0014_auto_20191228_2108.py
    - Alter field base_score on adjudicator
```

tienne-B:
The error is not because there are wrong dependencies; the error is caused as you're running migrations on `develop` first, whereas this patch modifies an existing migration. That also explains why making migrations on `develop` yields a new migration; this patch corrects that.

tienne-B:
I would very much prefer to keep the amendment in the same migration, but I suppose you'd rather it be in a new one?

philipbelesky:
Ok, think I got it. Happy to modify the prior migration.

philipbelesky:
Hmm, even on a fresh db this still shows an inconsistent migration history after this has been merged:

```
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration adjfeedback.0006_auto_20191109_1240 is applied before its dependency participants.0013_rename_test_score on database 'default'.
```

PR #1309, opened 2019-11-30
Remove speakerscorebyadj prefetch in adjfeedback

tienne-B:
Removed as it seems extraneous and unused anywhere.


czlee:
Ah, I see, it was used in the average margin column, which was deprecated in 405fbb7a7dd3bf89eb8e204cb28bd39cb6bdd47e. Just needed to search my memory for why it was there in the first place. Thanks, @tienne-B!

PR #1329, opened 2019-12-04
API Endpoints for Teams, Institutions, Adjudicators and Speakers

### LINKED ISSUE #1301: Create participant import API endpoints
For integration with external registration applications, there could be Tabbycat API `POST` methods to create these objects from registration.

jacklishufan:
It is extremely arduous to get the data into the right format for tabbycat when you have it ready to go on another server already. These APIs may make things a lot easier.

Closes #1301.


tabbycat/participants/api_import.py
```diff
@@ -0,0 +1,100 @@
...
+    password = request.POST.get('password','')
+    user = authenticate(username=username, password=password)
+    print(user)
+    if not user:
+        return HttpResponseBadRequest("BAD REQUEST:AUTHENTICATION FAILED")
+    if not user.is_staff:
+        return HttpResponseBadRequest("BAD REQUEST:NO PERMIT")
+
+
+def api_create_institution(request,**kwargs):
```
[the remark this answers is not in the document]
jacklishufan: It has been integrated using APIView from rest_framework.views

docs/use/api-importation.rst
```diff
@@ -0,0 +1,59 @@
...
+    Username of a staff account
+- password:
+    Password of the staff account
+- code_name:
+    Team code name for identification, it must be unique
+- reference
+    Reference of the team
+- institution
+    The name of institution
+- type (Optional)
```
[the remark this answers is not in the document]
jacklishufan: > Hey! This is great! Thanks for this work. I've left a few comments, of which should be generalizable to the extent of the whole patch.
> 
> I must say this is not quite what I expected with #1301; where we would be using django REST framework (like what we already do with serializers) to get these POST views. What I had in mind were three JSON endpoints:
> 
> * adjudicators:
>   
>   * name
>   * email (optional)
>   * institution (optional; institution code)
>   * base score (default 0.0)
>   * gender (optional)
>   * pronoun (optional)
> * institutions:
>   
>   * name
>   * code
> * teams:
>   
>   * reference (called name; short_reference created from that)
>   * uses_institution_prefix (default true)
>   * code (optional)
>   * institution (code)
>   * speakers (as list)
>     
>     * name
>     * email
>     * gender
> 
> The big difference would be merging the teams/speaker endpoints.
> 
> Comments on this conception of the import API?

Using JSON instead of Form Post does make more sense as it allows multiple data submissions in one POST request. Speaking of which, I find it better to use one endpoint for data submission instead of three as it allows the client to submit all relevant information regarding the registration of one institution in one POST request.

Here is a sample JSON data 

`{"institutions":[
	{ "name": "PK UNIVERSITY", "code":"PKU" }
	],
"teams":[
	{ "name": "PKU A", "uses_institution_prefix":"TRUE","code":"PKU A CODE","institution":"PKU",
	"speakers":[
		{"name":"Jason Wang", "email":"123@126.com","gender":"M"}
		]
	}
	]}`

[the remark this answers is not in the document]
tienne-B: Great! One thing: If using the institution prefix, the reference would just be "A", not "PKU A". Looks awesome!

tienne-B:
I've been looking a bit at what we already have in Tabbycat that could help with this. We already use django-rest-framework, especially for its model serializers. Instead of parsing the JSON explicitly, what are your thoughts on using serializers (like modelSerializer) and saving objects through that? Could be less error-prone and easier to implement than managing the fields "manually."

jacklishufan:
> I've been looking a bit at what we already have in Tabbycat that could help with this. We already use django-rest-framework, especially for its model serializers. Instead of parsing the JSON explicitly, what are your thoughts on using serializers (like modelSerializer) and saving objects through that? Could be less error-prone and easier to implement than managing the fields "manually."



> I've been looking a bit at what we already have in Tabbycat that could help with this. We already use django-rest-framework, especially for its model serializers. Instead of parsing the JSON explicitly, what are your thoughts on using serializers (like modelSerializer) and saving objects through that? Could be less error-prone and easier to implement than managing the fields "manually."

One of the issue with existing ModelSerializers are with ForeignKey / PrimaryKeyRelatedField. The primary key of Institution, for instance, is its id. So a team would be serialized as { ....'institution': 43,.....}. This could make it more complicated to add new teams for an existing institution. (Obviously, we would want the JSON data to be legible)

I propose three solutions:
1. Add additional APIs to allow the client to fetch the id of an institution by its name from the server, so that it could prase proper JSON for the existing serializer.
2. First manually translate institution names to existing ids (get or filter or whatever) and pass it to the ModelSerializer.create method.
3.Edit existing serializer or create new ones for APIs (which totally miss the point of 'integration')

In fact, such an issue usually would not exist in the first place as we could simpler structure the models, serializers in a different way, and maybe rewrite create method. However, given the existing construction of serializers, I see manually intervention as inevitable and necessary. I believe rewriting serializers or the create method could generate unforeseen problems as they were also used elsewhere (Importer? as I believe).

In conclusion, I find it better to simply 'manually' handle all the field instead of some for the sake of consistency. Amongst the three proposals which employ serializers, I find the third one preferable. 

Please tell me what you think.

tienne-B:
The foreign key relationships problem can easily be solved by avoiding `PrimaryKeyRelatedField` and instead using [`SlugRelatedField`](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) with the field name desired.

In your list of options to use serializers, the first two are discounted above. For the third, I don't really think DRY ("integration") really applies here as the uses of the serializers differ, but even then, you should be able to subclass existing serializers, just adapting them to this new use-case.

Especially how the serializers we currently have are for allocation views (none in `importer/`), if the serializers there aren't structured for data import, that shouldn't go against using the same framework in a different context.

Always try to find what has already been done for you ;) I don't see why "manual intervention [is] inevitable and necessary." Further, the serializers would handle all fields; that wasn't in question.

tabbycat/participants/api_import.py
```diff
@@ -0,0 +1,51 @@
...
+        if not user.is_staff:
+            return HttpResponseBadRequest("BAD REQUEST:NO PERMIT")
+        data = request.body.decode('utf-8')
+        received = json.loads(data)
+        received['tournament'] = kwargs['tournament_slug']
+        institutions_data = received.get("institutions",[])
+        teams_data = received.get("teams",[])
+        adjudicators_data = received.get("adjudicators",[])
+        serializers = []
+        for i in institutions_data:
```
tienne-B: You won't need any loops, but for future projects, don't think of "`i`" as in "iteration" but "increment" or "index". As such, I think of `i` as a number. Each iteration is not a number, but an object. Give them better names ;)

tabbycat/participants/api_import.py
```diff
@@ -0,0 +1,51 @@
...
+            serializers.append(InstitutionSerializerImport(data=i))
+        for i in serializers:
```
tienne-B: Don't need another loop, just save each as they come in the previous loop. You wouldn't need the `serializers` list.

czlee:
Hi @jacklishufan, thanks for your work on this so far! I'm not sure if you check your school email during the break, but I sent you an email asking to discuss this, mind dropping me a note there when you get a chance? 😊

tabbycat/api/urls.py
```diff
@@ -0,0 +1,16 @@
...
+from . import views
+
+
+urlpatterns = [
+
+    path('t/<slug:tournament_slug>/', include([
+        path('participants/', include([
+            path('speakerlist',
+                views.InstitutionViewSet.as_view({'get': 'list', 'post': 'create'}),
+                name='api-speakers')
```
tienne-B: This is to show and create institutions, not speakers.

tienne-B:
Hope you aren't just abandoning the work you've done for team/adjudicator import! Another thing is  that if the institution API is tournament-specific, the team set should be filtered for that tournament.

jacklishufan:
> Uh, would you mind merging in my PR branch so you can work on top of it, rather than you re-using some of the work and causing conflicts?

Sure, that would be wise.


> Hope you aren't just abandoning the work you've done for team/adjudicator import! Another thing is that if the institution API is tournament-specific, the team set should be filtered for that tournament.

No. But I think I should make sure one API works properly before proceeding to the next, especially given that the overarching framework (i.e. your PR) has not been finalized

tabbycat/api/serializers.py
```diff
@@ -64,3 +118,51 @@ def update(self, instance, validated_data):
...
+        speaker_data = validated_data.pop('speakers')
+        team = Team.objects.create(**validated_data)
+        for i in speaker_data:
+            Speaker.objects.create(team=team, **i)
+        return team
+
+
+class InstitutionSerializer(serializers.ModelSerializer):
+
+    teams = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
jacklishufan: According to Django reference, `queryset` is "used for model instance lookups when validating the field input. " It does not serve as a filter for representations. I did notice your implementation for SpeakerEligibilitySerializer, though.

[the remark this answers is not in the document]
tienne-B: (Django REST reference) I see. Now I have stuff to fix! :) Good catch; I was oblivious to this problem, but I found a more elegant solution.

I am still opposed to using `MethodField` as it is very low-level, thus requires us to create our own logic. It may also prevent us from using the field for writing, if that becomes a concern. Using a method field is inelegant...

To avoid needing it, in `InstitutionViewSet`, you can use the `.prefetch_related()` method of querysets with a [`Prefetch()`](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-objects) object to filter `team_set` to only the relevant teams (like `self.tournament.team_set.all()`). It is also better for the DB as it will require fewer queries.

tabbycat/api/views.py
```diff
@@ -61,3 +65,21 @@ class SpeakerEligibilityView(TournamentAPIMixin, AdministratorAPIMixin, Retrieve
 
     def get_queryset(self):
         return super().get_queryset().prefetch_related('speaker_set')
+
+
+class InstitutionViewSet(TournamentAPIMixin, AdministratorAPIMixin, ModelViewSet):
+    serializer_class = serializers.InstitutionSerializer
+    lookup_field = 'id'
```
tienne-B: No need to specify the lookup field as the primary key is the default.

jacklishufan: This results in an error as the application won't know which argument to look for in the url. 
In urls,it is declared as `<int:id>`
In built-in GenericAPIView, we have 
`lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field`
So either we change the url declaration to `<int:pk>` or we change the line in question from `lookup_field = 'id'` to `lookup_url_kwarg = 'id`, or keep it as it is. Deleting it is not a viable option. I don't see any reasons for us to prefer one choice over the rest two?

tienne-B: Using `<int:pk>` in urls is preferable. Then, we could delete this line ;)

tabbycat/api/serializers.py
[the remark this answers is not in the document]
jacklishufan: I decided to mark this conversation as unresolved because Prefetch does not work properly for detail views. In detail views for institutions, all teams get returned while in list views, they get filtered beforehand. This probably has something to do with the mechanism of Prefetch. We need either overwrite `retrieve()` method or `get_object()` method in the `InstitutionViewSet `, or reemploy `MethodField`.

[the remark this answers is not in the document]
tienne-B: I haven't tested this with Django REST, but this problem would be very odd as we have the expected result when doing `.get()` on a query set with a prefetch (that I tested), and DRF does just that by using `.get()` on the queryset.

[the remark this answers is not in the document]
jacklishufan: There was a mismatch between my test data and my test database, this was a false alarm

tabbycat/api/serializers.py
```diff
@@ -105,3 +117,67 @@ def update(self, instance, validated_data):
...
+class SpeakerSerializer(serializers.ModelSerializer):
+    url = SpeakerHyperlinkedIdentityField(view_name='api-speaker-detail')
+    categories = TournamentHyperlinkedRelatedField(
+        many=True,
+        view_name='api-speakercategory-detail',
+    )
+
+    class Meta:
+        model = Speaker
+        fields = ('url','id','name', 'gender','email','phone','anonymous','pronoun',
```
czlee: We normally try to keep spaces after commas. I don't know why it's disabled in flake8 at the moment, sorry, our bad. I just wanted to let you know that I'm about to reenable it.

Will look through the rest later, but at first glance this looks great! Thanks heaps for helping finally make this happen!

czlee:
Thanks! Looks good at first glance, I'll have a proper look at this for merging when I next get a chance, which might not be for a couple of weeks (I'm travelling) but I'll do it as soon as I can.

tabbycat/api/views.py
```diff
@@ -80,3 +88,32 @@ class SpeakerEligibilityView(TournamentAPIMixin, AdministratorAPIMixin, Retrieve
...
+
+    def get_queryset(self):
+        return Institution.objects.all()
+
+
+class SpeakerViewSet(TournamentAPIMixin, AdministratorAPIMixin, ModelViewSet):
+    serializer_class = serializers.SpeakerSerializer
+
+    def get_queryset(self):
+        return Speaker.objects.all()
```
tienne-B: Should filter by tournament.

jacklishufan:
FYI These endpoints are being tested in Berkeley IV 2020 , where the database is synchronized with a google sheet through Google Apps Script.

tabbycat/api/serializers.py
```diff
@@ -105,3 +118,77 @@ def update(self, instance, validated_data):
...
             self.instance.speaker_set.set(speakers)
         return self.instance
+
+
+class SpeakerSerializer(serializers.ModelSerializer):
+    url = SpeakerHyperlinkedIdentityField(view_name='api-speaker-detail')
+    categories = TournamentHyperlinkedRelatedField(
+        many=True,
+        view_name='api-speakercategory-detail',
+        queryset=SpeakerCategory.objects.all()
```
czlee: We use trailing commas when the closing parenthesis/bracket/brace isn't on the same line as the last time (see [this section of PEP 8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)) 😎

Our bad for not including it in the flake8 checks 😳 I'll fix this and add the flake8 checks, I just thought I should give a heads-up for next time!

PR #1333, opened 2019-12-11
Remove deprecated methods with expired removal dates

tienne-B:


philipbelesky:
Great, thanks for cleaning these up!

PR #1334, opened 2019-12-15
Re-work SingleObjectByRandomisedUrlMixin

tienne-B:
As private URLs are all by-tournament yet URL keys are not dependent on the tournament, it should not inherit from SingleObjectFromTournamentMixin. This change also lets us better use the SingleObjectMixin for private URLs.


czlee:
I'm a bit confused about why it shouldn't inherit from SingleObjectFromTournamentMixin? It's true that the restructure means that object is always a Person, but it's still an object restricted to be from a given tournament, so a special case of SingleObjectFromTournamentMixin, no?

tabbycat/results/views.py
```diff
@@ -620,11 +618,10 @@ def get_context_data(self, **kwargs):
         return super().get_context_data(**kwargs)
 
 
-class PrivateUrlBallotScoresheetView(RoundMixin, SingleObjectByRandomisedUrlMixin, BasePublicBallotScoresheetsView):
+class PrivateUrlBallotScoresheetView(RoundMixin, BasePublicBallotScoresheetsView, SingleObjectByRandomisedUrlMixin):
```
czlee: This inheritance order normally isn't best, normally you want all mixins to come ahead of the base view in the MRO. What needed this to make it work?

tienne-B:
Because a `Person` doesn't have a direct `tournament` field, inheriting from `SingleObjectFromTournamentMixin` would error with the inherited `.get_queryset()`. As for the inheritance order, it's also for the queryset function, but here, it should be inherited from the Feedback view.

tabbycat/participants/tables.py
```diff
@@ -56,6 +56,8 @@ def get_table(cls, view, participant):
                 queryset=DebateAdjudicator.objects.select_related('adjudicator__institution')),
             'debate__debateteam_set__team__speaker_set',
             'debate__round__motion_set',
+            Prefetch('debate__ballotsubmission_set',
+                queryset=BallotSubmission.objects.exclude(discarded=True)),
```
czlee: I definitely like the efficiency thought here! But since it's used in a different file (utils/tables.py) and returns a different collection of objects to what `ballotsubmission_set` normally would, I'd prefer it if this particular call to `Prefetch()` used the `to_attr=` argument to put it in a different attribute name. As well as being explicit, this is a fail-fast measure: since `TabbycatTableBuilder.add_debate_ballot_link_column()` no longer filters out discarded ballots, if it's (now or in the future) accidentally called by a table that _hasn't_ made the effort to filter out discarded ballots ahead of time, I'd like it to fail straight away (rather than wait for some unsuspecting tournament to discover the bug).

czlee:
>Because a `Person` doesn't have a direct `tournament` field, inheriting from `SingleObjectFromTournamentMixin` would error with the inherited `.get_queryset()`.

Oh, I see what you're saying. This is interesting. In principle there is a way to filter for persons from a given tournament:
``` python
Person.objects.filter(Q(speaker__team__tournament=self.tournament) | Q(adjudicator__tournament=self.tournament))
```
but it doesn't really fit cleanly into the paradigm of `SingleObjectFromTournamentMixin` as it stands.

I actually think this raises a lot of interesting questions about what the `Person` model is for. You can imagine a world in which `Person` represents a person, who might be speaking or adjudicating in multiple tournaments, but I don't think that's really what we (or @qlim) had in mind when choosing this structure—in fact, our schema only permits a person to be associated with one speaker or one adjudicator (or both, but that's technically a bug).

What I had in mind when I was confused was: Every time a participant accesses a private URL page of theirs, the URL in question is "under" a tournament, right? I tend to think that if the URL relates to a tournament, then we should also be filtering by tournament—that is, if a participant changes the tournament slug of the URL to a tournament they have nothing to do with, then the server should respond with a 404.

But then, how to implement in this case? One way would be to permit `tournament_field_name` in `SingleObjectFromTournamentMixin` to be a list, and if it is a list, then to interpret the list as "if any of these fields match the expected tournament". (The pattern would be similar to how `allow_null_tournament` currently works.) But I don't know if this is too contorted for just the `Person` model, just because it happens not to cleanly associate with a tournament…

And also, intuitively, given how our model structure works, a `Person` probably _should_ be able to associate cleanly with a tournament, in general. The concept is well-defined (as in the query above), it's just that currently you have to look in two places to find it. 🤔

Sorry to be raising more questions than answers! It took me a while to understand, but you've hit upon a great point.

PR #1336, opened 2019-12-17
Associate venue categories with tournaments

### LINKED ISSUE #469: Venue Categories visual importer should sort venues by display name and exclude venues from other tournaments
Currently venues from other tournaments show up in the `venues` field when editing Venue Categories, and remaining venues are not sorted alphabetically. I know `get_formset_queryset()` exists to override the queryset for the main model but couldn't easily figure out how to do it for a field. For australs purposes I've done so manually but it would be good to address this for future large tournaments.

tienne-B:
In addition to venues themselves, their categories should be associated with particular tournaments. This prevents the formset from being over-loaded with other tournaments' categories, which also has the side-effect of removing other tournaments' venues from their categories with saves.

In addition, the help text for venues and their tournament is removed as shared venues/adjudicators were removed with divisions, as well as setting tournament as blank=False.

Resolves #469.


philipbelesky:
Thanks for adding this. There may be some use case that still uses venue categories in this way (outside of super-large leagues) but I agree it's not worth keeping those kind of multi-tournament hacks around.

tienne-B:
Must hurt to use venue categories like that, seeing that the effort required to create a category and to add venues to one (or many) is the same, but with the problem of potentially deleting category-venue relationships from other tournaments...

PR #1337, opened 2019-12-18
Create VenueCategorySerializer for checkins

tienne-B:
The Venue Checkin Statuses page has been broken for some time because the serializer for the venue categories of each venue was culled with the transition towards the REST framework's serialization for objects.

This commit adds the 'categories' field to the VenueSerializer by creating a new VenueCategorySerializer.


philipbelesky:
Thanks for noticing, and for the fix!

PR #1028, opened 2019-02-10
Change licence to AGPLv3 and update donations text

philipbelesky:
As discussed, this shifts Tabbycat to using a proper OSI licence. This is incompatible with our previous approach of requiring donations for profit/fundraising use, so I have updated the docs. I've also added a note about the possibility of dual-licensing for people who would like to avoid the gpl terms.

Happy to discuss here or over email; I just wanted to use the PR to ensure we are all onboard before merging.

PR #1338, opened 2019-12-19
Replace "open-source" by "source-available"

tienne-B:
This is a distinction between the current licensing scheme being non-libre, but the source code and development being done openly. This change reflects the README which states it is not open-source (yet). Open-source is a subset of source-available.

Should be reverted with #1028.


philipbelesky:
I suspect the terms here may be a little confusing to some people, but happy to work these in an hopefully resolve #1028 soon. @czlee can you merge it in if you don't have objections?

PR #1342, opened 2019-12-24
Create basic default settings for REST

tienne-B:
This commit adds REST as an installed application and defines some settings such as the serializer format and authentification.

It also enables a browsable API view in the development settings profile.

The commit also includes a proof-of-concept of the discussion on #1343 using speaker/break categories.

### DIFF
--- a/tabbycat/api/apps.py
+++ b/tabbycat/api/apps.py
@@ -1,0 +1,7 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class ApiConfig(AppConfig):
+    name = 'api'
+    verbose_name = _("Application Programming Interface")
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -1,0 +1,66 @@
+from rest_framework import serializers
+
+from breakqual.models import BreakCategory
+from participants.models import Speaker, SpeakerCategory
+
+
+class BreakCategorySerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = BreakCategory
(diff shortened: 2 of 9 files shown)


tienne-B:
Something that will be important is an API reference, and preferably have it auto-generated, but I've been having problems with DRF's OpenAPI generator.

jacklishufan:
> Something that will be important is  an API reference, and preferably have it auto-generated, but I've been having problems with DRF's OpenAPI generator.

Thanks for the work, this is a good starting point for other API endpoints. As for reference, I am thinking of using apistar as mention in the [https://www.django-rest-framework.org/community/3.9-announcement/#built-in-openapi-schema-support](url)

tienne-B:
Getting a bit ahead of yourself there; generating the schema itself is problematic, using the schema in a webpage is not a concern (for now).

jacklishufan:
> Getting a bit ahead of yourself there; generating the schema itself is problematic, using the schema in a webpage is not a concern (for now).

Is there a problem with the generate schema command?

czlee:
Thanks, this is a great start! Happy to merge, if @tienne-B is okay with 105ae02 (let me know).

I don't have strong opinions about tournament and round prefixes, but we should watch out in case they get confusing. The round prefixes are more likely to, though their being numeric should help.

We should be using hypertext/HATEOAS, which the library doesn't do out-of-the-box, intentionally:
https://www.django-rest-framework.org/topics/rest-hypermedia-hateoas/
Since that involves design questions, it's a little nontrivial. Think of it as a next step, I don't intend for it to block merging this.

(I couldn't really follow the discussion above about schemas, sorry.)

PR #1344, opened 2019-12-26
Work on venue allocation highlights

tienne-B:
Read commit messages.


philipbelesky:
Was this ready for review/merging btw? I've had a quick look at the commits and implementation, but wasn't sure if you had more planned.

tienne-B:
Yes it is.

philipbelesky:
Great, thanks for adding this in. I couldn't get the category highlights working on my instance, so ended up re-implementing most of your changes from scratch to debug and remind myself of how the code worked. I ended up altering some of the implementation in a way that I think has less special case handing within Vue and python. 

Anyway, the commit for that is [over here](https://github.com/TabbycatDebate/tabbycat/commit/b5de7935596219e65344a8c5a08cfa1cc1ea8743). Should I merge that in, or did we want to try and preserve the edit history here and update this branch instead?

tienne-B:
I would prefer that it updates this branch, as it also has the removal of breaks, and it shows my progress in understanding Vue ;)

philipbelesky:
Ok, great, I'll merge it into this branch then merge this in.

PR #1345, opened 2019-12-27
Create view to create/edit break categories

### LINKED ISSUE #1335: Create view to add break categories
Break categories should be able to be created outside the database interface, as done with speaker categories. The benefit of having this interface is to automatically create the associated elimination rounds (on account of the number of teams), which would facilitate having multiple break categories, without the need to manually create rounds. As for the names of the rounds, "Open" could be substituted with the break name (different path for i18n).

tienne-B:
This commit adds a new interface to create and edit break categories, modelled on the creation of speaker categories. While traditionally done in the database area, the creation of break categories has more steps than just creating a model object, so is best done interactively.

When creating a break category, rounds are now automatically created for the number of teams breaking, using the same method for the open break rounds as in tournament creation. In order to have the appropriate round names, the round names have the category name, and the abbreviation has the first letter of it.

A method for modifying existing objects has been also subclass-ed for future use (if need to delete/append new rounds due to a change in break size).

The documentation has also been updated to reflect this new procedure.

Closes #1335.

### DIFF
--- a/docs/features/breaks.rst
+++ b/docs/features/breaks.rst
@@ -74,7 +74,7 @@ which is the default.
 Setting up break categories and rounds
 ======================================
 
-For each break category in your tournament, you need to do three things:
+For each break category in your tournament, you need to do two things:
 
   1. Create (and name) a break category
   2. Create break rounds for the category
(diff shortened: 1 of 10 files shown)


philipbelesky:
Hey, this looks great. Let me know when it's ready to check out?

tienne-B:
Ready.

philipbelesky:
Thanks for adding this in! I've taken it for a spin and everything looks great.

PR #1350, opened 2019-12-29
Mark private URL explainer as a warning

tienne-B:
The private URL landing page contains an explainer of the page with a warning about unauthorized access. Other pages, such as the private URL view of scoresheets, have similar warnings. This commit standardizes the type of the messages, changing 'info' to 'warning' on landing pages in line with scoresheets.

This commit also removes the note about tournament rules in the scoresheets as it is redundant with the explication why sharing the credential is bad, without resorting to authority.

Also removed an unused if statement.

I am also kinda against the "and/or" construction as the operator "or" does include the set covered by "and," but it's just my logician side, and I haven't replaced its usage in private URLs in this commit.

### DIFF
--- a/tabbycat/privateurls/templates/public_url_landing.html
+++ b/tabbycat/privateurls/templates/public_url_landing.html
@@ -11,9 +11,6 @@
   {% if object.speaker %}
     {% blocktrans trimmed with team=object.speaker.team.short_name %}({{ team }}){% endblocktrans %}
   {% endif %}
-  {% if object.speaker %}
-
-  {% endif %}
 {% endblock %}
 
 {% block page-alerts %}
@@ -24,7 +21,7 @@
     feedback for your debates. You may bookmark this page and return here after each
     debate for the available actions.
   {% endblocktrans %}
-  {% include "components/explainer-card.html" with type="info" %}
+  {% include "components/explainer-card.html" with type="warning" %}
 
 {% endblock %}
 
--- a/tabbycat/results/templates/privateurl_ballot_set.html
+++ b/tabbycat/results/templates/privateurl_ballot_set.html
@@ -9,8 +9,7 @@
     your own scoresheet.
     <strong>You must not share this URL with anyone.</strong>
     Sharing the URL will allow others to access <strong>all</strong> actions
-    from your personal landing page. Also, allowing others to see your
-    scoresheet may be a violation of tournament policies.
+    from your personal landing page.
   {% endblocktrans %}
   {% include "components/explainer-card.html" with type="warning" %}
 {% endblock %}


czlee:
Making it a warning sounds good!

I think with the tournament policies remark, I was worried that some user might not comprehend the technical bit, and since it is in fact true that it contravenes most tournament policies, that part might help get them not to do it. So although logically it's redundant, when working with humans it might not be. But I don't mind dropping it if you feel strongly about it—worst case someone screws up once, their URLs get compromised, people get angry, some tab director has to make an announcement, and word gets around not to try.

I think the "and/or" is there because sometimes only one will be applicable and sometimes both will. Technically in that sentence, both "and" and "or" will work fine (actually, if I had to use just one, "and" would be more intuitive to me), if both are applicable. If we can be bothered with the intelligence, I guess the ideal is to change the sentence according to what's applicable (i.e., choose between several sentences, because translation).

philipbelesky:
I'm happy to have this merged, I'll leave this to you @czlee to merge in case the wording changes needed to be hashed out more.

tienne-B:
I'm fine with leaving and/or, although it is clunky to me (thus I haven't changed it in this patch).

I don't find "Sharing the URL will allow others to access *all* actions from your personal landing page." or "*Do not share it with anyone;* anyone who knows this URL can submit results and/or feedback for your debates." to be technical. It states the clear warning of the effect sharing would have; it doesn't need to show the mechanism of the shared key or the anger of the officials to be persuasive.

tienne-B:
@philipbelesky: In the check-in info box, is there a "type='success'" to contrast with the warning type when not checked-in?

philipbelesky:
@tienne-B sorry which info box are you talking about? I suspect the relevant component to use for notifications of a completed action would be `"components/alert.html"` with a success type.

tienne-B:
For the box indicating that a person is currently checked-in, but isn't a button.

czlee:
>I don't find "Sharing the URL will allow others to access all actions from your personal landing page." or "Do not share it with anyone; anyone who knows this URL can submit results and/or feedback for your debates." to be technical.

You're not most people 😉

Though for me it's more, people might not see the link between "all actions" and "results and feedback", or maybe they just trust someone not to abuse their link to make fraudulent submissions. Anyway, there's no way to cover all our bases without expanding the text even further, and I'm happy just to remove that line to keep things succinct. We'll revisit this if we catch wind of some incident concerning the pages—in the unlikely event, for if this were likely, it probably would've happened by now.

czlee:
>In the check-in info box, is there a "type='success'" to contrast with the warning type when not checked-in?

Do you mean this one? It's currently `type='info'` by default (not pictured), but `type='success'` (pictured) totally works if you think we should do that:

image

tienne-B:
Exactly! I never saw it being used elsewhere, and wasn't with the other info cards in the style page, so just assumed that it wouldn't work. I'll make a patch for it in the morning (for me).

philipbelesky:
As an FYI, I've broadly reserved the success coloring for the completion of user-initiated actions. The time delay / remove of the checkin process doesn't strictly meet that criteria, but since it only returns checkins within the window I think it can suit this context if needed.

Issue #1343, opened 2019-12-24
Where should the API be located


tienne-B:
With the current work in a Tabbycat API, we should have a standard pattern for the API endpoints. The main decision would be where should the endpoints be located. Should they be:
* In a new app, like notifications
* In their constituent apps

I'd suggest separating the API code to a folder "api" of each app.

czlee:
Great question.

Starting from the URI: I think my first instinct is to have an "/api/" straight after the root—like, before the tournament name, e.g. "https://mytournament.herokuapp.com/api/". It might be "/api/v1/", if we want to do versioning in the URIs.

Of course, this can be done either in a standalone app or constituent apps. It's just a little bit easier in a standalone app, because we wouldn't have to do the urls.py gymnastics we do with the /admin/ URLs. I think the better answer's not that clear to me, but if it's a truly RESTful API, putting the API in its own app wouldn't be illogical if it doesn't have too much shared functionality beyond models.

Unless we think the serializers would actually be shared?

tienne-B:
I don't think serializers would be shared.

Other things to consider:
* The use of keys in APIs, whether to prefer asking for foreign keys or more readable fields (_i.e._ names instead of ids).
  I think names would be preferable so that less information (fewer requests) need to be made to be able to modify elements through the API. It is also easier to write without remembering codes.
* What formats to accept. There is a brief discussion of that in #911, between GraphQL, CSV, and JSON. With Django REST, `GET` requests could return both CSV and JSON with a `&format=` parameter.
  While data could be taken in the form of CSV, creating and updating objects should be in JSON.
* What to do with the makeshift API endpoints. We have basic CSV API endpoints for adjudicator scores and lists of feedback.
  Should deprecate and remove them, and possibly create other ways to access that information, without being beholden to the existing structure. The bulk create text boxes used for updating base scores as well as creating venues and institutions as well (not necessarily with an API endpoint in the admin interface).

Pinging @nickdbush and @jacklishufan as they have expressed interest and work in the development of the API.

jacklishufan:
On choices of keys, I prefer legible strings. However, I do see a problem given the way models are structured right now, as many of the name fields ('reference','name','code', or whatever) which could be used for API are not unique. For instance, neither institution names or codes are unique—they are 'unique together'. In an unlikely scenario where there are two institutions with the same code but different names, using code as the identifier would be clearly problematic. This applies to teams and speakers as well, where it is very likely for different objects to have the same name/reference.

[...]

PR #1351, opened 2019-12-29
Tournament endpoints and hyperlinked API

czlee:
I extended on @tienne-B's work in #1342, and added an API root and endpoints for tournaments to make it more compliant with the HATEOAS part of REST API guidelines.

One question raised by the existing structure, also touched on briefly in #1342, is where tournaments are located. Here are a few observations:

- The most _conventional_ naming would go:
    - **api/v1/tournaments** to list and create tournaments
    - **api/v1/tournaments/SLUG** to detail and update tournaments
- If we instead go **/api/v1/SLUG**, as we have currently, then the list of tournaments might more naturally go at the API root.
- Tournaments are so central to how we think about our data structures that it might actually be more intuitive to put the list at the API root, breaking REST API naming conventions. I've done this for now.
    - However, putting the list of tournaments at the API root means we can't really do pagination effectively, because pagination only really makes sense for list views.
    - On the other hand, maybe we should just never paginate this, and instead assume that there will never be enough tournaments on a site to paginate?
- We have to have a tournament creation endpoint somewhere, and the most conventional place is at "/tournaments". But if we're going to break this convention, maybe we should go the whole way and match the "/create/" URL that we have in the human user interface? I've done this for now.

Relates to #1301, #1343.

Ping @jacklishufan

### DIFF
--- a/tabbycat/api/fields.py
+++ b/tabbycat/api/fields.py
@@ -1,0 +1,21 @@
+from rest_framework.relations import HyperlinkedIdentityField
+from rest_framework.reverse import reverse
+
+
+class TournamentHyperlinkedIdentityField(HyperlinkedIdentityField):
+
+    def get_url(self, obj, view_name, request, format):
+        lookup_value = getattr(obj, self.lookup_field)
+        kwargs = {
+            'tournament_slug': obj.tournament.slug,
+            self.lookup_url_kwarg: lookup_value,
+        }
+        return reverse(view_name, kwargs=kwargs, request=request, format=format)
+
+    def get_object(self, view_name, view_args, view_kwargs):
+        lookup_value = view_kwargs[self.lookup_url_kwarg]
+        lookup_kwargs = {
+            'tournament__slug': view_kwargs['tournament_slug'],
+            self.lookup_field: lookup_value,
+        }
+        return self.get_queryset().get(**lookup_kwargs)
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -7,69 +7,56 @@ from tournaments.models import Tournament
 from .fields import TournamentHyperlinkedIdentityField
 
 
-class TournamentAtRootSerializer(serializers.HyperlinkedModelSerializer):
+class TournamentSerializer(serializers.ModelSerializer):
 
     url = serializers.HyperlinkedIdentityField(
         view_name='api-tournament-detail',
         lookup_field='slug', lookup_url_kwarg='tournament_slug')
 
-    class Meta:
-        model = Tournament
-        fields = ('name', 'short_name', 'slug', 'url', 'active')
+    class TournamentLinksSerializer(serializers.Serializer):
+        break_categories = serializers.HyperlinkedIdentityField(
+            view_name='api-breakcategory-list',
+            lookup_field='slug', lookup_url_kwarg='tournament_slug')
+        speaker_categories = serializers.HyperlinkedIdentityField(
+            view_name='api-speakercategory-list',
+            lookup_field='slug', lookup_url_kwarg='tournament_slug')
 
-
-class TournamentEndpointsSerializer(serializers.Serializer):
-
-    break_categories = serializers.HyperlinkedIdentityField(
-        view_name='api-breakcategory-list',
-        lookup_field='slug', lookup_url_kwarg='tournament_slug')
-    speaker_categories = serializers.HyperlinkedIdentityField(
-        view_name='api-speakercategory-list',
-        lookup_field='slug', lookup_url_kwarg='tournament_slug')
-
-
-class TournamentSerializer(serializers.ModelSerializer):
-
-    urls = TournamentEndpointsSerializer(source='*', read_only=True)
+    _links = TournamentLinksSerializer(source='*', read_only=True)
 
     class Meta:
         model = Tournament
-        fields = ('name', 'short_name', 'slug', 'seq', 'active', 'urls')
-
-
-class BreakCategoryEndpointsSerializer(serializers.Serializer):
-
-    eligibility = TournamentHyperlinkedIdentityField(
-        view_name='api-breakcategory-eligibility', lookup_field='slug')
+        fields = ('name', 'short_name', 'slug', 'seq', 'active', 'url', '_links')
 
 
 class BreakCategorySerializer(serializers.ModelSerializer):
 
+    class BreakCategoryLinksSerializer(serializers.Serializer):
+        eligibility = TournamentHyperlinkedIdentityField(
+            view_name='api-breakcategory-eligibility', lookup_field='slug')
+
     url = TournamentHyperlinkedIdentityField(
         view_name='api-breakcategory-detail', lookup_field='slug')
-    urls = BreakCategoryEndpointsSerializer(source='*', read_only=True)
+    _links = BreakCategoryLinksSerializer(source='*', read_only=True)
 
     class Meta:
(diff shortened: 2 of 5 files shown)


czlee:
Oh, another thing that's up for contest: I've been putting URLs that give more information about the object at hand, in a nested `"urls"` object. I find this more intuitive than, say, having `"break_categories_url"` and `"speaker_categories_url"` everywhere. But the library's not really cut out for this, so it also takes more overhead in code structure. Let me know what you guys think.

tabbycat/api/urls.py
```diff
@@ -7,7 +7,19 @@
...
+    path('create/',
+        views.TournamentCreateView.as_view(),
+        name='api-tournament-create'),
+
     path('<slug:tournament_slug>/', include([
+        path('',
+            views.TournamentDetailView.as_view(),
+            name='api-tournament-detail'),
+
         path('<int:round_seq>/', include([
```
tienne-B: Maybe prefix with `round/`?

czlee: Maybe! The answer will probably be clearer once we start actually implementing round-specific APIs.

tienne-B:
I'm fine with prefixing tournaments by `tournaments/`, and maybe should do the same with rounds! I think I prefer the conventional naming now that I've thought about it as much cleaner than what I was going to do, which was to have the tournament detail at root, but also have `tournaments/` as a list/create view.

tienne-B:
I had no idea what was HATEOAS or hypermedia APIs, however having these hyperlinks to actions that can be taken seems like a great idea!

For the creation of tournaments, what'd you think about it being similar to the creation wizard in the human interface with automatically generated rounds?

tabbycat/api/serializers.py
```diff
@@ -2,20 +2,74 @@
...
+        view_name='api-breakcategory-list',
+        lookup_field='slug', lookup_url_kwarg='tournament_slug')
+    speaker_categories = serializers.HyperlinkedIdentityField(
+        view_name='api-speakercategory-list',
+        lookup_field='slug', lookup_url_kwarg='tournament_slug')
+
+
+class TournamentSerializer(serializers.ModelSerializer):
+
+    urls = TournamentEndpointsSerializer(source='*', read_only=True)
```
tienne-B: From what I see, we should prefix this kind of field with an underscore to show this field is not in the data structure but "meta" or for the API use.

czlee:
Hmm, for the API I'd advocate automatically generating absolutely nothing, not even own-institution conflicts. I think that it's simplest for everyone if API users are asked to be explicit about everything they want. Since they have a script (or are writing a third-party client), this shouldn't add much work. But we should be careful to document any gotchas that this may raise (_i.e._, cases where things are automatically created in the UI).

As for using conventional naming, I think my instinct is the same as yours (to use `/tournaments/SLUG/rounds/SEQ/`), but it does raise more questions. Almost all useful queries for us are filtered by tournament, and at the moment we do this similarly to the UI, using the prefix `/api/v1/tournaments/SLUG/break-categories` (or `/api/v1/SLUG/break-categories`). But it seems like REST naming conventions recommend using the query portion of the URL, like, `/api/v1/break-categories?tournament=SLUG`. I'm personally fine with breaking this convention, to recognize the centrality of tournaments in our data structure; in a sense, we think semantically of tournaments like a "hierarchy", so it makes sense to make this relationship hierarchical rather than filtered. But I thought I should raise the question explicitly. (By analogy: https://developer.spotify.com/console/playlists/)

jacklishufan:
I think it makes perfect sense to use `/api/v1/SLUG/` given the significance of tournament. Just one thing to note, institutions, unlike all other models, are not strictly under the 'hierarchy' of tournament. If we try to filter everything by tournament, it might be difficult to access the full information of an institution.

To be specific, the teams in institution view are filtered by the 'overarching' tournament, so it is impossible to get all the teams from different tournaments under the same institution. To work around this, we can simply choose not to filter at all, or create a separate `/api/v1/institutions` under root. In this particular case, I see  REST naming conventions more reasonable.

tabbycat/api/views.py
```diff
@@ -26,6 +29,38 @@ class AdministratorAPIMixin:
     permission_classes = [IsAdminUser]
 
 
+class APIRootView(AdministratorAPIMixin, GenericAPIView):
```
tienne-B: To go with the `/tournaments/` prefix, the API root should just be similar to:
```json
{
  "_urls": {
    "tournaments": "/tournaments"
  }
}
```
As just "directions" to the tournaments (or institutions with that side-discussion).

tabbycat/api/views.py
```diff
@@ -26,6 +29,38 @@ class AdministratorAPIMixin:
...
+        queryset = self.filter_queryset(self.get_queryset())
+        serializer = self.get_serializer(queryset, many=True)
+        tournaments_create_url = reverse('api-tournament-create', request=request, format=format)
+        return Response({
+            "tournaments": serializer.data,
+            "urls": {"create_tournament": tournaments_create_url}
+        })
+
+
+class TournamentCreateView(CreateAPIView):
```
tienne-B: The tournament views could be a ViewSet with the one `TournamentSerializer`.

tabbycat/api/urls.py
tienne-B: And with the current discussion with the `/tournament/` prefix.

tabbycat/api/fields.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def get_url(self, obj, view_name, request, format):
+        lookup_value = getattr(obj, self.lookup_field)
+        kwargs = {
+            'tournament_slug': obj.tournament.slug,
+            self.lookup_url_kwarg: lookup_value,
+        }
+        return reverse(view_name, kwargs=kwargs, request=request, format=format)
+
+    def get_object(self, view_name, view_args, view_kwargs):
```
tienne-B: What would you think of moving the tournament filter to the `get_queryset` method like with the `SingleObjectFromTournamentMixin`? I think it would be nicer.

tabbycat/api/serializers.py
tienne-B: I also suggest adding a `url` field to `api-tournament-detail` with the slug

czlee:
I'm going to change it to the `/tournaments/SLUG/rounds/SEQ` convention, we'll see how that tidies things up 🙂

@tienne-B Where did you see the convention about underscore prefixes? I haven't seen anything like it. But I have thought about using something like `links` rather than `urls` to avoid confusion between `url` and `urls`.

tienne-B:
Sounds good.

I think the underscore prefix (`_links` comes from HAL, but I don't think we're following a specific format. I just like it as it specifies that the field is not part of the data but rather extra information for the API.

czlee:
Cool, I'm open to the underscore prefix convention!

tabbycat/api/fields.py
czlee: Hmm, it'd be nice, but then how would you access the relevant `Tournament` instance?

tienne-B: I see

tabbycat/api/views.py
czlee: I agree.

czlee:
There were merge conflicts with #1381. I intended for this PR to precede other work on the API, providing a basis for other work, but I'm just going to merge it now since we seem to be progressing a lot on develop and future merge conflicts seem inevitable.

czlee:
Leaving the branch open for further work!

PR #1353, opened 2019-12-30
Use pre-commit for flake8 hook

czlee:
I noticed that flake8 now recommends using [pre-commit](https://pre-commit.com/) for git commit hooks:
https://flake8.pycqa.org/en/3.7.9/user/using-hooks.html

Since there's a config file, I thought I'd check whether it would make sense to commit this to the repo so that everyone's using the same one. The first commit just adds the config file; as I understand, a developer without pre-commit won't be affected by this. The second commit changes the pip requirements and CI script to use pre-commit instead of flake8.

To try this out, after checking out the branch:
```
# should install pre-commit
pip install -r config/requirements_development.txt

# will skip all tests because "no files to check"
pre-commit run
```

To get it to run as a git commit hook:
```
pre-commit install
```

If you run pre-commit on our existing code base, using
```
pre-commit run --all-files
```
you'll find that it fails and modifies a bunch of files, mostly for line endings. I've added the files that I don't think are appropriate to `exclude` in the .pre-commit-config.yaml configuration. The ones that remain are the ones that I'd be happy to have fixed in a commit, if we decide to use pre-commit.

Let me know what you think!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,23 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+exclude: locale/|^tabbycat/static/
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.4.0
+  hooks:
+  - id: trailing-whitespace
+  - id: end-of-file-fixer
+    exclude_types:
+    - csv
+    - svg
+  - id: check-yaml
+  - id: check-added-large-files
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.7.9
+  hooks:
+  - id: flake8
+    additional_dependencies:
+    - flake8-import-order
+    - flake8-quotes
+    - pep8-naming
+    exclude: migrations/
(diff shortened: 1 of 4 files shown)


czlee:
(The Travis CI build fails because our current code base isn't newline-at-end-of-file compliant 🙂)

tienne-B:
The errors would be nice to fix, whether or not we decide to use a pre-commit hook. I seem to remember we used such a hook previously, but got rid of it. I found the hook annoying.

philipbelesky:
I use a manually-added precommit hook, so happy to switch to this one. While the consistency may seem somewhat pedantic, it is useful as even with minor issues (like file-ends) they do tend to pollute the commit history (e.g. my editor automatically fixes these on save) so best to have consistency guaranteed at the point of committing.

czlee:
Developers still have to install the pre-commit hooks themselves, it doesn't come automatically with the repo (with this system or the old one). Sometimes I have it misconfigured and disable it because I'm too lazy to fix my config, then I forget to reenable it, then I have to do a flake8 fix commit. This seems suboptimal. Hopefully pre-commit'll help.

Will merge and also do a "fix everything" commit once @tienne-B approves, or feel free to continue discussion 🙂

Also @philipbelesky if you wanted to modify the pre-commit config in this PR to add ideas from your config, feel free to!

czlee:
Thanks guys!

PR #1354, opened 2019-12-30
README: Not limited to 3v3/BP formats

tienne-B:
Generalize to all parliamentary formats, but explicitly include British Parliamentary as it is quite distinct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 </div>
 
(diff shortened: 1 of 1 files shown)


philipbelesky:
I don't think that the use of 'parliamentary' here is common knowledge, or is at least less intuitive than saying "3vs3" or "2 team". For instance, I have no idea if Australs, much less many of the more obscure 3vs3 formats from NZ would count as 'parliamentary".

tienne-B:
It is however the terminology we use in the project's description.

philipbelesky:
I'd prefer to update the description text in that case, perhaps "Debating tabulation software for British Parliamentary and a variety of two-team formats"?

tienne-B:
I'd be good with that, provided it's also reflected in the docs and README.

PR #1355, opened 2019-12-31
Codeclimate: Update to version 2

tienne-B:
This commit removes the deprecated 'ratings' and renames fields where needed.

It also marks i18n files as excluded.


tienne-B:
Okay, this patch is ready now. Also, seems that Codeclimate uses Python 2, which gives syntax errors with non-ASCII symbols (like alpha or chi)

PR #1359, opened 2019-12-31
Delete tournament history page

tienne-B:
The page is a couple years out of date and with dead links abounding. An example of a Tabbycat instance is given in the README (which is more recent than the tournaments listed on the page).


philipbelesky:
I would be happy to remove this particular page from the docs, but I would still like it kept within the repository; maybe just as a `.md` file in docs. The intent of the file was not to show of instances of the site, but to track the history of how/where Tabbycat was used. 

Anyway, let me know if you'd like me to add the file as a new one (after merging this) or move the file separate to this PR.

tienne-B:
Move the file separately from the PR; it'd be odd to have a deletion/recreation and would mess with history.

PR #1360, opened 2019-12-31
Add API endpoints for tournament preferences

tienne-B:
These endpoints are provided by django_dynamic_preferences and allows for retrieval and bulk editing of the preferences.

### DIFF
--- a/tabbycat/api/urls.py
+++ b/tabbycat/api/urls.py
@@ -1,6 +1,12 @@
+from django.conf.urls import url
 from django.urls import include, path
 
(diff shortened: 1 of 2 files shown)


czlee:
Wow, the default list view is really long, we should really use a less comprehensive serializer for it, and possibly pagination too.

tienne-B:
Don't think the length of the response should be of concern. However, you can add the URL query `&section=` to filter by section. We shouldn't modify what dynamic-preferences provides more than needed.

PR #1374, opened 2020-01-01
Add prefetches to eligibility API views

tienne-B:
Not that important of prefetches as it wouldn't reduce the number of queries much, but still good to use prefetches where possible.


tienne-B:
This patch is trivial. When #1351 is merged, I'd reuse this branch to change the teams/speakers lists into URI lists, if this patch isn't merged itself by then.

PR #1375, opened 2020-01-01
Use bulk query methods with draw creation

tienne-B:
This commit replaces the one-by-one creation of `Debate` and `DebateTeam` objects with `.bulk_create`s by splitting the `_make_debates` (from pairings) loop in two, associating each pairing with a debate and iterating over that dict in the 2nd loop.

Afterwards, venue assignation is optimized by using `.bulk_update`.

Work on #676. 
Supersedes 102fc4afd84e6568279f7888077a8a5de6ba25bd.


tienne-B:
I'll just change the venue assignation on top of 102fc4a and make a PR on that.

PR #1376, opened 2020-01-01
Use bulk query methods with draw creation

tienne-B:
This commit replaces the one-by-one creation of `Debate` and `DebateTeam` objects with `.bulk_create`s by splitting the `_make_debates` (from pairings) loop in two, associating each pairing with a debate and iterating over that dict in the 2nd loop.

Afterwards, venue assignation is optimized by using `.bulk_update`.

Work on #676.
Building upon @czlee's old work.


czlee:
If you'll pardon the self-quoting again:

From https://github.com/TabbycatDebate/tabbycat/commit/7a67893115a19fdf385878d7726237c6df37e4ba#r29957930:
> Where possible, `bulk_create`s are (I'm sure you'd agree) the best way to reduce N+1 issues with `INSERT` statements. But I'm a little reluctant to make this a standard pattern of ours. The `bulk_create()` method also bypasses the `save()` function, which for some of our models does important things, and unless we have good systems for tracking what `bulk_create`s happen where, I worry that one day we'll add a `save()` method to a model that doesn't already have one, and break some `bulk_create()` in a file we forgot was there, but it'll be latent for ages and it'll show up as a really, really obscure bug and we'll have to spend ages tracking it down, if we succeed at all. Weighing up that against some slow transactions when a lot of things are being saved—I think I'm okay with transactions that involve inserts/updates (as opposed to selects) taking quite a long time, if it gives me more confidence that it's being done right.

From https://github.com/TabbycatDebate/tabbycat/issues/676#issuecomment-417996833:
>Despite this proof of concept, as explained [above](https://github.com/TabbycatDebate/tabbycat/issues/676#issuecomment-410428954), I'd prefer that for database writes (`INSERT` and `UPDATE`), that we don't bother trying to resolve the N+1 inefficiency, and instead just take the hit. I'm open to arguments to the contrary, but for now I think the additional risk of bugs (intrinsic to using `bulk_create()`) is too high to offset the time cost for functions that are used once per round.

Can you comment on your thoughts on this principle, @tienne-B, before we take this PR any further?

tienne-B:
I agree that we should be careful, seen the risks of bypassing signals and if `.save()` is implemented. I am however much less worried with using bulk methods here. The models created and modified here are very well established and less likely to evolve or to subclass `.save()` in future development of Tabbycat. Furthermore, the creation of these are limited to the method modified here, and bulk methods are already currently used for `Debate` in the DB area.

I find the performance increase really non-negligible; when comparing using `bp88team` with random and power-paired rounds, the processing time more than halved.

In all, I don't like having these stale branches; rather fix them up and discuss to get them behind us!

philipbelesky:
Just following up on this: I agree RE: the stale branch issue. I'm also happy to accept the risks @czlee raises as long as you're happy to take be on the hook for them @tienne-B. Also, as a side note, the performance increase might be worth noting in the `CHANGELOG`?

philipbelesky:
I was going to merge, but aren't able to handle the conflicts atm. Could you merge this in when you had a moment @tienne-B ?

tienne-B:
Sure; done. To refine a bit my argument in favour of bulk creations here at the expense of not being able to use things like `.save()` methods, I'd point to having `Pairing`s as an intermediate data structure on which debates are created. With the pairings structure, the logic and validation needed are handled outside the database model objects.

PR #1379, opened 2020-01-03
Correct outdated information in "Starting a Tournament"

tienne-B:
Divisions have been removed and the "Edit Database" link had been moved in the menu. This commit corrects these. Found this while translating!


philipbelesky:
Good eye, thanks!

PR #1382, opened 2020-01-05
Rename "Venue" to "Room" in UI

tienne-B:
This commit changes most instances of "venue" found in the user UI and shown messages to "room," including for categories and constraints.

The term "room" seems more appropriate than "venue" as the latter normally refers to a building, and not the specific rooms inside it, although large rooms such as auditoriums (like for break rounds) can be called venues in and of themselves. However, they are normally treated as rooms with numbers and such.

As a start, this commit only changes the interface. To go deeper with this change, we would need to rename models and fields, then the whole "venues" application.

Marking this as a draft as to keep the other parts of this refactor in this PR and for discussion on whether to go forwards.


philipbelesky:
Hey thanks for making these changes. I'm happy to go ahead with the change if @czlee is. I suspect there will be a bunch of strings in the docs that also need changing. Further, I'd prefer that source code file names match (e.g. `EditDebateVenuesContainer`), if just to help future contributors.

tienne-B:
@philipbelesky: Fixing uses of "venue" in the internal code (not user-facing) will be my next commit; it's just much more involved to make sure nothing gets broken. And yes, there will be docs to modify!

A consideration to keep is that this change in terminology is anglo-centric and not a change in semantics thus I wouldn't consider translations of these terms fuzzy as a result of these changes. We'll probably have to re-place the current translations manually on Crowdin after synchronizing these changes.

philipbelesky:
Sounds good!

czlee:
I'm good in principle with the terminology change. I don't feel strongly about doing source code and UI strings at the same time, but would back anyone who thinks doing them at the same time is preferable. I agree that if we can we should see if we can get Crowdin not to mark these strings as needing to be retranslated, at least not in those languages that have already agreed on using "room".

tienne-B:
Modifying uses of "venue" in the source code would be important for maintainability; not having two sets of vocabulary. This would be very difficult though as there is a "venues" application, and renaming entire apps is not easy.

philipbelesky:
This has been open for awhile so I would be fine to merge it in with the UI-only changes. Happy to defer the issue of changing source references to a later date, if at all. @czlee is there work needed on the Crowdin issue you mentioned, or is this good to go?

czlee:
> @czlee is there work needed on the Crowdin issue you mentioned, or is this good to go?

Umm, to be honest @tienne-B probably knows more about Crowdin than me, but I can't see any easy fixes. I'm tempted just to say it's easiest to do the annoying manual work of pulling in the translations from translation memory. Hopefully the translation memory will make this slightly less tedious.

tabbycat/utils/tables.py
```diff
@@ -726,9 +726,9 @@ def construct_venue_cell(venue):
 
             if len(descriptions) > 0:
                 if len(descriptions) == 1:
-                    categories_sentence = _("This venue %(predicate)s.") % {'predicate': descriptions[0]}
```
czlee: Now that I'm looking at this again, I guess this data structure makes no sense in a translated string, since the predicate doesn't also change languages, lol.

tienne-B:
The structure of Crowdin translations doesn't permit "statuses" of translations other than not having a translation, having one, or it being approved. Another status, which is supported and used by gettext is "fuzzy" which is when a translation's source has been updated, and so the translation should be checked for if it's still applicable.

While I set Crowdin to merely unapprove these "fuzzy" strings, it doesn't mark them to be re-checked; we should just notify the translators of the change and ask if they'd change their strings based on our change.

tabbycat/utils/tables.py
tienne-B: What do you mean, "data structure"? By the predicate not changing languages, the grammar of other languages might place them elsewhere.

czlee: I mean the fact that we ask users to provide a predicate, e.g. "has step-free access", which would show up as "This venue has step-free access" in English. But "Esta sala has step-free access" doesn't make very much sense 😂

We should probably just abolish the "description" field and rely on category names only.

czlee: (This is not related to this PR, I just noticed it while reviewing the code changes.)

tienne-B:
Once this is merged, I'll update all the strings in Crowdin (not only those changed here)

philipbelesky:
All merged!

PR #1384, opened 2020-01-13
Redirect to tournament page if only one active

tienne-B:
This is a change from redirecting if there is only one tournament, active or not. However, the whole list of tournaments would be listed if not redirected.


czlee:
Ah, this makes sense, thanks! Questions:
1. Do we do sensible things everywhere else? Like, should we separate inactive tournaments from active tournaments on the home page, or hide them?
2. If there's only one active tournament, is there a way to force a visit to the home page to see the inactive tournaments (_i.e._, a way to get to "/?redirect=false" without typing that into the URL box)?

tienne-B:
1. I would be hesitant to filter the tournament list, as I think of it as a list of all the tournaments, including inactive ones, as people may still want to access them. With the filtering, there would be no page to list the inactive tournaments, nor for admins to access them (as the sidebar is already filtered)
2. No. I thought adding the query string would be sufficient, and I am hesitant to add a new page as I'm unsure about the discoverability of such a page.

czlee:
By (1) I meant separate, not filter. So like, present a list of active tournaments, then a list of inactive ones, rather than a single list with them mixed together.

By (2) I meant is there a link to the page with the query string somewhere? Normally you don't design websites such that the _only_ normal way to visit a (non-error) page is by typing in the URL.

czlee:
Oh, actually, this only affects the public view, right? So inactive tournaments are hidden from public view always (and there's intentionally no way to access them), and admins always see the full home page anyway? Do we show inactive tournaments on the public (non-admin) home page?

tienne-B:
Only affects the public view, but gives the full tournament list if there is not a sole active tournament.

philipbelesky:
I'm not sure what 'active' means either, but given it exists as an option it makes sense for it to have the behaviour developed here (rather than doing nothing). Thanks for adding this in!

PR #1385, opened 2020-01-14
Broadcast when ballots are confirmed

tienne-B:
When ballots get confirmed, they were not being broadcasted to the consumer, and thus didn't update in the graph as identity was checked, not the string. Also placed it in the previous (identical) if block and set the comparison with the model (static) attribute and not per-object.


philipbelesky:
Thanks for looking into this. I've had a play and it seems to work, but I never noticed any problems with the old code either. Was the problem that it assumed the debate status always == the ballot status?

tienne-B:
The first commit is somewhat useless as it still worked and the operator does match. Just think it's best to use the appropriate operator rather than another (with the same behaviour in our case)

PR #1386, opened 2020-01-14
Remove public ballot submission redirect view

tienne-B:
The ballot submission views now redirect to better pages, like the tournament homepage or private URL page, where appropriate. The old page is removed, as it was a relic of when there were no suitable pages.

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -470,9 +470,6 @@ class BasePublicNewBallotSetView(PersonalizablePublicTournamentPageMixin, BaseBa
         messages.success(self.request, _("Thanks, %(user)s! Your ballot for %(debate)s has "
                 "been recorded.") % {'user': self.object.name, 'debate': self.matchup_description()})
 
-    def get_success_url(self):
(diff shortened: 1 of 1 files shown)


philipbelesky:
This comment in `PostPublicBallotSetSubmissionURLView` explains why this page was added. We've had ongoing problems where the messages that are provided to users — e.g. "You submitted a ballot for X vs Y" getting mixed up such that people would get notifications for debates they were not present in. We couldn't find the root cause for this, but assumed at it was due to the caching layer; hence this page.

tienne-B:
Remember that three years ago, there was no private url landing page and that is why there were errors. Now, there would be no error as a redirect to the landing page would serve the same purpose and be more useful.

philipbelesky:
It still goes to the public index when accessing `OldPublicNewBallotSetByIdUrlView` though right? That said, people using a public/open form probably don't have the scale to worry about that particular issue which seems to crop up when there are many simultaneous submissions.

tienne-B:
Good point. Reduced the scope of the patch to just private URLs.

Issue #1352, opened 2019-12-30
Privateurls: Indicate actions to be taken by participant


tienne-B:
There are three volets to this task, for a goal to further streamline participants' interactions with Tabbycat through their private URLs. This should reduce the number of duplicate feedback.

* [ ] Visually highlight the _Submit Ballot_ button for adjudicators if their assignment in the current round has no ballot. This is not as important as the other points as the normal flow for ballots is to open the ballot at the beginning of the debate, but I've seen others open at the end.
* [ ] Highlight the _Submit Feedback_ button after the ballot is entered (for judges) and the participants when there is expected unsubmitted feedback. The button could also show how many are to be submitted.
* [ ] Indicate to whom feedback needs to be submitted. This could deter duplicate submissions and submissions to wrong people (I suppose that could happen). The problem to consider is we want to keep the drop-down list items short as they are already cut off (see efec9c3cd7fb78d4b52a942eb695f1c85aec89bf) on mobile.

The idea behind these points are to streamline as much as possible and indicate as much as possible (in as many ways as appropriate) of what is expected from participants.

philipbelesky:
These all sound good, although I'd be wary of cluttering that page up with alerts given it is primarily accessed on mobile devices. Look forward to seeing the prototype!

Issue #812, opened 2018-09-09
Rethink UX of feedback form to prevent unnecessary submissions


philipbelesky:
The problem of people submitting feedback on panellist rather than the adjudication-deliverer seems to be relatively endemic. I added a warning to the form in fa70381d930b9628ec8bb62c2a85ad3faeb7e89e but I suspect this is not going to be enough.

I suspect a more robust method would be to structure the questions in the form to properly limit choices based on what happened in the debate. My first thoughts for that structure would be:

1. Which round is this for? 
    - *This then filters (3) based on the answer*
2. Did the chair deliver the adjudication?
    - *This then filters (3) to either be a disabled/single-option `select` or a multi-option `select`*
3. Normal adjudicator selection drop down

The intermediary step of (2) would only show if the relevant `feedback_from_teams` pref matches. It is probably also worth adding a `help_text` or similar that reiterates the expectation of the setting (i.e. a slightly more subtle of the alert added in fa70381d930b9628ec8bb62c2a85ad3faeb7e89e).

If it makes the implementation easier I wouldn't mind if these three questions were actually presented prior to the standard form being shown; i.e. as a two-part formset or as separate forms. Having them as separate pages might also further reinforce that it is a task that doesn't need to be repeated for every available option.

Finally, in the 'nice to have' bucket, I wonder if its worth having some way for teams/adjs to see what feedback they have already submitted. This could help the case where different members of the team accidentally submit duplicate feedback not knowing someone else had done so. We could potentially roll the team record links into the private URLs for this purpose or do something more limited.

czlee:
For what it's worth, I think this is enough work to qualify as "outside the scope of Tabbycat" and "should be delegated to a third-party system". Alternatively, if we're giving up on the idea that someone else might develop a meaningful adjudicator feedback system, our overhaul wishlist also includes other things of possibly higher priority. Most notably, we often get requests to differentiate between feedback on chairs and feedback on panellists, which I've been resisting because it would entail a structural rework of the feedback forms.

I don't think anything's going to be really enough to satisfy us in terms of user experience and guidance, short of a structural rethink of how feedback submission works. Although we've added lots of bells and whistles, the current system's architecture is still mostly based on the Australs 2010 system, when data entry was limited to tab assistants and we didn't need to foolproof the system from debaters. The only appreciable structural change since then was in 2014, when we added the submission version numbering system to support public submissions for the first time.

philipbelesky:
[...]

PR #1387, opened 2020-01-15
Group adjudicators in feedback by round

tienne-B:
This commit further reduces the width of the adjudicator selection in feedback forms by placing the round name as a grouping rather than inline.

Thinking of adding further information in whether feedback has already been provided, to avoid duplicates. This will come in a later commit.

Relates to #1352.3
Relates to #812


philipbelesky:
This looks very nice, I hadn't considered that subgroups like this would be a great way to organise the list.

PR #1389, opened 2020-01-31
Use intcomma with number of DB rows

tienne-B:


czlee:
Now that I'm looking at this, we should really `intcomma` that 10,000 too.

czlee:
…unless translators can be relied on to get that comma-ization right? I guess they can?

I just realized it only looks weird when that string hasn't been translated yet… and then the first number has localized commas, while the second number (10,000) has the English comma.

PR #1404, opened 2020-02-17
Re-structure CONTRIBUTING file

tienne-B:
This commit re-structures the CONTRIBUTING file into sections for the activity and subsections for parts of development and maintenance tasks.

It also changes the introduction to list some direct actions people may take before deferring with suggestions/feedback. A new section on translations has also been added.

We might want to expand the translation section with more guidance (perhaps indicate the amount of translations to do and it being best to do in groups)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -2,7 +2,7 @@
 Contributing
 ============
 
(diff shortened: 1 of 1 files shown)


philipbelesky:
This looks great, thanks for the update.

PR #1406, opened 2020-02-22
Localize numbers in ballot receipt emails

### LINKED ISSUE #1405: Reporting of total speaks in emails uses integers rather than floats
This was a bug report to the FB page, just reposting here for better tracking and if anyone else wants to look at it. Haven't tried to reproduce/look at the code but the report seems clear

In the (default?) email template that is sent to adjs the line *Proposition: Team A (Loss - X total speaks)* reports *X total speaks* as an integer which can lead to incorrect totals in formats with half points.

tienne-B:
As a fix to the speaker points being trimmed to integers, should also respect the number formatting by locale.

Fixes #1405.

### DIFF
--- a/tabbycat/notifications/utils.py
+++ b/tabbycat/notifications/utils.py
@@ -11,6 +11,7 @@ Objects should be fetched from the database here as it is an asyncronous process
 thus the object itself cannot be passed.
 """
 
+from django.utils import formats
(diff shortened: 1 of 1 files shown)


philipbelesky:
Hey, thanks for picking this up. I'll have a quick look at the other hotfix Issue over the next few days and roll out a release.

PR #1407, opened 2020-02-24
Ensure ballot graphs do not include decimals

tienne-B:
For smaller tournaments and elimination rounds, the ballots graph may include ticks for non-whole numbers, which doesn't make sense in this situation.

Other things we might consider for the graph is a legend for the colours or labels for the axes.


philipbelesky:
Good catch, thanks!

PR #1410, opened 2020-02-29
Simplify database row count query

tienne-B:
The query to find the number of rows in the database (warn against limited database plans) had extra information and handling, which is not needed for the task. This commit changes the query to directly take the sum of the live tuples, without fetching the names or ordering.

The summation is no longer handled by Python either.


czlee:
That's awesome, thanks!

tienne-B:
Ran an `EXPLAIN ANALYZE` on both queries as I was interested. Apparently the new query is slightly longer, just due to the aggregation, but it does save in memory and Python computation! :)

PR #1420, opened 2020-03-08
Re-implement debate postponement

### LINKED ISSUE #534: Allow next draw to be generated with unconfirmed (but existent) ballots
Currently, we prevent draws from being generated before all ballots are confirmed. In some contexts, particularly where pairings within brackets are random (WUDC), it makes sense to press ahead before all ballots are confirmed, hoping or assuming that the results (not speaker scores) are correct. This needs to be done with various checks to prevent stupidity, though.

### LINKED ISSUE #1283: Add ability to flag submitted debates for later
It would be useful to have a quick checkbox for debates in the admin/assistant input views. Debates could be marked if some follow-up is required or potentially so, rather than keeping track separately. This would be best in the table itself, but would have to be connected to websockets so that it is reflected when selected/deselected.

tienne-B:
In brief, debate postponement is to signal that a ballot should get re-checked or is not vital to the creation of the next round, and so you can advance to the next round with postponed debates.

Debates can be postponed with or without a ballot; to postpone without results, each debate will have a button on the results page in order to postpone the debate. They can only be un-postponed by adding results, and in the assistant view, it is automatic. The button will work if there are results, as the status could be used to indicate that the ballot should be re-checked.

Postponing a debate activates the ballot status consumer, which updates the table.

A similar feature was previously available with divisions, but was recently removed.

Fixes #534, with this change as a necessary "guard."
Fixes #1283

Not quite ready; want to make the table buttons update with the consumer, similarly to the icons.

### DIFF
--- a/tabbycat/draw/models.py
+++ b/tabbycat/draw/models.py
@@ -20,7 +20,7 @@ class DebateManager(models.Manager):
 
 class Debate(models.Model):
     STATUS_NONE = 'N'
-    STATUS_POSTPONED = 'P' # obsolete
+    STATUS_POSTPONED = 'P'
     STATUS_DRAFT = 'D'
     STATUS_CONFIRMED = 'C'
     STATUS_CHOICES = (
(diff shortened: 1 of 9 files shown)


philipbelesky:
I wonder if this (along with many other 'league' features) are better kept behind a flag? Theres a complexity and chance of inadvertent actions that I'd rather keep hidden from standard use.

tienne-B:
Superseded by #1563

PR #1421, opened 2020-03-08
Move Backups doc page to Use section

tienne-B:
"Tabbycat doesn't provide an in-built backup system; instead you should create copies of your database directly." Thus backups isn't really a feature of Tabbycat, and is oddly categorized under "Features" in the documentation. Not denying the value of the page, the page should be kept.

This change moves the page to the installation menu, as I see these instructions being more in line with how to install the DB/data and the application itself. An alternative would be a "Backup" section in most/all of the installation method pages.


philipbelesky:
Wouldn't it be better under the 'using' section? The installation instructions are all one-offs and split by platform. Backing up should ideally be neither.

tienne-B:
Sure. That page is hard to categorize. I'm less sure of where it should be than that it shouldn't be counted as a "feature"

czlee:
Hmm. "Feature notes" is perhaps not a great name for the category; I think the idea of those pages is that they each cover a particular thing that you might use in Tabbycat. I'd back moving it to "use", except that to me the "Using Tabbycat" section is more of a step-by-step guide that summarizes what you do through a tournament, rather than more detailed information about a specific feature.

While I agree that backups aren't a "feature" of Tabbycat, if we ignore the section names for a moment, to me backups belong among that group of pages, because they each describe a single "thing" (for want of a better term). In any case, the name "feature notes" sort of makes it sound like they're just notes describing a feature, when it's actually meant to be a guide for using them. Is there a different heading we could use for what is currently called "feature notes", to reflect that they're about particular parts of Tabbycat and other tools one might use while using Tabbycat?

philipbelesky:
We seem to broadly agree it makes sense in 'using' so ill merge this. I think it makes sense in the context of a step-by-step guide, even if may not be something that happens every round.

Maybe rather than 'feature notes' it could be 'feature guides'?

czlee:
Oh, sorry, I meant that we're not necessarily describing features, per se. "Feature guides" is an improvement, but the point is that these are optional things in the software that users can choose to use, and these guides describe how to use them. So "backups" should fall in this category too—the fact that it is not a feature of Tabbycat isn't relevant, the fact that it is a thing that you can optionally do is.

PR #1423, opened 2020-03-10
Add selectors for adjudicator positions in draw emails

tienne-B:
This commit adds three selectors in the Email Adjudicators' Draw view, for chairs/panellists/trainees, which can be used to give different instructions to these classes. To do so, type "type" in the checkboxes has been generalized.

In addition, the `get_success_url()` has been simplified to just the pattern with the `get_redirect_url()` of the Tournament/Round mixins doing the rest.

### DIFF
--- a/tabbycat/draw/views.py
+++ b/tabbycat/draw/views.py
@@ -5,6 +5,7 @@ from itertools import product
 
 from django.conf import settings
 from django.contrib import messages
+from django.db.models import OuterRef, Subquery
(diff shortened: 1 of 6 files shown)


tienne-B:
BTW, the use of `OuterRef` and `Subquery` was needed to get the adj positions to keep the query as a queryset of people.

czlee:
The new trailing commas check has struck 😮

You might need to `pre-commit install` to update the checks (though, I don't remember having to do this).

PR #1425, opened 2020-03-16
Add optional config variables for email

tienne-B:
For instances that have problems with SendGrid, this commit will let people setup other providers/bypass the SendGrid add-on.

### DIFF
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -99,10 +99,18 @@ CHANNEL_LAYERS = {
 }
 
 # ==============================================================================
-# SendGrid
+# Email / SendGrid
 # ==============================================================================
 
-if environ.get('SENDGRID_USERNAME', ''):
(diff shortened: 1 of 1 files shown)


philipbelesky:
Great, thanks for adding this. I don't have time to test at present, but just a note that we should make sure to check the different permutations (if you haven't already) during the next release candidate.

tienne-B:
I was thinking about that, and I'd agree to use the conditional. However, SendGrid should be the `elif` condition here, as custom configuration should have priority over SendGrid configuration even if SendGrid variables are still present.

czlee:
Sure. Either way, was just thinking it'd make more sense to treat them as mutually exclusive cases rather than a per-variable fallback.

tienne-B:
To reopen this; could be useful to not limit admins with one service (SendGrid) over another.

czlee:
Merging. Do we want to add docs? 🙂

PR #1426, opened 2020-03-17
Corrections to strings

tienne-B:
See commit messages.

### DIFF
--- a/docs/features/breaks.rst
+++ b/docs/features/breaks.rst
@@ -36,7 +36,7 @@ which is the default.
   :widths: 30 70
 
   * - Rule name (string to use in ``importtournament`` CSV files)
-    - Descrption
+    - Description
 
   * - Standard (``standard``)
     - The top *n* teams break. This is the default, and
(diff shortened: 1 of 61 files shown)


philipbelesky:
My first though that more explicit var names, i.e. `round_name` over 'name' would be preferred, but happy to defer here given you are much more involved in translation.

tienne-B:
My point was to have explicit yet consistent variable names, so remove the `_name` suffix to keep `round` from `round_name`, not the opposite! Different places used those three variations (and others...)

czlee:
The reason the variable name `round` wasn't used was to avoid conflict with the variable `round` that the context processor sets:

https://github.com/TabbycatDebate/tabbycat/blob/74622b05a150cdf4a1441e0212e2824840033cd9/tabbycat/utils/context_processors.py#L17-L26

While it might not have practical effect on some _particular_ page to overwrite this context variable, I'd still prefer to avoid variable naming conflicts in templates as a matter of good practice.

tienne-B:
From my tests and the documentation, the (re-)definition of template variables have the scope of the block tag, and becomes « normal » again afterwards. Am I wrong in that?

czlee:
I actually think it's bad practice to have two variables with the same name, even if the scope of one is more local than the other, because it can be confusing. So I prefer to avoid having to ask the question you just did in the first place. (It's fine if they're in two disjoint scopes.)

PR #1438, opened 2020-04-29
Sort debates secondarily by bracket

### LINKED ISSUE #1437: Adjudicator Allocation Page
Currently when sorting the Adjudicator Allocation page by importance it secondarily sorts by bracket in ascending order. My suggestion is that it would be preferable for the secondary sort to be by bracket in descending order. This better conforms to the default sorting which is bracket in descending order.

tienne-B:
This commit adds a second criteria when sorting debates by importance in the adjudicator allocation view. Instead of stopping after importance, the brackets are used to sort `DESC` instead of the default `ASC` as the previous order of the page.

Fixes #1437

### DIFF
--- a/tabbycat/templates/allocations/DragAndDropStore.js
+++ b/tabbycat/templates/allocations/DragAndDropStore.js
@@ -177,14 +177,17 @@ export default new Vuex.Store({
     },
     sortedDebatesOrPanels: (state, getters) => {
       let debatesOrPanels = getters.shardedDebatesOrPanels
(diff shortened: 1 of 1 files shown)


philipbelesky:
Hey, thanks for adding this. I think I had assumed that the previous sort would become the secondary sort option (as the list would preserve some parts of the order) - but it looks like it actually was doing the reverse (sorting by bracket descending). This makes sense as a behaviour; at least until proper secondary sorting is added

tienne-B:
Ascending and descending are confusing terms, as ascending values get bigger going down a list. We want it descending :)

PR #1455, opened 2020-05-13
Set number of preformed panels for outrounds

### LINKED ISSUE #1452: Preformed panels supporting outrounds
I have been having a play with preformed panels and from my quick attempt to generate them for outrounds, it seems to generate preformed panels as if it was generating panels for an additional preliminary round rather than a break round. 

For example this is the preformed panels that generated when I generated preformed panels for quarter finals for one of our tournaments.

image

We did end up changing some thing to do with the round sequence for these rounds (we added 2 additional in rounds, deleted the octo finals and edited the sequence numbers, but this round is set up as per the settings below:

image

tienne-B:
This commit creates a new cached property for rounds which calculates the number of debates a round would have without any prior knowledge.

With this property, the anticipated information used for preformed panels can defer to this property if the info is not available.

Fixes #1452

### DIFF
--- a/tabbycat/adjallocation/preformed/anticipated.py
+++ b/tabbycat/adjallocation/preformed/anticipated.py
@@ -3,6 +3,7 @@
 import itertools
 
 from breakqual.utils import calculate_live_thresholds, determine_liveness
+from draw.generator.utils import ispow2, partial_break_round_split
 from participants.prefetch import populate_win_counts
 
 
@@ -26,13 +27,35 @@ def calculate_anticipated_draw(round):
 
     nteamsindebate = 4 if round.tournament.pref('teams_in_debate') == 'bp' else 2
 
-    if round.prev is None or not round.prev.debate_set.exists():
-        # Special case: If this is the first round, everyone will be on zero.
+    if round.prev is None or not round.prev.debate_set.exists() or round.is_break_round:
+        # Special cases: If this is the first round, everyone will be on zero.
         # Just take all teams, rounded down -- if this is done, it'll typically
         # be done before availability is locked down. Also do this if the last
         # round hasn't yet been drawn, since that's premature for bracket
         # predictions.
-        npanels = round.tournament.team_set.count() // nteamsindebate
+        #
+        # Also occurs for elimination rounds as everyone is just as live.
+
+        nteams = 0
+        if round.is_break_round:
+            break_size = round.break_category.break_size
+            nprev_rounds = round.break_category.round_set.filter(seq__lt=round.seq).count()
+            partial_two = nteamsindebate == 2 and not ispow2(break_size)
+            partial_bp = nteamsindebate == 4 and ispow2(break_size // 6)
+            if nprev_rounds > 0 and (partial_two or partial_bp):
+                # If using partial elimination rounds, the second round is the first for
+                # the powers of two, so start counting from here.
+                nprev_rounds -= 1
+
+            if nprev_rounds == 0 and nteamsindebate == 2:
+                nteams = partial_break_round_split(break_size)[0] * 2
+            else:
+                # Subsequent rounds are half the previous, but always a power of 2
+                nteams = 1 << (break_size.bit_length() - 1 - nprev_rounds)
+        else:
+            nteams = round.tournament.team_set.count()
+
+        npanels = nteams // nteamsindebate
         return [(0, 0, 0) for i in range(npanels)]
 
     # 1. Take the (actual) draw of the last round, with team points
--- a/tabbycat/draw/generator/utils.py
+++ b/tabbycat/draw/generator/utils.py
@@ -11,8 +11,8 @@ def nextpow2(n):
 
 
 def partial_break_round_split(break_size):
-    """Returns a tuple `(debating, bypassing)`, where `debating` is how many
-    teams will debate in the first break round, and `bypassing` is how many
+    """Returns a tuple `(debates, bypassing)`, where `debating` is how many
+    debates there is in the first break round, and `bypassing` is how many
     teams will bypass the first break round, qualifying directly for the
     second."""
 


philipbelesky:
Hey, this looks good! I'll leave it to CZ to review as he wrote the original function.

tienne-B:
Could be in as a hot fix, as not showing any slides for preformed panels in elimination rounds is not expected.

tabbycat/tournaments/models.py
```diff
@@ -597,3 +597,21 @@ def motions_good_for_public(self):
...
+    @cached_property
+    def num_expected_debates(self):
+        """Returns the number of debates expected for a round with perfect team availability,
+        without checking the size of previous rounds (for elimination)."""
+        nteams = self.tournament.team_set.count()
+        if self.is_break_round:
+            rounds = list(self.tournament.round_set.filter(break_category=self.break_category).values_list('seq', flat=True).order_by('seq'))
+            if rounds.index(self.seq) == 0:
+                # First elimination round - includes partial
+                nteams = self.break_category.break_size
```
czlee, line 610: Partial elimination rounds don't include all the teams, right? So for example if the break size is 17, then there's just one debate in the partial elimination round? `partial_break_round_split` in draw/generator/utils.py does this calculation for two-team formats.

Our support for two-team and BP formats here actually differs. For two-team formats, we support any integer. For BP, however, we only support break sizes of `4*2^n` or `6*2^n` for some integer `n`. (I can't remember whether this was because I couldn't figure out how to make other partial break rounds coherent, or whether I was just trying to get it done.) See draw/generator/bpelimination.py and also: https://github.com/TabbycatDebate/tabbycat/blob/7a1b4570b0698243cbe734d14f0663cd1660c4a7/tabbycat/draw/manager.py#L250-L270

tabbycat/tournaments/models.py
```diff
@@ -597,3 +597,21 @@ def motions_good_for_public(self):
...
     def ballots_per_debate(self):
         return self.tournament.ballots_per_debate(self.stage)
+
+    @cached_property
+    def num_expected_debates(self):
+        """Returns the number of debates expected for a round with perfect team availability,
+        without checking the size of previous rounds (for elimination)."""
+        nteams = self.tournament.team_set.count()
+        if self.is_break_round:
+            rounds = list(self.tournament.round_set.filter(break_category=self.break_category).values_list('seq', flat=True).order_by('seq'))
```
czlee, line 607: If the purpose of this query is to determine how many break rounds in the same break category precede it (I might've missed something, but it looks to me like both references are `rounds.index(self.seq)`), I think you can do this with a single query that returns the desired number 😉

tienne-B:
Oh, the anticipated file is better, just thought it was maybe too specific to preformed panels and this could be used elsewhere, but I wouldn't think of anywhere else.

tabbycat/draw/generator/utils.py
```diff
@@ -11,8 +11,8 @@ def nextpow2(n):
 
 
 def partial_break_round_split(break_size):
-    """Returns a tuple `(debating, bypassing)`, where `debating` is how many
-    teams will debate in the first break round, and `bypassing` is how many
+    """Returns a tuple `(debates, bypassing)`, where `debating` is how many
+    debates there is in the first break round, and `bypassing` is how many
```
czlee, line 15: oooh nice catch thanks

tabbycat/adjallocation/preformed/anticipated.py
```diff
@@ -26,13 +27,34 @@ def calculate_anticipated_draw(round):
...
+            partial_bp = nteamsindebate == 4 and ispow2(break_size // 6)
+            if nprev_rounds == 0 and nteamsindebate == 2:
+                return partial_break_round_split(break_size)[0]
+            elif nprev_rounds > 0 and (partial_two or partial_bp):
+                # If using partial elimination rounds, the second round is the first for
+                # the powers of two, so start counting from here.
+                nprev_rounds -= 1
+
+            # Subsequent rounds are half the previous, but always a power of 2
+            nteams = 1 << (break_size.bit_length() - 1 - nprev_rounds)
```
czlee, line 53: Would this be easier to read if you just treated (1) two-team break rounds, (2) BP break rounds and (3) non-existent previous rounds as wholly distinct cases, rather than this sort of nested logic?

tabbycat/adjallocation/preformed/anticipated.py
```diff
@@ -26,13 +27,34 @@ def calculate_anticipated_draw(round):
...
+        # Also occurs for elimination rounds as everyone is just as live.
+
+        nteams = 0
+        if round.is_break_round:
+            break_size = round.break_category.break_size
+            nprev_rounds = round.break_category.round_set.filter(seq__lt=round.seq).count()
+            partial_two = nteamsindebate == 2 and not ispow2(break_size)
+            partial_bp = nteamsindebate == 4 and ispow2(break_size // 6)
+            if nprev_rounds == 0 and nteamsindebate == 2:
+                return partial_break_round_split(break_size)[0]
```
czlee, line 46: Isn't this an integer? (Haven't tested it, just making good on my promise to say things immediately rather than holding up till I get time to try it.)

tienne-B: `partial_break_round_split()` returns a 2-tuple (debates, bypassing), and we only want the first item.

tabbycat/adjallocation/preformed/anticipated.py
tienne-B: Fell into "smart" coding a bit here, but this shows how everything is related and similar. Treating as distinct cases wouldn't help with the number of `if`s (could even have more) as everything would be tested, but would be longer with duplication, so errors could slip in.

PR #1460, opened 2020-05-18
Fill submission data for AdjudicatorFeedback

tienne-B:
Saving Adjudicator feedback does not fill confirmer data, so that is now part of the base form. In addition, feedback does not have a discarded field so setting it was useless.


philipbelesky:
This looks good! I'll pass to CZ for review/merging.

tienne-B:
Just a note that I intentionally do not blank fields associated with confirmation (user/time) on de-confirmation, like with the `.save()` method of Submissions.

PR #1463, opened 2020-05-25
Refactor consumers.py to simplify code.

teymour-aldridge:
I think this makes it easier to read the code.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,9 +58,9 @@ author = 'Philip Belesky, Chuan-Zheng Lee, Étienne Beaulé'
 # built documents.
 #
 # The short X.Y version.
-version = '2.5'
+version = '2.6'
 # The full version, including alpha/beta/rc tags.
-release = '2.5.1'
+release = '2.6.0-dev'
 
 rst_epilog = """
 .. |vrelease| replace:: v{release}
--- a/tabbycat/settings/core.py
+++ b/tabbycat/settings/core.py
@@ -22,9 +22,9 @@ SECRET_KEY = r'#2q43u&tp4((4&m3i8v%w-6z6pp7m(v0-6@w@i!j5n)n15epwc'
 # Version
 # ==============================================================================
 
-TABBYCAT_VERSION = '2.5.1'
-TABBYCAT_CODENAME = 'Nebelung'
-READTHEDOCS_VERSION = 'v2.5.1'
+TABBYCAT_VERSION = '2.6.0-dev'
+TABBYCAT_CODENAME = 'O'
+READTHEDOCS_VERSION = 'v2.6.0-dev'
 
 # ==============================================================================
 # Internationalization and Localization


tabbycat/utils/tables.py
```diff
@@ -370,19 +385,7 @@ def _result_cell_two(self, ts, compress=False, show_score=False, show_ballots=Fa
...
-                    'link': reverse_tournament('old-results-ballotset-edit',
-                            self.tournament, kwargs={'pk': ts.ballot_submission_id}),
-                })
-            elif self.tournament.pref('ballots_released'):
-                cell['popover']['content'].append({
-                    'text': _("View debate ballot"),
-                    'link': reverse_tournament('results-public-scoresheet-view',
-                            self.tournament, kwargs={'pk': ts.debate_team.debate_id}),
-                })
+        self._show_ballots(show_ballots, cell, reverse_tournament, ts)
```
tienne-B, line 388: Calling this function implies that we wish to show ballots. The `show_ballots` conditional should be here and the other call, not in the new method.

tabbycat/utils/tables.py
```diff
@@ -370,19 +385,7 @@ def _result_cell_two(self, ts, compress=False, show_score=False, show_ballots=Fa
             cell['popover']['content'].append(
                 {'text': _("Total speaker score: <strong>%s</strong>") % metricformat(score)})
 
-        if show_ballots:
-            if self.admin:
-                cell['popover']['content'].append({
-                    'text': _("View/edit debate ballot"),
-                    'link': reverse_tournament('old-results-ballotset-edit',
```
tienne-B, line 377: Notice the link is different between both? Is this taken in consideration?

tabbycat/utils/tables.py
```diff
@@ -341,6 +341,21 @@ def _result_cell_class_four_elim(self, advancing, cell):
             cell['sort'] = 0
         return cell
 
+    def _show_ballots(self, show_ballots: bool, cell, reverse_tournament, ts):
```
tienne-B, line 344: `reverse_tournament` is a function with a file-scope. No need to have it as an argument.

tabbycat/draw/generator/powerpair.py
```diff
@@ -172,7 +172,7 @@ def _pullup(self, brackets, pos):
         Operates in-place. Does not remove empty brackets."""
         pullup_needed_for = None
 
-        for points, teams in brackets.items():
+        for _, teams in brackets.items():
```
tienne-B, line 175: If we only need the values (not the keys), how about `.values()`?

tabbycat/draw/generator/powerpair.py
```diff
@@ -469,9 +469,9 @@ def _pullup(self, brackets, indices):
 
         # Tuples: (teams_list, side, number_needed)
         # List by highest bracket first.
-        pullups_needed_for = list()
+        pullups_needed_for: list = list()
```
czlee, line 472: Hey, thanks for doing this. Can you give us a rundown on why type hints like this one are useful? I know it's legal, but my understanding was the type hints were mainly to be used when it's not obvious from context (mainly with function parameters and return values, which I definitely agree we could do with more of, especially with functions/methods called from other files), and for static type checkers (which should be able to figure out this one, since `list()` can't be anything other than a `list`).

tabbycat/draw/generator/powerpair.py
```diff
@@ -630,8 +630,8 @@ def _intermediate_brackets_2(cls, brackets):
 
             # Find lesser and greater of number of aff and neg teams.
             nums_teams = list(map(len, list(pool.values())))
-            n = min(nums_teams)
-            m = max(nums_teams)
+            n: int = min(nums_teams)
+            m: int = max(nums_teams)
```
czlee, line 634: Ditto for these ones.

tabbycat/draw/generator/powerpair.py
teymour-aldridge: Yes, that should make it easier.

tabbycat/draw/generator/powerpair.py
teymour-aldridge: The type here could be inferred – I like making types as explicit as possible, but that's just personal preference. My feeling is that having only partial typing means that not everything has a type (that's not `any`). While I agree that here you could infer that `pullups_needed_for` has type list, type annotations (using `typing.List`) would be useful to work out what the list is a list of.

tabbycat/draw/generator/random.py
```diff
@@ -2,7 +2,7 @@
 
 import random
 
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext as gettext
```
tienne-B, line 5: The `as` keyword "renames" the import. You're renaming a function to itself, so it isn't necessary to have an `as`.

teymour-aldridge: Oops!
I admit, I used find and replace to do that.

tabbycat/draw/generator/random.py
```diff
@@ -2,7 +2,7 @@
 
 import random
 
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext
```
czlee, line 5: Importing this as an underscore is standard in internationalization: 
https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#internationalization-in-python-code
https://docs.python.org/3/library/gettext.html#gettext.gettext

One can argue about whether this is a good thing, but it's the standard we've chosen to follow, since we translate a lot of strings and this is the most common way of avoiding clutter with `gettext`.

teymour-aldridge: Ok. Seems a bit strange, however.

philipbelesky:
Hey, thanks for looking at this file @teymour-aldridge. Would you consider adding tests for this page also? I am happy to help if needed. I mention it here in particular as this code path is potentially quite at risk of regressions — each function appears relatively simple but its deep in our async stack and our current instrumentation appears not to work in the `worker` processes that this runs on.

teymour-aldridge:
Sure.

tienne-B:
Sorry for the delays in reviewing this! While I think there is a global improvement with this patch, and the willingness to reduce the code to just what is being used, this patch goes a bit too far here. It does give clarity to indicate whether we only use the values or keys of a `dict`, but replacing unused variables with underscores doesn't seem that useful in that it doesn't improve the readability of a line (or even its length in some cases) and notating that its unused doesn't help in understanding the logic behind the code.

I suggest keeping the iterator variables (like `i` and `j`) and names of unused variables from tuple-returning functions (like `delete()`).

tienne-B:
I put your changes as #1684 (Sorry I couldn't update this PR directly)

PR #1467, opened 2020-05-27
Fix unnecessary assignment

teymour-aldridge:


tienne-B:
Seems to be duplicating #1468.

teymour-aldridge:
Yes, I accidentally branched from by local branch instead of the upstream branch.

PR #1468, opened 2020-05-27
Fix format call

teymour-aldridge:
There was a missing 'v'

### DIFF
--- a/tabbycat/adjfeedback/views.py
+++ b/tabbycat/adjfeedback/views.py
@@ -373,8 +373,6 @@ class BaseAddFeedbackIndexView(TournamentMixin, VueTableTemplateView):
                 'tooltip': _("Institution"),
             }, [team.institution.code if team.institution else TabbycatTableBuilder.BLANK_TEXT for team in tournament.team_set.all()])
 
-        adjudicators = tournament.adjudicator_set.all()
-
         adjs_table = TabbycatTableBuilder(view=self, sort_key="adjudicator", title=_("An Adjudicator"))
         adjudicators = tournament.adjudicator_set.all()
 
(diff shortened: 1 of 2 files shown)


tabbycat/adjfeedback/views.py
```diff
@@ -373,7 +373,7 @@ def get_tables(self):
                 'tooltip': _("Institution"),
             }, [team.institution.code if team.institution else TabbycatTableBuilder.BLANK_TEXT for team in tournament.team_set.all()])
 
-        adjudicators = tournament.adjudicator_set.all()
```
tienne-B, line 376: What's the use of this line? (Why keep it at all?)

teymour-aldridge: I don't know.

tienne-B: I suggest you remove the line completely; the statement does not change the state, so does nothing useful by itself (not assigned).

tienne-B:
Remove the f1da709 commit, as it's duplicated in your other PR (#1467) and is causing Travis to fail.

tienne-B:
I'll merge this after #1463 is corrected, and f1da709 is removed. You can use [`git rebase`](https://git-scm.com/docs/git-rebase) for that. The command would be `git rebase -i 
f1da709^` then you delete the first line (in Vim, press `d` twice) and save (type `:x`).

philipbelesky:
Hi @teymour-aldridge were you still interested in finish this off, or should we just directly commit the fix?

teymour-aldridge:
I'm pretty busy at the moment; directly comitting would be good!

tienne-B:
Removed the gitignore commit and merged the two commits that affected the same line.

PR #1470, opened 2020-05-29
Add sort property to break eligibility table

### LINKED ISSUE #1469: Break elligibility sort by number of speakers not working
Tabbycat Break Eligibility page on v2.4.0A is not allowing sorting by ESL speakers.

Initial View:

image

After click ESL Speakers once:

image

After clicking ESL speakers twice:

image

tienne-B:
For the number of speakers in categories per team, the text is stored as text, which may get sorted incorrectly. An integer is used with the new specific 'sort' property of the cells in the columns.

Should fix #1469.

### DIFF
--- a/tabbycat/breakqual/views.py
+++ b/tabbycat/breakqual/views.py
@@ -302,6 +302,7 @@ class EditTeamEligibilityView(AdministratorMixin, TournamentMixin, VueTableTempl
         for sc in speaker_categories:
             table.add_column({'title': _('%s Speakers') % sc.name, 'key': sc.name}, [{
                 'text': getattr(team, 'nspeakers_%s' % sc.slug, 'N/A'),
+                'sort': getattr(team, 'nspeakers_%s' % sc.slug, 0),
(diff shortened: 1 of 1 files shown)


philipbelesky:
Did you test this out? To me the problem seemed to the duplicate key used for both the checkbox and speaker count columns (this was being reported in the console). That was fixed in 326a5485da93305e3b1e98e5a22bf4b000e369eb, although I've also noticed that sorting by the checkbox fields doesn't work.

tienne-B:
I did test it, but I didn't see any difference. The status quo was working for me as well; the name/slug must have slipped my mind as my "Novice" slug was in lowercase. Could use the `sort` field for booleans ;)

philipbelesky:
Yea, the sort field is missing on the checkbox fields. Which then makes that property non-reactive when updates are triggered (i.e. the checkbox changes). I've fixed that in 9fa23c968272e4da728c1af3985f5881224fb4fc

PR #1473, opened 2020-05-30
Include round and room as ballot subtitle

### LINKED ISSUE #1472: Ballot Entry page should show debate venue
The ballot entry page does not show the debate venue on it. Showing the venue can be helpful when locating where the ballot is, whether physically or online in this case (for this tournament, ballot are being submitted in a discord channel specific to a room but I need to go back to the main results entry page to work out where to find the mark sheet for the debate).

If I am not mistaken the old ballot entry page did show this?

image

tienne-B:
To mirror the public ballot submission page showing that information.

Resolves #1472.

### DIFF
--- a/tabbycat/results/templates/public_enter_results.html
+++ b/tabbycat/results/templates/public_enter_results.html
@@ -10,12 +10,6 @@
 {% endblocktrans %}{% endblock %}
 {% block extra-head %}<meta name="robots" content="noindex" />{% endblock %}
 
-{% block sub-title %}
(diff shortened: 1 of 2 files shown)


philipbelesky:
Thanks for adding this. I initially thought this commit added the title only to the public form, rather than resolving it across all of the forms. Unfortunately I had already pushed a (flawed) commit of my own, which I have now overwritten to use these changes. My mistake!

tienne-B:
It'd be better if you merge this PR instead of copying this work in your commit and force pushing to `develop`. Your commit is still erroneous.

Issue #1471, opened 2020-05-30
Frequent Internal Server Errors


0zlw:
Running v2.4.0.A (https://whippersnappa-2020.herokuapp.com/whippersnappa-2020/) for a 60 team 3v3 tournament.

Tabbycat has been running fine all day but has at the end of R4 (last round of the day) started throwing internal server errors about 30-50% of the time you try to load a page. Affecting public submissions and admin error.

Debates resume at 11am NZT tomorrow so would appreciate if someone could have a look before then - I am happy to give backend access to the tournament.

Running autoscaling dynos.

![image](https://user-images.githubusercontent.com/3196468/83325752-e5253f00-a2c2-11ea-8892-b40473a43d31.png)

0zlw:
I did a "Restart all dynos" on the heroku app which seems to have resolved the problem. 

I received a "Redis Labs Email Alert – Number of connections is over 80% of your connections limit" alert although this was after I had restarted.

I don't think we were putting anything through then a normal debate tournament load with feedback and marksheet submissions.

tienne-B:
Heroku Redis and Redis Labs are different Redis services, but we use both of them; Heroku for channels and Labs for caching (storing pages so they don't need to be reloaded).

In your Dashboard, click on Redis Labs and it'll give you details on the Redis DB. Try following their instructions on [flushing](https://docs.redislabs.com/latest/rs/administering/database-operations/flush-db-crdb/) to free up some connections. The free tier has a limit of 30 connections, but connections should be short.

0zlw:
Thanks for the advice here. Things have gone back to normal but I will follow those steps if I encounter future issues.

tienne-B:
Don't see anything actionable.

PR #1474, opened 2020-05-30
Specify sets of keys in results assertions

tienne-B:
When asserting a DebateResult is loaded, sets are created to determine the equality of all components. However, sets of elements of dicts cannot be the same as for lists (without each element a 2-tuple). It should only test the keys of the dicts. This commit resolves this confusion.

One of the problems mentioned in #1471.

Closes BACKEND-39C
Closes BACKEND-397


tienne-B:
Didn't realize when sets are made from dicts that only the keys are taken.

PR #1478, opened 2020-05-31
Reduce some messages

tienne-B:


tabbycat/participants/models.py
```diff
@@ -94,7 +94,7 @@ class Person(models.Model):
     name = models.CharField(max_length=70, db_index=True,
         verbose_name=_("name"))
     email = models.EmailField(blank=True, null=True,
-        verbose_name=_("e-mail address"))
+        verbose_name=_("Email address"))
```
czlee, line 97: should be lower case?

tabbycat/notifications/consumers.py
```diff
@@ -34,10 +34,10 @@ def _send(self, event, messages, records):
         try:
             mail.get_connection().send_messages(messages)
         except SMTPException as e:
-            self.send_error(e, _("Failed to send e-mails."), event)
+            self.send_error(e, _("Failed to send emails."), event)
             raise
         except ConnectionError as e:
-            self.send_error(e, _("Connection error sending e-mails."), event)
+            self.send_error(e, _("Connection error sending emails."), event)
```
czlee, line 40: Just making a note, while trying to get this message to show up, I actually get
> 'NotificationQueueConsumer' object has no attribute 'send_error'
Because `send_error` isn't defined anywhere in this class or any of its parents.

I think this issue might go back as far as fa3f9e828b1dafdd3947c5d74843a5d46ab8cd19. The `send_error` method is defined in `tournaments.mixins.TournamentWebsocketMixin`.

tabbycat/utils/tables.py
```diff
@@ -630,7 +630,7 @@ def add_motion_column(self, motions, show_order=False):
             } for motion in motions])
 
         motion_data = [{
-            'text': motion.reference if motion.reference else _('?'),
+            'text': motion.reference if motion.reference else _('??'),
```
czlee, line 633: Hmm, I tried to test this too, but `motion.reference` is a mandatory field, lol. Where does it show up?

tabbycat/notifications/consumers.py
czlee: Opened #1547 to keep track of this.

PR #1495, opened 2020-06-07
Include API in next release

### LINKED ISSUE #910: CSV API for feedback
This should be relatively simple to setup using Django Rest and a library for CSV output. It probably wouldn't be a huge effort, once the view is setup, to have both JSON and CSV outputs.

### LINKED ISSUE #911: CSV API for adjudicators
To support #910 there should be a list of basic adj information in terms of name, institution etc. This would allow the feedback list to refer to primary keys.

### LINKED ISSUE #1301: Create participant import API endpoints

### LINKED ISSUE #1428: Support variable permissioned API endpoints
As with public/admin/assistant views, API views should be partitioned with permissions to allow for public use when appropriate. In a second point, if permissions were subdivided more for specific tasks, we could "secure" endpoints and passwords for specific tasks. For an example, checkins could be safely delegated without giving full admin/assistant access.

For public endpoints, we would be looking at two things: while read-only, different serializers would be required if the object has sensitive information. The `get_serializer_class()` method of ViewSets can help. The endpoints would also have to be restricted by preferences, for if public views are activated or not.

### LINKED ISSUE #1436: Postings output to API
Hi there,

I'm working (in a non-technical capacity) with Yaatly.com and am hoping that we can ask for postings to be output in a CSV file.  With such a file from TabbyCat, we can directly upload the posting to Yaatly and automatically create rooms for an online tournament (complete with team names and positions as well as judge names).

Any chance this is simple to do?

tienne-B:
JSON should be preferred over CSV, and would suggest a new task if there's interest for GraphQL.

Closes #1436
Closes #1428
Closes #1301
Closes #911
Closes #910


tienne-B:
- Designed with Pairing objects in mind rather than Debate objects themselves. Just happened that it fit well with the model object. I wouldn't change it now, as it's already being used and there isn't much different.
- We can do that when we start diverging more from the models. Separating them is just another layer of complication, and enums shouldn't really rename or reduce themselves. Don't think it's necessary at this time.
- Paging is already available with limit/offset, or filters can be added for source / target / round
- Feel like the `source` field suddenly changed on me :( Fixed.
- Fixed.
- Fixed.
- Added field. If empty will be the truncated `reference` field.
- Will raise a 400 Validation error
- More precise message added.
- Will look into.

tabbycat/api/serializers.py
```diff
@@ -398,6 +404,11 @@ def __init__(self, *args, **kwargs):
             if not t.pref('public_break_categories'):
                 self.fields.pop('break_categories')
 
+    def validate(self, data):
+        if data['use_institution_prefix'] and data['institution'] is None:
+            raise serializers.ValidationError("Cannot include institution prefix without institution.")
+        return super().validate(data)
+
```
czlee, line 412: There's a `clean()` method of most models, including `Team.clean()` and `Round.clean()`. Does the REST API library have a means of leveraging this? Would using it help with DRY?

czlee:
> Designed with Pairing objects in mind rather than Debate objects themselves. Just happened that it fit well with the model object. I wouldn't change it now, as it's already being used and there isn't much different.

Hmm, okay. We normally use the terminology "debate" in Tabbycat. "Pairing" is fine, I'd just brace ourselves for potential confusion. I guess the different name might be a good flag for developers not to expect it to behave like a "debate" in the user interface, but to think of the API model as different.

> We can do that when we start diverging more from the models. Separating them is just another layer of complication, and enums shouldn't really rename or reduce themselves. Don't think it's necessary at this time.

I guess I was thinking more it might be now or never, if we want to preserve compatibility. We've renamed enums before, normally because of some change in functionality. (For example: there used to be a distinction between "first elimination" and "subsequent elimination" draw types.) Also, the enum letters aren't very enlightening, and the enum values are more designed for internal convenience than informativeness.

That said, on reflection, it might be better to redefine this and _clearly_ break things as the need arises, rather than try to design how this information is communicated now. So let's leave it as it is. We can add new fields under different names and deprecate old fields later, if the need arises.

> Paging is already available with limit/offset, or filters can be added for source / target / round

That's cool, then there should also be some sensible default limit. There's a general principle where calling something without defaults should do something sensible, even if it's not quite what you want, and I don't think going ahead with all of the several thousand entries is sensible, since it might overload systems.

tabbycat/api/views.py
```diff
@@ -411,7 +414,6 @@ def get_queryset(self):
             Prefetch(
                 typ.__name__.lower() + "_set",
                 queryset=typ.objects.all().select_related('question', 'question__tournament'),
-                to_attr=typ.__name__,
```
czlee: Removing this line reduced the number of SQL queries for a 70-feedback GET from to 1328 to 8. I now suspect that when I had suggested that paging be enabled by default with sensible limits, the actual symptom I was experiencing wasn't too many objects, but just a N+1 SQL query issue. I still think pagination should be enabled by default, but just wanted to say that it's not the exact issue I was having.

czlee: Actually… I might go back on myself with the pagination thing, let me explore this a little first.

tabbycat/api/serializers.py
tienne-B: I'll check again, but I think I tried and there was some compatibility problem between DRF's ValidationError and Django's. I do like my `==` solution for XNOR logic ;)

czlee:
On pagination: Actually, having resolved the feedback problem, it now seems overkill to enable pagination for everything, but we'd also want to have some protection against queries that would overload the system. I suppose for an MVP we can overlook this though, and then see what problems actually arise once it's in the field, then that would probably give us a better understanding of what we should do.

tabbycat/tournaments/models.py
```diff
@@ -242,17 +242,17 @@ def current_rounds(self):
...
                 return [r]  # short-circuit everything else
             elif r.debate__count > 0:
-                current_elim_rounds.setdefault(r.break_category, r)
+                current_elim_rounds.setdefault(r.break_category_id, r)
         return [
-            current_elim_rounds.get(category)
-            for category in self.breakcategory_set.order_by('seq')
+            current_elim_rounds.get(category.pk)
+            for category in self.breakcategory_set.all()
             if category in current_elim_rounds
```
czlee, line 256: You missed the `.pk` in line 256, so this function now always returns an empty list when it's not short-circuited.

On another note, the `.order_by('seq')` actually was important, but I think this probably should've been commented rather than left implicit. The returned list is intended to be in the order specified by the tab director using the `seq` fields in break category. As for the first `order_by` in line 245, that makes sure that if there are multiple break rounds in a category that aren't completed but have debates, it's the first one that is considered to be the current round, rather than whichever happens to come first in the database queryset order. While it's technically redundant because of the `Meta.ordering` options set in `BreakCategory` and `Round`, it's made explicit here so because it's critical to the function of the method, as opposed to a user interface nicety.

tabbycat/api/serializers.py
```diff
@@ -117,10 +117,10 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = Round
-        fields = '__all__'
+        exclude = ('tournament',)
 
     def validate(self, data):
-        if (data['break_category'] is None) == (data['stage'] == Round.STAGE_ELIMINATION):
+        if (data.get('break_category') is None) == (data.get('stage') == Round.STAGE_ELIMINATION):
```
czlee, line 123: Ah, this "XNOR". Haha, it's a bit obscure but you're right, I can't think of a cleaner way, I'd probably do the same.

tabbycat/templates/registration/password_change_form.html
```diff
@@ -12,13 +12,21 @@
...
   <div class="col-lg-8 ml-lg-auto mr-md-auto">
 
+    {% if user and user.is_staff %}
+      {% blocktrans trimmed asvar p1 with api_token=user.auth_token.key %}
+        API Token: <samp>{{ api_token }}</samp>
+      {% endblocktrans %}
+      {% trans "The API token will allow you to authorize external applications to access the site as staff." as p2 %}
+      {% include "components/explainer-card.html" with type="info" %}
+    {% endif %}
+
```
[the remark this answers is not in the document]
tienne-B: That's more involved as we'd need a new global page (thus a migration is made), and probably an HTML file and set context. Adding it to the Change Password page is much easier as the user object is already there. Will think about a new page maybe.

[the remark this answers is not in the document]
czlee: Wait, why would that entail a migration?

But yeah there'd be the HTML file and everything. I think this is fine for now, it's easy enough to spin off later when the time comes.

PR #1496, opened 2020-06-08
Add a URL field to Venues; closes #1430

### LINKED ISSUE #1430: Add venue information field
The idea is to add a "venue information" field to the [`Venue` model](https://github.com/TabbycatDebate/tabbycat/blob/af5f10b9cd3e37c28b2b3790106d80af262a30fe/tabbycat/venues/models.py#L7-L15), which would contain additional information about the venue. The first motivation is to help support online tournaments, but the field should be able to be used more generally. For example, it might contain:
* in an online tournament, a link to an online virtual room
* a link to a pin drop in Google Maps
* an image of a map
* special directions to avoid confusion, for example, "take the stairs on your left as you walk in, not the one on your right"
* detailed information about accessibility

Minimum implementation I have in mind:
1. Add an `information` field to `Venue`. I suggest making it a [`TextField`](https://docs.djangoproject.com/en/3.0/ref/models/fields/#textfield).
2. Add code to [`BaseTableBuilder.add_debate_venue_columns.construct_venue_cell`](https://github.com/TabbycatDebate/tabbycat/blob/af5f10b9cd3e37c28b2b3790106d80af262a30fe/tabbycat/utils/tables.py#L709-L739) to display this information.
    * `cell['popover']['content']` (line 737) is a list of dicts. Another dict should be added to this list for the venue information.
    * Django normally escapes HTML characters for security reasons, so in order for the venue information field to support HTML, the string should be marked safe using [`mark_safe`](https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.safestring.mark_safe).
    * If the field is blank, and there's nothing else to display in the popover, it should not display the popover at all

This implementation doesn't address how to input the information data, but for now that can be done through the Edit Database area.

Flagging potential conflicts with #1382, but so long as this is done on the `develop` branch I think it should be fine. That said, if we want to escalate the urgency of this patch, we might want to think about how it affects our release cycle.

philipbelesky:
This is pretty simple, but I think handles the main use case without over complicating anything.

The conditionals in `round_team.html` are getting pretty ugly, but I don't think there's an easier way to conditionally include a url in a `blocktrans`?


czlee:
Yeah, I can't think of another way. The fact that we can't "build" strings with translation is kind of annoying.

Did you intend to replace all instances of "You are debating in" with "Your debate is in"? I think you might've missed a couple of spots?

philipbelesky:
The switch between those two types wasn't deliberate per se; I don't think it needs to be standardised. The main switch is phrasing the info as two sentences. All good to merge?

czlee:
I noticed that on the third-person pages, the venue sentence still ends up showing in the second person. I moved the venue sentence outside the conditionals, and also moved it to a separate cell. That seems to simplify the conditions a bit?

I was going to remove the team name from the venue sentence, but then I realized that in BP when sides aren't confirmed, the side sentence doesn't show at all, and it'd just be a lone "Their" in the first sentence. This is pretty minor, but I ended up restoring it, and changing it to `{{ team }}'s debate is in {{ room }}`. One annoying thing about this string is that it doesn't play nice with some languages, _e.g._ French, where the form of the possessive changes depending on whether the owner starts with a vowel (like a/an in English).

Thanks for bringing attention to this!

tienne-B:
I'd think the team does not have to be specified for the room sentence, because it is evident from the matchup strings. "The debate is in ..." _ou_ "le débat est dans (la salle) ..."

czlee:
I thought that at first, but realized the the matchup strings don't always show. If it's BP and sides aren't confirmed, then there's no matchup string.

tienne-B:
Or that the box where the string is shown is called "In this round" and the page is for specific participants. Even without the matchup strings it seems evident. However, that could also apply to the team name in the matchup (Becoming "You"/"They're")

czlee:
Yeah, I think when I originally wrote that sentence my idea was that the highlight (in purple) would draw attention to it as the focus, and it would deliver information as a human would read it, hence the single natural language sentence. More than one way to skin a cat, though—for example, the following would deliver exactly the same information:

> Side: Negative
> Venue: East 114
> Opponent: Goldenrod 1

I prefer the natural language sentence, despite the complications, since I think it's the clear<i>est</i> and, well, most friendly. For this reason, most of the key sentences in that box are in natural language ("The debate starts at 11:50 a.m."). But it's certainly not unclear using the format above, which does have the advantage of being easier to scan once you're used to the format.

philipbelesky:
Thanks for the commits comments. I think the two sentence structure is fine for now and the conditional structure better than what I had. Open to further improvements though. Closing this as it doesn't seem to have picked up that its been merged into develop.

PR #1498, opened 2020-06-12
Synchronously check in/out in auxiliary methods

tienne-B:
It can happen that the checkins consumer is disconnected, and if that's the case, methods of performing check ins/outs that don't involve directly opening the websocket may silently fail. This happens with private URLs and API features. So, instead of having the websocket perform the action, these pages will perform it themselves and broadcast the result, which would be picked up if the connection is active.


philipbelesky:
This makes sense, thanks for the PR. I'm a little wary around changing the socket code prior to release, but think this should be ok as its confined to privateURLs.

tienne-B:
Clarification that it affects not only private URLs but API endpoints. Also socket code is not changed, just bypassed ;)

PR #1504, opened 2020-06-15
Update deploy_heroku.py script

czlee:
* Removes the explicit ordering of buildpacks. I _think_ we used to specify order because the buildpacks would be added at the top. It appears Heroku has updated this so that it adds buildpacks at the bottom (which I find more intuitive), and also because indexing starts from 1, the current script ends up having them in the order: [nodejs, python, nginx], while the app.json file has [nginx, nodejs, python]. As far as I can tell, it doesn't seem to make a difference. Did it use to?
* Adds timeout argument for Heroku Redis
* Adds time zone argument

Mostly I'd like feedback on the buildpack ordering before merging.

### DIFF
--- a/deploy_heroku.py
+++ b/deploy_heroku.py
@@ -62,6 +62,8 @@ config_group.add_argument("--tab-cache-timeout", type=int, default=None, metavar
                           help="Set the tab page cache timeout to TIMEOUT")
 config_group.add_argument("--enable-debug", action="store_true", default=False,
                           help="Enable Django debug pages")
+config_group.add_argument("--time-zone", type=str, default="Australia/Melbourne",
+                          help="Time zone name from the IANA tz database")
 
 # Import tournament arguments are copied from importtournament.py, and should be
 # updated when these options in importtournament.py change.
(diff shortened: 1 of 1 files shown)


deploy_heroku.py
```diff
@@ -171,12 +173,12 @@ def get_git_push_spec():
 
 # Add the redis add-ons (the heroku one needs a config flag)
 run_heroku_command(["addons:create", "heroku-redis:hobby-dev",
-                    "--maxmemory_policy", "allkeys-lru"])
+                    "--maxmemory_policy", "allkeys-lru", "--timeout", "1800"])
```
philipbelesky, line 176: @czlee Just a heads up that we should keep an eye on socket timeout behaviours. Daphne defines its own `websocket_timeout` and channels defines a `group_expiry`. I'm not super clear how these all interact.

philipbelesky:
I think the buildpack order did used to matter, as there was some complexity of how/when nginx communicates with the dyno process. Looking at the docs it seems like it doesn't matter any more. Any failures here should also be immediately obvious on deployment.

deploy_heroku.py
czlee: I see, to be honest I wasn't sure, I was just copying the app.json manifest.

Issue #259, opened 2016-01-18
Consider switching admin UI library


philipbelesky:
Suit seems to be pretty slow to updates, and is looking a bit dated. Consider alternatives such as:

- [Jet](http://jet.geex-arts.com)
- [Django Material](http://forms.viewflow.io)
- The default admin

Relates to #171

philipbelesky:
Have swapped out suit for [Jet](http://jet.geex-arts.com) and liking it so far. Will close for now, reopen if there are any usability issues or bugs with it.

PR #1507, opened 2020-06-15
Upgrade Python to 3.8.2 and Django to 3.0.7

### LINKED ISSUE #1501: Upgrade to Django 3.x

czlee:
Closes #1501

Mostly making a PR some there's somewhere to discuss with reference to code, if needed.


czlee:
So… first roadblock! The django-jet package isn't compatible with Django 3. There's a nice loud message on their README: https://github.com/geex-arts/django-jet

Basically, their reincarnation is the new [Jet Bridge](https://github.com/jet-admin/jet-bridge), which seems way overkill for what we want. It also requires a Jet account, which presumably means that everyone who installs Tabbycat will have to create a Jet account, and it won't just add itself seamlessly by being included in Tabbycat.

Barukimang has patched django-jet for Django 3.0: https://github.com/geex-arts/django-jet/issues/430
But if django-jet's no longer under active development, I'm inclined to say we should ditch it altogether.

Of course, we could just remove Jet altogether and revert back to the original Django admin site for the Edit Database area. But there's a good reason we had Jet in the first place—the Django admin site isn't that friendly (and isn't meant to be).

Or we could find a substitute. Thoughts?

czlee:
… Otherwise, everything seems fine. Our Travis PR hook is broken, but it happened on the branch: https://travis-ci.org/github/TabbycatDebate/tabbycat/builds/698352467

tienne-B:
Not the first time changing DB admin UI: #259 We could try the default; it may have improved.

czlee:
> We could try the default; it may have improved.

Just did. It has not. 😅 But on a quick search, there are countless others!

czlee:
django-suit has a v2.0 in development that according to [this issue](https://github.com/darklow/django-suit/issues/475) is more or less functional, but not polished.

Other packages here: https://djangopackages.org/grids/g/admin-styling/

czlee:
* Django Suit 2.0a1 seems alright, if we're prepared to take a punt on a pre-release version that's unlikely to be properly released any time soon.
* Django Grappelli I'm not really a fan of, for our application. 
* Django Admin Tools doesn't seem right for us at first glance either.

philipbelesky:
Only a few of those other packages are themes, and most of them are relatively minimal.

I'd be inclined to stick with the (patched) Jet and push the question back to whenever a future breaking change occurs. By then Suit 2.0 may be released, a Jet fork may be solid, or there may be other options.

czlee:
I see. My main concern about running with the patched Jet is that I don't think it's officially released on pip (understandably so, given that it's a fork). We could reference Barukimang's repository directly from our requirements file, but I think I'm just worried about the dependency reference potentially breaking one day, since it's not "official", so to speak. For a small project I wouldn't bat an eyelid, but I don't know how many people deploy Tabbycat sites per day these days 🙈

config/requirements_core.txt
```diff
@@ -30,7 +28,13 @@ channels==2.2.0                         # Channels; also includes the Daphne ser
...
+ipython==7.15.*
+munkres==1.1.2                          # Algorithm for adjudicator allocation
 redis==3.3.8
 qrcode==6.1                             # QR codes for printed private URL sheets
 html2text==2019.8.11                    # Compatibility with plain-text email clients
+
+# Django admin (Jet)
+# This is an unofficial fork of no-longer-maintained geex-arts/django-jet,
+# patched to support Django 3.
+git+https://github.com/Barukimang/django-jet.git@1.0.8-django3
```
czlee, line 40: @philipbelesky Just drawing attention to this line, which is the patched django-jet

philipbelesky: Ok. Also to address your comment above, we could always do a fork ourselves if we were concerned about it vanishing or having a breaking update.

czlee: Yeah, we could. I'd describe my concern as more about the consequence than the probability of it happening—I doubt it'll disappear, but if it does, Tabbycat will become undeployable until we release a patch with our own fork.

czlee:
More digging:

* Bakurimang has the repository I referenced above: https://github.com/geex-arts/django-jet/compare/1.0.8...Barukimang:1.0.8-django3
* RamsesMartinez also has one: https://github.com/geex-arts/django-jet/compare/1.0.8...RamsesMartinez:2.0.8, which he has released on pypi: https://pypi.org/project/r-django-jet/

It seems that there are mostly just two changes to the original django-jet repository:
1. The [`django.utils.encoding.python_2_unicode_compatible`](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.encoding.python_2_unicode_compatible) decorator [was deprecated in Django 2.0 and removed in Django 3.0](https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis).
    * Bakurimang deals with (1) by importing it from `six` instead ([compare](https://github.com/geex-arts/django-jet/compare/1.0.8...Barukimang:1.0.8-django3#diff-c9f8a275d83f89e5328ba30ca5e8acf6))
    * RamsesMartinez deals with (1) by removing it altogether ([compare](https://github.com/geex-arts/django-jet/compare/1.0.8...RamsesMartinez:2.0.8#diff-c9f8a275d83f89e5328ba30ca5e8acf6))
    * We don't care for Python 2 compatibility, and Django 3.0 doesn't support Python 2 anyway, so it makes more sense to me to remove it altogther.

2. The `admin_static` template tag library [was deprecated in Django 2.1 and removed in Django 3.0](https://docs.djangoproject.com/en/3.0/releases/3.0/#features-removed-in-3-0).
    * Both Bakurimang and RamsesMartinez deal with this by replacing it with `static`.

3. RamsesMartinez has also updated a bunch of minified CSS files. I don't know if this would have any effect?

Do you have thoughts on picking an existing fork vs making our own? Having figured out the above, I'm now kind of divided about it. Again, for a personal project I'd just pick one and not worry too much about it, but I guess this isn't really a personal project anymore.

philipbelesky:
Using r-django-jet sounds good if it is up pypi. Altered css files should be no problem.

PR #1508, opened 2020-06-15
Pass whole context to serializers in create

tienne-B:
Avoid problems with missing context.

Closes BACKEND-3BT


czlee:
I see, makes sense.

I notice (and hadn't seen before) that we currently use a bunch of `self.fields.pop()` statements in constructors to enforce security. We probably want to see if we can integrate this into a better framework, so that serializer classes only need to specify the information, not the logic. For example, one idea would be to add a (special) `staff_only_fields` to `Meta` classes, and writing a base class for model serializers within Tabbycat to use this. I notice it might get more complicated with honoring tournament preferences, though.

PR #1511, opened 2020-06-17
Re-work conditionals for names in registration card

tienne-B:
There's no harm in showing the full name if the short name is shown, even when identical. That conditional is thus removed. Thus, it should show on private URL pages.

If one of the names cannot be shown, the emoji should be moved to the full name to still be shown.

### DIFF
--- a/tabbycat/participants/templates/team_registration_card.html
+++ b/tabbycat/participants/templates/team_registration_card.html
@@ -9,17 +9,19 @@
       </h4>
     </div>
 
-    {% if not use_code_names and team.short_name != team.long_name %}
+    {% with show_codes=team.code_name and pref.team_code_names != 'off' show_names=not use_code_names and team.short_name != team.long_name %}
       <div class="list-group-item">
-        <strong>{% if participant %}{% trans "Team name:" %}{% else %}{% trans "Full name:" %}{% endif %}</strong>
-        {{ team.long_name }}
-        {% if not use_code_names and team.code_name and pref.team_code_names != 'off' %}
+        {% if private_url or admin_page or show_names %}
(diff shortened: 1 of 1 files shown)


tienne-B:
Should qualify for hotfix as the status quo can lead to private URL pages not showing code names even if used.

czlee:
I'm confused, what's the motivation for this change / the case where the status quo is counterintuitive/undesirable?

tienne-B:
If the short name and long name of a team is identical (_e.g._ doesn't use an institutional prefix) and uses code names, the private URL registration box shows none of the names.

czlee:
Oh, I see. So as originally conceived, the short name of the team would be in the title of that box, because that was back when participant pages were always in the third person, before we had private URL landing pages. But now, because private URL landing pages are by person and not by team, that registration card shows the person's name, and the (short) team name is nowhere to be found.

That registration card is still fundamentally about the team, so I feel like the remedy is to change its title to reflect what it actually does? Right now it looks like it's listing the "speakers" associated with a speaker, which is a minor detail but a little weird 😂
image

I'm okay in principle with showing both the short and long names in body cells, though I still think we should only show the name once in the body if the short and long names are identical. In that case, the title on the private landing page can just be "Team Information", without the team name.

tienne-B:
Can't be team information as it also has speaker categories. The box still only shows the long name and never the short name.

philipbelesky:
Maybe we can delineate (a separate box?) between speaker and team data. That may help cases (e.g. institution or categories) where it is ambiguous what the data is linked to.

tienne-B:
I wouldn't say that it's ambiguous, just clunky. It's a good idea to separate, but the speaker box wouldn't have enough content to justify it being separate in my view. We could try though in develop; think such a change wouldn't qualify as a hot fix.

philipbelesky:
Sure, happy to revisit as an enhancement on develop.

Issue #1491, opened 2020-06-03
Use ORM for metric rankings


tienne-B:
SQL window functions allow queries to have fields which reference other records, which can be used to get statistics like the percentile of the value in the row in the distribution of all in the partition. Django supports these annotations natively in its ORM.

The window function that seems useful is `RANK()`, which assigns an order on the records. This function could be used in standings to defer rank calculations to SQL for rankings and subrankings.

czlee:
I think I looked at this once, then decided that for more complicated ranking annotations like subranks it was too obscure and unmaintainable to be worth the effort. Will be happy if there's a clean way to do it, analogous to `TeamScoreQuerySetMetricAnnotator`, though.

tienne-B:
Subranks could be accomplished by using the ranking criteria as partitions (`ORDER BY` becomes `PARTITION BY`)

PR #1512, opened 2020-06-18
Optimize use of querysets for standings

tienne-B:
See commit. Instead of re-evaluating querysets with each query metric, these commits convert the process into two distinct steps, with each step applied to the metrics at once.
1. Annotate the queryset
2. Annotate the standings

Avoiding re-annotations give a performance boost, as even though the query is more complex, there are fewer.

These changes start the work on #1491, with the goal of having rankings also determinable with SQL in some/most cases.


tabbycat/standings/teams.py
```diff
@@ -256,16 +264,12 @@ def get_field(self):
...
-
+    def get_annotation(self, round=None):
         # If the number of ballots carried by every team is an integer, then
         # it's probably (though not certainly) the case that there are no
         # "weird" cases causing any fractional numbers of votes due to
         # normalization. In that case, convert all metrics to integers.
-        if all(tsi.metrics[self.key] == int(tsi.metrics[self.key]) for tsi in standings.infoview()):
-            for tsi in standings.infoview():
-                tsi.metrics[self.key] = int(tsi.metrics[self.key])
+        return Cast(super().get_annotation(round), output_field=IntegerField())
```
czlee, line 272: This changes the numerical value of the annotation in the situation alluded to in the comment just above it.

(To try it out, try a five-person panel, or just emulate one by changing the "votes possible" field in a team score in the database.)

tienne-B: Oh now I see. Casting to integer won't require a modification to the queryset, just the standings, and the cast won't change any numerical values.

czlee: The rounding can change things, because the metric isn't actually guaranteed to be an integer, at least not with the current implementation. It normalizes to a fixed number of adjudicators (currently 3), so that every debate has equal weight, even when panels are of different sizes, so if (e.g.) a team got 3 judges from a panel of 5, the value of the metric will be 1.8, but the cast will round it to 2.

tienne-B: I'm talking about your approach, before this change. If the value was 1.8, no values would be rounded as 1.8 != 2.

czlee:
What's the anticipated use case for numbers of thirds and number of fourths? Should these be sorted in ascending order (fewer fourths beats more fourths)?

tienne-B:
Think it would be good to include just for completeness, and with an order similar to 1sts/2nds in that like 2nds, it would be odd to use without higher point value metrics not being before it in the precedence. There, such an ordering would not really matter for 4ths, but more 3rds with ties in ½ is preferable than fewer.

czlee:
Ah. I thought about this when implementing WUDC rules—it's not obvious, but under WUDC rules (despite its being specified in the constitution), number of thirds doesn't make a difference either, because they're all used as a tiebreak after number of team points. If you fix the number of team points, firsts and seconds, you've also determined the number of thirds (and fourths).

I'd rather include them only if there's some actual use for it. It doesn't have to be widespread, but it should be something that's not hypothetical.

tienne-B:
That's true. Removed the commit. Ignore the Travis build fail; it's due to the linting problems that was fixed after the commit was made.

Tabby85 (commented):
You owe me money for using me

czlee:
One issue I noticed: A team who doesn't have any debates ends up placing first with this new code. (They place last in the old code, or if the system with non-queryset metrics is invoked.) After I noticed this I added a unit test for the scenario in 2e3b593beba32a99c8e532227d23fdd17cb34b65 (on develop), feel free to cherry-pick it into this branch (or whatever other history management works).

image

tienne-B:
Yep, forgot to cherry-pick that commit from USUDC.

tienne-B:
The travis build should be ignored with the comment https://github.com/TabbycatDebate/tabbycat/commit/2e3b593beba32a99c8e532227d23fdd17cb34b65#r43155097 .

czlee:
Very cool! Nice work 👌🏻

czlee (approved):
Very cool! Nice work 👌🏻

So if I understand right, all the old non-SQL code is still there as a fallback in case querysets don't work on it because of e.g. a repeated metric? Maybe just want to add some comments to make this clear to future developers—might be confusing because it at first seems like a DRY violation, but actually it has a (good) purpose? Or if I misunderstood, then comments to make clear whatever should be understood by future readers.

PR #1513, opened 2020-06-18
Return to private landing page if checkin errors

tienne-B:
Closes BACKEND-3C3

To be clearer, 866ac902b42f6d53bdfd2d7d42571c5c1d5beb07 made the channels pass a dict to the checkins websocket. However, there would be no dict created if no action was taken (say checking in/out when already in that state). This causes a crash. By returning before sending the web socket message, there is no necessary action to take after determining nothing to be done, and so we can redirect back safely and without the unfound variable error.

### DIFF
--- a/tabbycat/checkins/views.py
+++ b/tabbycat/checkins/views.py
@@ -269,7 +269,7 @@ class ParticipantCheckinView(PublicTournamentPageMixin, PostOnlyRedirectView):
         except Person.DoesNotExist:
             raise Http404("Person does not exist")
         except PersonIdentifier.DoesNotExist:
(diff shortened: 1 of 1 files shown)


czlee:
What was it doing before?

tienne-B:
Go back to their landing page with a message. However now there's a 400 if it errors.

czlee:
Wait, I'm confused, so what's the effect of the change? To introduce a 400?

tienne-B:
Whoops, meant 500. It will now redirect to their URL page with a message like "Already checked in/out"

PR #1515, opened 2020-06-20
Remove duplicate institution name

czlee:
This removes three lines added in https://github.com/TabbycatDebate/tabbycat/commit/bd1d7231a569ae9afc7ce5ff6e35f8e93c74f010#diff-39d5e7bd3528e2f207017c56d7c13024R558-R560, which I believe seem to cause adjudicator institutions appear twice:
image

@philipbelesky, just checking I haven't missed anything / this doesn't mess up the work to expand the "show adjudicator institutions" option before?


philipbelesky:
Looks good!

PR #1518, opened 2020-06-21
Expand boolean variables within 'with' statement

tienne-B:
The 'with' templatetag does not seem to support boolean operators, as
were used for team registration cards. This commit expands these vars
where used, keeping in mind Django's precedence.

Fixes BACKEND-3CH.

### DIFF
--- a/tabbycat/participants/templates/team_registration_card.html
+++ b/tabbycat/participants/templates/team_registration_card.html
@@ -9,19 +9,20 @@
       </h4>
     </div>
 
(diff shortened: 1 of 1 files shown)


czlee:
rip, how did we miss this? fwiw I actually prefer this anyway, I think it's clearer than with the `with` tag.

tienne-B:
I much prefer the `with` so that it better shows what the logic is to do and if used in various places. Also don't like applying _De Morgan_ as it complicates noticing similarities in logic.

czlee:
Merging, there are still some outstanding edge cases to clean up but this fixes the crash issue so merging to roll out in hotfix 2.4.2.

PR #1519, opened 2020-06-22
Partially rework record pages

czlee:
For team names:
- Moved emoji to a separate line so that its display is conditional only on one thing
- Show both short name and long name whenever they differ
- Show code name whenver code names in use, but show "No code name assigned" in muted text if the team doesn't have one
- Change line breaks (`<br />`) to divs (`<div>...</div>`), mostly to avoid problems with stray line breaks in edge cases
    - Not sure if a class should be added to these divs
    - I sort of think we might have enough info there now that a two-column table would be better, so might investigate that

More generally:
- Add `list-group-item-dark` to items where it is not public information. I'm not wedded to this particular style, but I feel like it'd be nice to visually differentiate it with more than just a message.

Extends #1511 and #1518, as discussed. Work in progress for now.

I think this should be targeted at Nebelung, since it's not just a bug fix.

### DIFF
--- a/tabbycat/participants/templates/team_registration_card.html
+++ b/tabbycat/participants/templates/team_registration_card.html
@@ -55,7 +55,7 @@
         {% for speaker in team.speakers %}
           {{ speaker.name }}{% if not forloop.last %}{% trans ", " %}{% endif %}
         {% empty %}
-          <span class="text-muted">{% trans "None" %}</span>
+          {% trans "None" %}
         {% endfor %}
       </div>
     {% endif %}
@@ -67,7 +67,7 @@
           {% for category in team.break_categories.all %}
             {{ category.name }}{% if not forloop.last %}{% trans ", " %}{% endif %}
           {% empty %}
-            <span class="text-muted">{% trans "None" %}</span>
+            {% trans "None" %}
           {% endfor %}
         </div>
         {% if participant %}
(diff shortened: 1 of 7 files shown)


tabbycat/participants/templates/team_registration_card.html
```diff
@@ -10,17 +10,38 @@ <h4 class="card-title mb-0">
...
         {% endif %}
-        {% if team.code_name and pref.team_code_names != 'off' %}
-          <br />
-          <strong>{% trans "Code name:" %}</strong>
-          {{ team.code_name }} {% if pref.show_emoji %}({{ team.emoji }}){% endif %}
+        {% if pref.show_emoji %}
+          <div>
+            <strong>{% trans "Emoji:" %}</strong>
+            {{ team.emoji }}
+          </div>
```
tienne-B, line 44: Thought we were to append the emoji to the last displayed name rather than a new line for one character. Also, there might not be an emoji assigned.

czlee:
I can't remember what was still "in progress" about this, lol

tabbycat/participants/templates/team_registration_card.html
czlee: Yeah, I think at some point I came to the view that with all the conditional statements above it was too much effort to maintain that principle, lol. Definitely open to better ways to present this—I think the current layout isn't super easy to parse.

Good point about the possible null emoji, I added a line for that similar to code name.

czlee:
I haven't done a very good job of it because I can't bootstrap well, but I gave tables a go to see what might happen:

Original:
image

As a table:
image

tienne-B:
I like it

PR #1520, opened 2020-06-25
Make API more resilient to errors

tienne-B:
This commit changes a bit of square notation towards .get() as model field validation (IntegrityError, TypeError) is done in the superclass (and made into ValidationError). It also adds checks when directly saving objects.


tienne-B:
* BACKEND-3D1: Incorrect field (`use_institution_prefix` became `use_institutional_prefix`) gets ignored until caught by DRF
* BACKEND-3CS: Default conflicts are created only when not already created (`.get_or_create()`)
* BACKEND-3E1: Add a default empty list for participants in eligibility endpoints
* BACKEND-3E2: Fix tournament field for speakers in eligibility endpoint to `team__tournament`

czlee:
I notice a lot of lines provide some sort of default value for `dict.get()` calls. This would avoid a `KeyError`, but what's the general principle for determining what that value should be? For example, in this line, `'stage`' defaults to elimination:
``` python
if (data.get('break_category') is None) == (data.get('stage', Round.STAGE_ELIMINATION) == Round.STAGE_ELIMINATION):
```

Is it true that the key won't exist when the user didn't include it in the request, or is it only on creation (and the framework auto-populates for existing objects)? Is the fallback value a hard-coded default, or is it meant to match the default specified in the model, or is it meant to match the existing value in the database (for objects that exist and are being updated)?

tienne-B:
Yes, the key won't exist if not included in the request, and would not be caught beforehand (see BACKEND-3D1). It does not provide a default when creating or updating; not specifying it would still be caught, but at a later stage.

czlee:
Cool, then how did you choose the default values in the `dict.get()` calls?

tienne-B:
* `STAGE_ELIMINATION` to give an error if no break category specified or incorrectly (with the none default there)
* Lists as empty lists so that iteration works
* Default model values

tienne-B:
BACKEND-3KA: Couldn't specify categories for venues as the model/queryset to choose from wasn't specified.

tienne-B:
BACKEND-3KQ: Can't update an instance with nested serializers (Speakers being nested in the Team serializer)

tienne-B:
BACKEND-3KT: Same as BACKEND-3CS but for adjudicators

czlee:
> * `STAGE_ELIMINATION` to give an error if no break category specified or incorrectly (with the none default there)
> * Lists as empty lists so that iteration works
> * Default model values

Okay. I assume you've thought this through, but you just want to be careful with DRY/multiple specification, I think. If it's reverting to the default, then in cases where the default is semantic, it should be referencing the model's default, if it can. I know that's really hard though, so no rush, just wanted to flag it in case it becomes an issue in future.

I got a bit lost with the elimination round thing—I get the effect, but it's preferable if the code can be written to make the intent clearer. At the moment it looks like "rounds are elimination by default", and I couldn't quite tell if you actually meant "if the user doesn't specify, assume the worst".

PR #1522, opened 2020-06-27
Fix #1521, partially revert 326a548

### LINKED ISSUE #1521: Break eligibility checkboxes don't have any effect
In the "break eligibility" view:

Expected behaviour: Clicking a checkbox should initiate a request to the server to toggle the team's break eligibility, and the server should send back a response so the front-end knows it has been saved.

Actual behavior: The checkbox toggles but no request appears to be sent from the browser. The server logs don't show any request when the checkbox is clicked.

Browser console:
image

Reported at https://www.facebook.com/groups/tabbycat.debate/permalink/1577099212458030/

czlee:
Changes the "type" of columns in the break eligbility table to the break category's ID, rather than its name. This changes the value understood by CheckboxTablesContainer.vue:108.

This partially reverts 326a5485da93305e3b1e98e5a22bf4b000e369eb. Since this is @philipbelesky's commit, tagging him for review to ensure this is the intended structure, as opposed to the other way round, changing the ID to slug somewhere else to change the value understood by CheckboxTablesContainer.vue:101:

https://github.com/TabbycatDebate/tabbycat/blob/5606afff0afb9ce3d2c98cc9f3358104e3446ca2/tabbycat/templates/tables/CheckboxTablesContainer.vue#L98-L113

This fixes #1521. If @philipbelesky approves, I'd go ahead and release this hotfix straight away.

### DIFF
--- a/tabbycat/breakqual/views.py
+++ b/tabbycat/breakqual/views.py
@@ -296,7 +296,7 @@ class EditTeamEligibilityView(AdministratorMixin, TournamentMixin, VueTableTempl
                 'checked': True if bc in team.break_categories.all() else False,
                 'sort': True if bc in team.break_categories.all() else False,
                 'id': team.id,
-                'type': bc.slug,
+                'type': bc.id,
             } for team in teams])
 
         # Provide list of members within speaker categories for convenient entry


philipbelesky:
Merged - this fixes the saving issue without causing a regression on the sort behaviour.

PR #1523, opened 2020-06-28
Protect user submission fields from deletion

tienne-B:
Instead of cascade-deleting submission objects (Adjudicator Feedback and Result Ballots) with the deletion of a user, the affected fields (submitter, confirmer) should be set null. These fields are not required as they can be bypassed with public submissions. Users should not normally be deleted, but this vital data should not be deleted with the user.

Places looking for these fields thus accept null values already.

Should `SET_NULL` be used, or `PROTECT`? `PROTECT` would prevent the user from being deleted to begin with, and users should not really be deleted, and rather marked as inactive. Also note that `actionlogentry` also cascades, and could also be altered.


czlee:
> Should `SET_NULL` be used, or `PROTECT`?

Yeah, this is a great question. I think you're right, there's at least a case to be made for using `PROTECT` here, for the exact reason you describe: Maybe forcing tab directors to think twice about deleting users that have already done things, or go through the work of clearing out existing references themselves, is a good idea. Ditto for `actionlogentry`.

@philipbelesky, thoughts?

philipbelesky:
Yea, I'd similarly favour `PROTECT`. It prevents an experienced user from making a mistake with relatively dire consequences. More experienced users should likely understand the relations and the need to clean the other data first.

tienne-B:
Updated the commit

PR #1524, opened 2020-06-29
Rename deprecated Django methods

tienne-B:
Django 3 deprecates a few names, using them as aliases to the new names
- `force_text` to `force_str`
- `ugettext_lazy` to `gettext_lazy`
- `is_safe_url` to `url_has_allowed_host_and_scheme`

### DIFF
--- a/tabbycat/breakqual/base.py
+++ b/tabbycat/breakqual/base.py
@@ -4,7 +4,7 @@ generator (which just takes the top teams)."""
 import logging
 from itertools import groupby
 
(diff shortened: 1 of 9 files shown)


czlee:
Great, this sort of thing with keeping up with deprecations in Django I'm happy to merge in early in our release cycle, similarly to upgrading packages early, to give time for potential problems to arise with any of us before release.

czlee:
(Also, merging without testing, since I'm happy to have deprecation updates tested via general use in develop.)

PR #1525, opened 2020-06-29
Use PostgreSQL's ArrayField where applicable

tienne-B:
See commit messages.

Depends on Django 3.1


czlee:
Oooh, are the `RunSQL` statements manually written? Just thinking through how futureproof they are. Not really a big concern, just wondering.

tienne-B:
Yes, they are manually written. I don't see how this affects future-proofing as before the migration, the format of the table is known in that the separator is fixe.

czlee:
Ah, so the tricky thing with migrations is that they're the only part of the code that has to run on old databases using current code. (Most of the code can be assumed to be run only on matching databases.) That's why they're hard to future-proof, and it's why for example there are very specific ways of retrieving models in them. The specific concern with running SQL code in a migration is that you don't know what version of Postgres the user has, and since they're migration, it's plausible that it might be old.

That said, I wouldn't be _too_ worried about it, for a few reasons. First, cases where it actually strikes are a rare scenario, and I think these things are known to be fragile. Second, there are ways to fix it later, if we need to, since all that matters is that the current code base works with whatever versions of Postgres are currently in fact in use. More something to be aware of for later, than something to fix now. 🙂

czlee (commented):
I'm a fan! I think this makes things a lot cleaner. Sacrificing compatibility with non-PostgreSQL backends is fine imho, I'm pretty sure that's a lost cause already. Just a couple of things:

- There's a migration conflict, easily fixed with a merge migration, which I just added. Can you please check that this is going to be okay? It's fine so long as the migrations in question (after the migration fork) don't get in each other's way.

- The admin (Edit Database) field for `ArrayField` isn't very good. For question choices it renders as a comma-delimited short text field, which probably isn't what we wanted (or if it is, we need to update the help text). For flags, which now have constrained choices (`DRAW_FLAG_DESCRIPTIONS`), it renders as a single dropdown box (the default for text fields with a "choices" field), so you can't input multiple flags.

    Would this library help? https://pypi.org/project/django-better-admin-arrayfield/

tienne-B:
We already sacrificed compatibility with MySQL and other databases with the `notifications` models (`SentMessage` uses JSONField).

The migration is good; I'll look into the widgets and the like in another PR.

tienne-B:
For now, the linked library is only half-useful: gradam/django-better-admin-arrayfield#19 . Also think they're asking too much for just changing a few widgets.

czlee:
Cool, thanks for checking it out! We'll want to figure out something, since the Edit Database area is the only means of editing any of those fields iirc?

PR #1539, opened 2020-07-01
Remove deprecated URL paths

tienne-B:
Were going to be removed in 2017; un-named so unlinked elsewhere.


czlee:
Hehe, this is why we leave comments like this I guess 😄

czlee:
Merging without running, since it's just removal, and my node setup appears to be broken at the moment.

PR #1543, opened 2020-07-02
Rename ranking-display to rank-display

tienne-B:
The former is incorrect, but was used in adjfeedback and style. The class must be used as `rank-display ranking-##`.

Candidate for hotfix.


philipbelesky:
Thanks! I could have sworn I'd made the same fix myself sometime in the last couple of months, but not sure what happened to that commit.

PR #1549, opened 2020-07-05
Replace last instances of 'venue' in messages

tienne-B:
Includes .po


philipbelesky:
Great, thanks for catching these

PR #1557, opened 2020-07-14
Disable contextual alternates for plain URLs

tienne-B:
Inter (font) provides alternates for glyphs when used in various circumstances. However, one of these alternates 'x' (U+0078) with '×' (U+00D7) when between two digits (supposing a multiplication). This is not the case though with URL keys when shown in the URLs in the list.

This commit disables the alternates in this case, and provides a CSS class to use elsewhere if needed (.mixed-text).


philipbelesky:
Interesting! Out of curiosity, did this cause the URL to break? In my test it happily converts back to "x", but I'm not wholly sure of how unicode glyphs are handled in these sort of cases.

tienne-B:
No, didn't cause the URLs to break, nor can it change the character code; it just substitutes the glyph for 'x' with '×'. It's purely aesthetic. Fonts by themselves cannot change characters.

PR #1564, opened 2020-07-24
Correspond max team names lengths to generation

tienne-B:
This commit modifies the Team model to allow for longer team names (short_name and long_name) as their max lengths did not correspond with the maximum possible lengths of the strings generated for those fields.

As such, the max lengths for these fields is shown as the max length of the institution code/name plus a space and the team [short_]reference. It also truncates the full reference to the max size of the short reference if the latter is unavailable for the short_name.

A DataError occurs when the generated names is longer than permitted (but still in the size contraints for the depended fields), so creating/saving affected teams is impossible.

Closes [BACKEND-3NB](https://sentry.io/organizations/tabbycat/issues/1803874449/)
Closes [BACKEND-3NC](https://sentry.io/organizations/tabbycat/issues/1803899403/)


philipbelesky:
Thanks for picking this up. Do we want this in hotfix despite the migration? I'm assuming this mostly affects API users?

tienne-B:
Yes, for a hotfix. This affects everyone, not just API use. Both Sentry issues are not from the API, but from the simple team importer or the database area.

philipbelesky:
Ok, looks good to me. I've also asked CZ to review quickly.

czlee (approved):
Just rebased onto hotfix and made consequential change to merge tree, but the underlying PR looks great, thanks for catching this!

PR #1586, opened 2020-08-14
Add field for participant in Submission

### LINKED ISSUE #683: Public ballot submissions should save which adjudicator's link was used
Kind of funny that we haven't thought of this before now, but at the moment it just saves IP address, given that private URLs are the norm it probably makes sense to take note of which adjudicator's URL was used to submit the ballot.

(Doesn't make so much sense for feedback since that's effectively already done.)

tienne-B:
This commit adds a new field to Submission models as a Person foreign key. This field is then populated when using private URLs as an alternative to IP addresses (although still collected).

With this, ballots can be more easily distinguished from which adjudicator had submitted the ballot. This is added to the "other ballots" list. It can also be used to find which team member submitted feedback, although not shown in any template.

Fixes #683


philipbelesky:
This looks good to me!

czlee (approved):
Great!

Do we also want to modify the subtext in the results entry table? 
image image


Relevant line is line 173 here:
https://github.com/TabbycatDebate/tabbycat/blob/c83a6554cd6995fc0e69a845dd81026e390a555d/tabbycat/results/models.py#L170-L175
but I don't know if there are knock-on effects from changing the serializer, and also it occurs to me that with usernames, participant names and IP addresses all as legitimate things to appear there, we might want to a clean unambiguous visual way to tell between them at a glance.

tienne-B:
Good point, the subtext should also be changed. I don't think we'd need to differentiate between IPs and participants/users as name formats are different. Even between participants and users, usernames are not normally the owner's full name. To distinguish if we have to, we could add a subtle indication or link to their record page.

czlee:
I agree, it'll be obvious most of the time, but I thought some tournaments might use just first names or something (for participant names), so something subtle might be nice 🙂 e.g. a different colour or something, then once we're there we may as well also just distinguish three-way.

tienne-B:
Got a comment a few days ago that keeping track of who submitted (participants) was important in preventing non-chairs from submitting ballots (asked that we keep IPs from checkins to cross-reference)

PR #1616, opened 2020-09-19
Add initial support for rich text in infoslides

### LINKED ISSUE #1106: Motion info-slides should have customizable formatting
It would be nice if images could be inserted in info-slides, as is sometimes done in PowerPoints in tournaments.

Quite easy as a task, as the Summernote widget could be used for the input. However, the display would also need some modifications for this purpose.

### LINKED ISSUE #1517: Summernote boxes don't show
I've had this happen a few times today, not sure if reliably reproducible yet. Only affects develop branch. Master branch works fine. Probably a dependency issue. Making a note so we remember to do something about it before Nebelung.

image

image

Browser shows error:

> Refused to display 'https://hidden-caverns-06472.herokuapp.com/summernote/editor/id_tournament_staff/' in a frame because it set 'X-Frame-Options' to 'deny'.

mitchellpalmer:
Fixes #1106 and also fixes #1517

I guess the question is if the project happy to allow uploading of images. It is possibly a new attack vector, but given this functionality should only be available to authenticated users, it shouldn't be too dangerous. And then, if uploading is to be done, how it should be managed. As far as I can see, there is no precedent in the codebase for accepting uploaded files from users, so I guess its a design decision that needs to be made about where those are to be stored.

I don't believe there is a need to migrate data -- plain text should be handled fine (`SummernoteTextField` is just a wrapper around `models.TextField`)  except I am not as yet sure how plain text with newlines will be rendered.


philipbelesky:
Hey thanks for the PR! I can’t review it right now, but to answer your question: we don’t have any accordance for storing files as Heroku (our main deployment target) doesn’t have an out of the box means of storing user files. 

There might be an addin that could easily enable this.  But if the use case is just for info slides and the home page it might be that uploading elsewhere + embedding is a good enough solution

mitchellpalmer:
Ahh that Heroku limitation makes sense. 

From a cursory search around the internet, it appears the only option which wouldn't either require users to get an S3 account or require an add-on be added to Heroku, is uploading the images directly to the database. There are some drawbacks to that approach - mostly to do with DB bandwidth, which I presume Heroku limits.  Is that something worth investigating? Probably not worth it to be honest.

philipbelesky:
Ah, storing the images as blobs in the database would be a nice way of handling this! The primary limitation on [the free plan](https://elements.heroku.com/addons/heroku-postgresql) is the row count - the total storage space is up to 1GB. So if we could easily set it up to store attachments I don't think capacity would be a concern.

That said, I don't think the issue around attachments needs to be resolved in this PR. If it ends up being non-trivial we could delay it to a future issue or PR.

philipbelesky:
Sorry, the 'mark for review' there was accidental, but I have had a play with things. Quick thoughts:

1. The layout we use in cases of multiple motions aligns images a little akwardly, but I think its fine to let people handle that themselves with linebreaks or CSS.
2. Given the media file uploading will not work OOTB on heroku, we could perhaps set the value of `'disable_upload'` to false using the `heroku.py` config set. That would enable the standard uploading process in local installs or other platforms.

Let me know what you'd like to do to resolve this PR. I think we could:

1. Split off the fix to #1517 into a direct commit (we probably want to release this in a hotfix soonish) and leave this PR as to work on the database-uploaded media option
2. Merge this as-is and create a new issue/PR for the database-uploaded media option

mitchellpalmer:
Yeah I agree with that course of action. Will do!

tienne-B:
Closing this as superseded by #2229, which was merged as 430798f5b14348b0cfe1188f6f66e792f5ad00b5.

PR #1617, opened 2020-09-21
Add round weightings for points

tienne-B:
This commit implements a mechanism for tapered scoring as introduced in [International Debate](https://international-debate.com/2020/03/18/tapered-points/) in March. Here, weights can be added to rounds that act as coefficients to the sum of points team metric. Weights are customizable as model fields, with a migration created to add it.

Weights only apply to the points metric, and not to metrics counting the number of a rank received (number of firsts, wins, etc.)

This PR is based on #1512 to use the new structure for metrics.


philipbelesky:
Does  #1512 need to be merged before this? And did you want to review this @czlee as part of your role in the tapered method? Otherwise I am happy to merge.

tienne-B:
Merging this would also in effect merge #1512.

czlee:
I'll merge it when I look at #1512. Conditional on my code review of #1512, the further changes in this one should be good to go, at least from a technical standpoint.

In terms of need, weighed against complicating our code base… it remains to be seen how much tapered scoring will catch on. But I'm fine with adding this, and then seeing what happens.

@tienne-B Have you checked for N+1 queries as a result of any these changes? It occurs to me that this sort of change would be particularly prone to N+1 inefficiencies. I briefly checked my changes but should do so more thoroughly.

tienne-B:
There are no new N+1 inefficiencies that I can see. The places where we check the round weights already check other Round attributes, and including in querysets do not affect the number of queries. (#1512 reduces the total number of queries!)

czlee:
(just rebased to develop to make the diff clearer now that #1512 is merged)

tienne-B:
Please undo? Rebasing messes up my workflow; merging in `develop` is also cleaner.

PR #1619, opened 2020-09-23
Fix wrong type in annealing without initial

tienne-B:
As the initial allocation is given by an allocator (Hungarian) if None in the method, initial could be a tuple, when it's only expecting a list of AdjudicatorAllocations. This commit ignores the warnings (second element of the allocate function), and also does some optimisations.

Should this allocator be outright deprecated? It was written >10 years ago without much change (other than changing the default initial to basically supersede this allocator).

### DIFF
--- a/tabbycat/adjallocation/allocators/anneal.py
+++ b/tabbycat/adjallocation/allocators/anneal.py
@@ -14,10 +14,7 @@ class SimulatedAnnealingAllocator(BaseAdjudicatorAllocator):
 
     key = "simanneal"
 
-    SCORE_ADJ_TEAM_CONFLICT = 10000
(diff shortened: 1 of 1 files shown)


philipbelesky:
I'll defer the question of redundancy to @czlee, but I can't imagine a future use case that requires it to be maintained.

czlee:
Hello, sorry I'm late getting to this. I think it'd be fair to say we don't care for maintaining it. It's mostly there as a historical relic. I think @qlim might've just been experimenting with different allocation algorithms. Some other tab systems use simulated annealing.

qlim:
Think this emulated the original tabbie allocation algo. Don't think it was ever used at a tournament while I was maintaining! :D

PR #1637, opened 2020-10-03
Add a preference to allow reply speakers without substantives

### LINKED ISSUE #1584: Allow dedicated reply speakers (w/o substantives)
Although this is rare, some World Schools-style tournaments (especially smaller/internal ones) allow the fourths member of a team to act as the reply speaker. This is currently not allowed in Tabbycat (tested in `master`, but the code looks the same in `develop`) as [a check in forms.py](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/results/forms.py#L508) throws a validation error in that case ("The reply speaker for this team did not give a substantive speech").

As I'm tabbing a tournament with this rule in the near future, I have [disabled that check](https://github.com/franga2000/tabbycat/blob/horrible-hack-master/tabbycat/results/forms.py#L508) in my fork and it seems to be working fine.

My suggestion is to add a tournament-level setting to toggle this behavior. I still agree not permitting it should be the default, but I don't see any harm in giving users the option.

I can contribute the feature if you'd accept it since I have plenty of Django experience, but I'm not sure how your licensing works with external contributions. What do you suggest?

franga2000:
Unbelievably late, as usual, but here's the PR I promised like 2 months ago. 

Sorry : )

Closes #1584

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -494,6 +494,15 @@ class ReplyScores(BooleanPreference):
 
 
 @tournament_preferences_registry.register
+class RequireSubstantiveForReply(BooleanPreference):
+    help_text = _("Whether to limit reply speeches to speakers who gave a constructive speech in the debate")
+    verbose_name = _("Require reply speaker to have given a substantive speech")
+    section = debate_rules
+    name = 'require_substantive_for_reply'
+    default = True
+
+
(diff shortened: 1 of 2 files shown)


philipbelesky:
No worries, thanks for adding this. I'll add a note and credit to the CHANGELOG.

PR #1638, opened 2020-10-03
Fix debate flags display error and form field

czlee:
This fixes a couple of issues that probably arose as a result of #1525.

1. Currently, the draw page displays the characters of `str(debate.flags)` on one line each (so "[", "]" for no flags). This commit fixes it so that it just displays the human-readable versions of the flags, restoring the behavior before #1525 was merged.

   Before: , After: 

2. Currently, the form field for flags in the Edit Database area rejects the value whenever one or more fields are selected. This commit modifies the flags fields to use a new `ChoiceArrayField` that is like an `ArrayField`, but handles choices using a `TypedMultipleChoiceField`.

   The ChoiceArrayField code was taken from: https://gist.github.com/danni/f55c4ce19598b2b345ef

   See also related Django issue tracker: https://code.djangoproject.com/ticket/27704

   I couldn't figure out how to (easily) get the form field in the Edit Database area to be taller than 2.28571rem, the height of most fields in django-jet, without an exception made for `select[multiple]`.


czlee:
Just a note that this should showstop Nebelung, or if the `ChoiceArrayField` isn't a great idea then we at least need to fix problem 1 above before releasing.

tienne-B:
This is ugly as using `.get_field_display()` should be in Django 3.1 with django/django#12040 and with the inactive migration (doesn't do anything...)

We should also see about the form field for the choices in Feedback and Answers (multiple-choice array) in the admin.

czlee:
Oh, that's promising, I tried using `.get_field_display()` and it didn't work. Let me try again later today.

Though the other issue was that `ArrayField` doesn't seem to play nice with `choices`. From what I can tell, `choices` needs to be specified in the `base_field`, not the `ArrayField` itself. But then `ArrayField` isn't very good at rendering that, so in Edit Database it shows up as a text field. (If we specify `choices` in `ArrayField`, as we did previously, it renders as a single dropdown coercing to a string, not an array of one string, which then never passes validation.)

tienne-B:
Django got my hopes up... I do think Django should handle representations of the keys used in arrays, but guess this should do for now. The Django bug mentioned this use but seems to have been forgotten when the task was closed.

czlee:
Hmm, yeah, I see there are a couple of false starts in [ticket 24858](https://code.djangoproject.com/ticket/24858) that might've been closer to what we would've wanted. Though even the original description in that ticket isn't quite what we had in mind—it returns a single comma-delimited string, not a list of strings. And I think there's an assumption that `get_foo_display()` should always return a string, so in the context of Django that makes sense, so under that assumption we wouldn't be using `get_foo_display()` anyway since we separate with line breaks. (I guess we could change that to commas, but I feel like the conflicts/flags column is easier to read with a "one thing per line" format.)

I agree support in Django would be more proper, just probably not by overloading `get_foo_display()`. I think ideally we'd be iterating through the `ArrayField` (`debate.flags`), and we'd be able to access some sort of method on each instance of the `base_field` contained in the `ArrayField`. So similar to now, but without that ugly direct reference to `_draw_flag_dict` which in turn just points to `DRAW_FLAG_DESCRIPTIONS`.

(Note that the custom `ChoiceArrayField` only affects the form field, it doesn't have anything to do with `get_foo_display()`. Everything will still work without it, except that the Edit Database area will show it as a simple text field rather than a `SelectMultiple` field.)

Okay to merge @tienne-B? Or do you want to see if there's a way to access `base_field.choices` more directly?

PR #1644, opened 2020-10-13
Set number of preformed panels for outrounds

### LINKED ISSUE #1452: Preformed panels supporting outrounds

tienne-B:
This commit adds cases for the number of debates break rounds should have to anticipated calculations, without adding info for brackets or liveness.

Also clarified `partial_break_round_split` as it returns the number of debates, not the number of teams in the debates.

Continued from #1455 as it wasn't in the 2.4.5 hotfix, so couldn't be merged.

Fixes #1452

### DIFF
--- a/tabbycat/adjallocation/preformed/anticipated.py
+++ b/tabbycat/adjallocation/preformed/anticipated.py
@@ -3,6 +3,7 @@
 import itertools
 
 from breakqual.utils import calculate_live_thresholds, determine_liveness
+from draw.generator.utils import ispow2, partial_break_round_split
 from participants.prefetch import populate_win_counts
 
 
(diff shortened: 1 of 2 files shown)


czlee (approved):
Great!

PR #1645, opened 2020-10-17
Fixing testing emaill BUG

jacklishufan:
Hi
While sending testing emails for SIDO2020, I found some simple errors in the code(which is possibly caused by changes in schema).


tienne-B:
Good catch! I squashed and rebased your commit as 5f49b7937a5620ef00c648481332ee6852aaf5db (and it's merged). Don't know why GH doesn't show it that way.

PR #1647, opened 2020-10-19
Use order_by() for queryset-based standings

tienne-B:
This commit adds an order_by clause to standings based on querysets that orders based on the ranking metrics, and then by the specified tiebreak field (as duplicated from the StandingsInfo-based standings). This makes the database responsible for even more, if able.

Also some code reductions were done.

Depends on django/django#13569 to be fixed. (Merged; will be in Django 3.2 for April 2021)


tienne-B:
Marking as draft as a reminder to not merge before Django 3.2

tabbycat/standings/metrics.py
```diff
@@ -107,10 +107,7 @@ def get_annotated_queryset(self, queryset, round=None):
     def annotate_with_queryset(self, queryset, standings):
         """Annotates items with the given QuerySet."""
         for item in queryset:
-            metric = getattr(item, self.key)
-            if metric is None:
-                metric = 0
-            standings.add_metric(item, self.key, metric)
```
czlee, line 113: Just a heads-up, I think it might have been written like this because `item[self.key]` sometimes exists and is `None`. For example, if the field in question is a `SUM` aggregation but no records match the relevant filter, then I think the SQL field might be populated as `NULL` (as opposed to 0, the sum of one or more zeroes). This can also happen in things like variance, when there's one record (because to compute variance meaningfully you need at least two). PostgreSQL or other behaviour might've changed since then, but this should be checked and noted to be sure. (Also, this really should have been commented with the reason at the time—I think that was me, sorry!)

I tried it quickly, to reproduce just mark four teams as unavailable before generating the round:
image

I think this used to crash the standings generator because Python can't compare `NoneType` to `int`. But that problem presumably disappeared because you moved ranking annotations to SQL, right?

Still wanted to give a heads-up about how it shows though—this might mean there's an inconsistency between "team exists but has no results" ("None") and "can't find the metric" (0).

config/requirements_development.txt
```diff
@@ -12,4 +12,4 @@ pre-commit
 selenium==3.141.*                           # Functional testing (in here for CI tests)
 
 # Debug
-django-debug-toolbar==2.*
+django-debug-toolbar==3.2.1
```
czlee, line 15: I think we normally wildcard the minor version so that we don't have to update this so often, since it's just a development tool 🙂

config/requirements_core.txt
```diff
@@ -1,7 +1,7 @@
 # The necessary dependencies to run Tabbycat
 
 # Django
-Django==3.1.1
+Django==3.2.1
```
czlee, line 4: Huh, Django 3.2.1 doesn't seems to exist for me and it says [expected May 4, 2021](https://docs.djangoproject.com/en/3.2/releases/3.2.1/). Was this meant to be Django 3.2?

czlee (commented):
Just wanted to let you know about a few things I noticed, but feel free to merge before/after dealing with them (though definitely deal with them before the next release).

tabbycat/standings/metrics.py
tienne-B: So for SQL-based rankings (with all the ranking precedence in a single query), nulls are handled (see 7a1b68c94f229998e45e2c91acf65c9163a5eb53) and sorting is included, so Python's sorting is not called on "`None`-able" fields. In the general case where Python's sorting is required, I added a default of 0 in the metric getter in order to handle `None`s. I think it would be more valuable to show that such info is missing than show 0.

Good point for the inconsistency. Changed to give `None`.

PR #1670, opened 2020-11-28
Couple refactors based on CodeClimate suggestions

tienne-B:


tabbycat/adjfeedback/management/commands/generatefeedback.py
```diff
@@ -88,24 +88,19 @@ def handle(self, *args, **options):
...
+    def handle_round(self, round, **options):
+        return self.handle_object(round, 'name', delete_all_feedback_for_round, add_feedback_to_round, **options)
 
-        self.stdout.write(self.style.MIGRATE_HEADING("Generating feedback for debate {}...".format(debate.matchup)))
-        try:
-            add_feedback(debate, **self.feedback_kwargs(options))
-        except ValueError as e:
-            raise CommandError(e)
+    def handle_debate(self, debate, **options):
+        return self.handle_object(debate, 'matchup', delete_feedback, add_feedback, **options)
```
czlee, line 106: While this would clearly be flagged by Code Climate, I'm in two minds about whether this refactor makes the code easier or harder to follow. It's in that region where it's short enough to understand quickly by reading, and the two instances are next to each other, and there's a (nonzero, finite) cost in readability and maintainability to using tools like `getattr()`, so it's a pretty borderline call. Curious as to what your opinion is?

Thinking out loud: An alternative—slightly more verbose, but arguably clearer—would be to pass in a lambda function `lambda x: x.name` or equivalently `attrgetter('name')` (from the `operator` module). I consider this to be highly debatable, though, as in, potentially worse.

Also, you missed generalizing the word "round" in the printed strings (it's "debate" in the other one).

czlee (commented):
Thanks for taking the time to find these opportunities for deduplication! This raises questions that I think are interesting. The `addowninstitutionconflicts.py` I think is cleverly done. The `draw/consumers.py` cases I think is probably sensible. The middle one I wonder whether it's Code Climate being overzealous in applying formulaic rules.

The question for me is whether the code is more maintainable, which is a function of both readability and deduplication. In patterns like this (including ones I've written myself), I feel like there's a trade-off between readability and deduplication. The `draw/consumers.py` and `generatefeedback.py` cases both push the needle more towards deduplication—function with many arguments, most of which serve only as "meta"-like specifications, tend to be harder to read.

I think people can reasonably differ on this, so I'd be curious to hear your thoughts.

tabbycat/adjfeedback/management/commands/generatefeedback.py
tienne-B: I think it's still well readable by following the calls: This method calls these other standard ones but adds some control. I like this `attrgetter()`! Also won't have to remember to modify similar methods. (Think the draw consumer is a tighter call; wouldn't combine the teams method)

I've also removed the model name from the print as it would be clear from the object name.

czlee:
Just noting current checking status:

1. `addowninstitutionconflicts.py` looks good.
2. `draw/consumers.py` had a bug but I fixed it in a6d3e22e2c0f076dfbe207baf45a73544e07c72c.
3. `generatefeedback.py` is not currently fully testable due to a bug on `develop` caused by #1525. I'm working on a fix for that.

czlee:
Okay, aforementioned issues fixed in cac8df11609e086a39461d45ac32fe9be60c2ab7 and 10b7e386d41a01f2a4928097f7f7ddde2529e6c6 and it looks good merged locally.

PR #1690, opened 2020-12-13
Add production Docker configuration

CreatedBySeb:
Given Postgres and performance limits with Heroku, I've started self hosting tabs using Docker. I had to use different configurations to get identical functionality to Heroku, and have cleaned up these tweaks and created a branch with them. I've tried to avoid making intrusive changes, mostly adding files rather than modifying the existing Docker setup. This setup has been used to successfully host two comps now, one on 2.4.7 and one on 2.5.3.


config/requirements_docker.txt
```diff
@@ -0,0 +1,8 @@
+-r requirements_core.txt
+
+# Serving
+honcho==1.0.1                           # Allows a Procfile to run multiple processes
+gunicorn==20.0.4                        # Primary server for wsgi
+
+# Cache
+django-redis==4.12.1                    # Use redis for cache (on heroku; local optional)
```
tienne-B, line 8: If Redis itself is in core, might make more sense to include it there (and remove the comment).

docker-compose.prod.yml
```diff
@@ -0,0 +1,65 @@
...
+      - pgdata:/var/lib/postgresql/data
+
+  redis:
+    image: redis:6
+    volumes:
+     - redis_data:/data
+
+  web:
+    build: .
+    image: tabbycat:2.6.0-dev
```
tienne-B, line 30: `develop` over `master`?

tienne-B (commented):
Haven't tested it out yet, but it looks good. It's also making me think about other settings "divisions", like `DEBUG` being activated in `local.py` rather than `development.py`, and with the branches to use between `master` and `develop`.

I normally squash commits in PRs that are only to correct small errors or style, you may with your second commit.

bin/docker-run-honcho.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+#   Shorthand to migrate and run honcho in docker
+
+cd tabbycat
+
+# Migrate (can't do it during build; no db connnection)
+python ./manage.py migrate --no-input
```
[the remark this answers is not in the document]
CreatedBySeb: This was just me following the pattern of bin/docker-run.sh, in that script it cd's into the tabbycat directory to use manage.py, so I did it the same way for consistency with the development docker script. I can't imagine it makes much difference, but I was just trying to keep them similar.

docker-compose.prod.yml
philipbelesky: Perhaps it would be worth splitting out a `docker-compose.dev.yml`?

philipbelesky:
Hey, thanks for adding this! I originally added the Docker config, but never really ended up using it for either deployment or development. That said, I've always thought it would be highly useful for both. If you knew of a docker hosting service that is as easy to setup as Heroku that would be great to know. If so, we may want to look at promoting that as a new mainstream deployment option given the performance benefits. 

Anyway, this all looks great. Given the setup is now largely identical to that of Heroku, I'd be happy to remove the current `docker-compose` and supersede it with this one. Either way, we should also add this to the CI system (in `.github/workflows`).

tienne-B:
Don't know much about Docker; how does it improve performance?

philipbelesky:
@tienne-B @CreatedBySeb probably knows more, but my assumption is that when on Heroku, the various add-on services (e.g. Heroku/Redis) are a common pool of shared resources. In contrast, a Docker setup maintains independent instances of each service, similar to a virtual private server. A performance gain might also come from a better ability to directly deploy Tabbycat to a dedicated / higher-power server rather than Heroku's shared model and the limits of its free tiers.

tienne-B:
Ah, reading through https://devcenter.heroku.com/articles/how-heroku-works, it seems to indicate that each dyno is a separate pool of resources, so Tabbycat uses 2 by default, so each hour Tabbycat is running and not asleep counts as 2 free hours. The Redis resources are not shared with the web resources, and Postgres is separate to the dyno system.

CreatedBySeb:
> If you knew of a docker hosting service that is as easy to setup as Heroku that would be great to know.

I know you can deploy Docker services relatively easily to both Google Cloud and AWS using manage Kubernetes, but since this costs money to run I imagine most people will probably end up sticking to Heroku, because of its free tier. At the moment I've just been deploying the Docker service on a VPS (1 vCPU, 2GB RAM) which has been performing perfectly well even for reasonably large tournaments and only costs about 3€ a month with Hetzner. I can't imagine this would become a widespread option for users in the near future, because either way you need some more technical know-how and it incurs a base cost.

Regarding performance, there are three main differences:

- Networking: This service configuration uses all local, virtual networking, so has negligible latency, compared to the real dyno and service networking Heroku has.
- Sleep: Heroku free nodes and sometimes services go to sleep if they aren't used for 1 hour, which means they need to cold start when a user first visits them, which can slow down the site, and seems to be especially bad if it wakes to a burst of activity from however Heroku handles holding the requests till it starts (also, sleeping is sometimes more aggressive than the documented 1 hour, so it is possible for dynos to sleep during a round even, when the tab is not in use as much).
- Resources: Heroku free nodes and services use shared resources with no guarantees and often low performance, even my inexpensive VPS runs circles around a Heroku free dyno in terms of performance. This seems like a small thing but actually can make a dramatic difference when a ton of people are hitting the site at the same time, with large API loads or when generating/allocating large draws. (With my documented setup, I've sometimes observed up to 12x faster performance with synthetic API loads compared to Heroku)

In terms of replacing the current docker-compose, I am not sure how it is used, but it seems to provide a very different function. The production docker config I have made is intended to be reliable and performant, so forgoes some of the benefits of the current development focused docker config, such as having access to the app inside the container, live reload etc. Only the tabbycat/settings directory is accessible and persisted (aside from the databases) in the production config. It also disables Debug pages, and adds additional components like an nginx reverse proxy to mimic Heroku. If the current Docker configuration isn't actually used this is all probably fine, but if it is it probably makes sense to have a separate production orientated config too, given the changes.

PR #1692, opened 2020-12-15
Mention the Tabbycat Debate Association in the footer

tienne-B:
In the place of the printed ballots, as I'd think the usage would be quite low (especially with the printable Tabbycat ballots).

The new message will link to the association's website.


philipbelesky:
Happy to bump the mention of the printed ballot site. One proposed wording tweak: "...for **advancing** open debate technology"

czlee:
:+1:

PR #1696, opened 2020-12-16
Use inheritance to share Docker configs (and add redis to the dev setup)

philipbelesky:
This is still a little bit a work in progress, but it should:

- Use inheritance to reduce duplication between the default and production docker configs
- Use redis in the default docker config


philipbelesky:
Merged in as f9b41472b6d2bb8cba24674f2b28303b92ab177e

PR #1698, opened 2020-12-20
Update docs for array fields to say choices are comma-delimited

czlee:
Relates to #1525 and #1638
Discussion on Facebook: https://www.facebook.com/groups/tabbycat.debate/permalink/1747279095440040/

Update documentation and help text to say that the list is comma-delimited. The underlying field is actually an `ArrayField` (#1525) but we don't have a good widget for the `AdjudicatorFeedbackQuestion.choices` field because the user can specify anything in it (unlike for draw flags). But the widget for it is currently just a text box with comma-delimited list, and we haven't found a better solution yet.

One important thing I'm not sure about: Does this affect the API? As in, does the same help text appear in API docs, and would it be incorrect in that context?


tienne-B:
No, doesn't affect the API:
https://github.com/TabbycatDebate/tabbycat/blob/24d2aa9c3fad3bbf92718fe318b2ff63aa2cf0e6/docs/api-schema.yml#L3104-L3108

tienne-B:
As it changes a model field, a migration is expected.

czlee:
> The choices are already an array rather than a delimited list.

Does the help text go into any API documentation? i.e., will API users be thrown off by help text suggesting it's a comma-delimited string?

(If it does, then this "comma-delimited" hint needs to be inserted in an admin form field override, not the help text of the model field. Which is a little complicated but something we know the pattern for, so easy enough… just checking which is preferable.)

tienne-B:
I ignore the `help_text`s found with the `OPTIONS` verb and point towards the schema file, which is not automatically updated.

tienne-B (approved):

PR #1702, opened 2020-12-27
Use django-better-admin-arrayfield for adjudicator feedback

### LINKED ISSUE #1701: Find better widget for AdjudicatorFeedbackQuestion.choices
Currently the choices box just renders as a text field with comma-delimited entries. Among other things, this prevents users from entering any choice with a comma in it, which probably isn't that uncommon. (That's why the previous separator was "//".) We'd want some better widget to handle this field.

In #1525 we discussed this library: 
- https://pypi.org/project/django-better-admin-arrayfield/
- https://github.com/gradam/django-better-admin-arrayfield

But it doesn't support choice fields, so can't be used for e.g. draw flag descriptions. We dealt with that in #1638 instead. But the solution in #1638 doesn't help `AdjudicatorFeedbackQuestion.choices`, which doesn't have choices of its own, and django-better-admin-arrayfield might work better for that instead.

Relates to #1525, #1638, #1698.

czlee:
Closes #1701.

Uses https://pypi.org/project/django-better-admin-arrayfield/ for the `ArrayField` in `AdjudicatorFeedback.choices` and `AdjudicatorFeedbackManyAnswer.answer`.

(Leaves `Debate.flags` and `DebateTeam.flags` untouched.)

Good:
- Does what it needs to do.
- No need to mess around with special characters reserved for use as separators.
- Doesn't have to be applied for all fields, so the draw flag description fields still use the `ChoiceArrayField` introduced in #1638.

Bad:
- The "delete" button for each field doesn't show up. I presume this is a django-jet compatibility thing that we can't do anything about, since django-jet is no longer officially maintained. **However**, if you just blank out the field, it appears to interpret it as a deletion. So there is still a way to delete entries.
- The "Add another" button shows in a funny place.

Before:
image

After:
image


tienne-B:
Looks good, and I do not see your points under "bad;" your CSS might just have been cached? I see the delete and "Add More" buttons in the right places (and stylistically correct)

The `flags` fields do not need to be changed, as they already use a multiple select widget.

PR #1705, opened 2020-12-27
Add warning about multiple confirmed ballots

czlee:
This adds a warning for the situation where a debate has two or more confirmed ballots, as happened at USUDC 2020.

This should never happen in normal use, and we had never heard of a case before USUDC. The `save()` method of `Submission` (the parent class of `BallotSubmission` and `AdjudicatorFeedback`) has a fairly elaborate duplicate prevention mechanism:
https://github.com/TabbycatDebate/tabbycat/blob/04541b006eda88e7a21abd9cf386829ca213b5e7/tabbycat/results/models.py#L72-L92

It was probably caused by a third-party bot directly modifying the Tabbycat database via SQL. The warning is therefore just watching out for adventurous users, to provide some hope of detecting it before the tab is released. As such, it doesn't provide any further information on narrowing down where the issue is—unlike the trainee scoresheets check in #922, it doesn't check how many debates are affected, nor list what they are. The warning message tells the user that it's probably due to direct SQL queries, and the user is pretty much on their own to figure out what to do with it.

One could probably reasonably argue about whether this check is worth putting in in the first place, but I tend to take the view that if it can hit us without us realising it, it's worth checking programmatically for the rare occasion where it saves a major headache.

### DIFF
--- a/tabbycat/results/templates/admin_results.html
+++ b/tabbycat/results/templates/admin_results.html
@@ -101,6 +101,25 @@
     {% include "components/alert.html" with type="danger" icon="alert-octagon" %}
   {% endif %}
 
+  {% comment %}
+    Another warning about potential database inconsistency, but this one's only
+    ever happened as a consequence of direct database edits via SQL, so we've
+    probably already got all the consistency checks we can reasonably have. This
+    is just an extra courtesy for users who try to hack too much.
+  {% endcomment %}
+  {% if debates_with_multiple_confirmed_ballots_found %}
+    {% blocktrans trimmed asvar message %}
+      One or more debates in this round appear to have more than one confirmed
+      ballot submission. This can cause unexpected results (or crashes), because
+      Tabbycat assumes that each debate has at most one confirmed ballot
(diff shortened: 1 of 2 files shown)


philipbelesky:
@czlee is this still a draft? Or was it waiting for a review?

czlee:
Awaiting review, I think 🙂

tienne-B:
Not a big fan of adding such a warning as this is really only possible if you're inputting raw SQL commands, which is already not recommended and is known to be dangerous for precisely this reason. Also, even with this warning it would be hard to know how to resolve manually.

czlee:
I think that's a reasonable way to look at it. @philipbelesky, thoughts?

philipbelesky:
I'm in favour of the argument in the initial post: if there is a database state that is easily overlooked, but potentially compromises the tab, there should definitely be some sort of warning about it. People seeing that warning may not be clear about how to resolve it, but they should at least be able to escalate it to someone who can. I'd err on the side of adding it unless there is some sort of non-minor performance penalty to the check.

Issue #1699, opened 2020-12-25
Deprecate anorak/boots importers in favour of API


tienne-B:
Tabbycat's importers from CSV files are very hard to use for custom tournaments, requiring database pushes on Heroku (with CLI), explained in many sections of many pages of the documentation, including for coding custom importers. I can't imagine custom importers are created/used much, and I find the code a bit difficult to grasp.

The `importtournament` importers could be spun-off to a new project, converting the CSV format to use the Tabbycat API, similar to @tasdidtahsin's [TabbycatImporter](https://github.com/tasdidtahsin/TabbycatImporter) project. It would allow data to be more easily exported live.

The `importtournament` command would have to be removed though, and an alternative format found for the demo tournaments to be effectively loaded. Fixtures would not be usable as they rely on primary keys, and complexity must be kept low.

czlee:
I support this move in principle.

It wouldn't be offensive to me for the demo tournaments to be imported using dedicated code that is not used more generically. The current mechanism for loading demo tournaments in the GUI (by which the `importtournament` command is called programmatically using `management.call_command()`) is a bit of a hack. Of course, it is helpful for development to be able to set up demo data quickly from the command line, so we probably want _some_ notion of this without plug-ins, even if it can't be used for anything other than the demo data.

(Of course, there's also nothing wrong with making the code that loads the demo data also capable of loading user files, I'm just saying I no longer think it needs to if that would be a pain. And it probably would be—the existing importer code has a _lot_ of confusing error-checking going on to be flexible with what it accepts, and could probably be simplified greatly if it knew demo data was guaranteed to be very clean.)

tienne-B:
I would be inclined to have DebateXML files for the demo tournaments, and use #1076's importer with the files.

czlee:
Sounds good. You've probably already thought of this, but we'll want to plaster the `importtournament` command with deprecation warnings and keep it around for at least a couple of versions after deprecation, so that users who tab a lot and have their own favourite personal processes have ample warning about the need to migrate their processes over. Our user base seems to be large enough now that we can't assume people aren't relying on it just because it's an "advanced" feature.

PR #1707, opened 2020-12-31
Deprecate importtournament command

tienne-B:
For #1699; a new repo [TabbycatDebate/CSV-Importers](https://github.com/TabbycatDebate/CSV-Importers) will have the importers.


czlee:
What's the current status and plans for https://github.com/TabbycatDebate/CSV-Importers? We should start deprecating formally once the alternative is up and running with instructions (which will be different, because at a minimum it involves checking out a new repo). No rush, just wondering.

teymour-aldridge:
Now that https://github.com/teymour-aldridge/tabbycat-import/ exists, I guess this might be worth looking at again @tienne-B

PR #1717, opened 2021-01-06
Associate more motions to ballots

tienne-B:
Currently, motions are only associated with ballot submissions directly
if the 'enable_motions' preference is enabled. This creates two paths to
get motions from debates, and ignores edge-cases for special motions but
without selection. This commit adds the motion foreign key to all
ballots regardless of the preference, added automatically if there is
only one option.

A migration is there to populate the field, but is conservative to not
prefer a motion over another and to only update where motions are null.

### DIFF
--- a/tabbycat/participants/tables.py
+++ b/tabbycat/participants/tables.py
@@ -55,7 +55,6 @@ class AdjudicatorDebateTable:
             Prefetch('debate__debateadjudicator_set',
                 queryset=DebateAdjudicator.objects.select_related('adjudicator__institution')),
             'debate__debateteam_set__team__speaker_set',
-            'debate__round__motion_set',
         )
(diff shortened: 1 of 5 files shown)


czlee (approved):
Looks good to me! Don't forget to include something explaining the change (like, in some detail) in the changelog, this change in behaviour might throw off some advanced users.

tienne-B:
This has actually been merged in ef39c12d995a6e8be2aa9c0e1b9f5b4132508e9b

czlee:
Wait, how did that happen? Did #1752 include ef39c12 by accident?

tienne-B:
No, did an octopus merge on both, but pushed to my fork and deleted the branch before pushing it here. As such, the branch was deleted without merge from the perspective here.

Issue #473, opened 2017-07-04
If we only expect FeedbackFromTeams on the orallist, multiple submissions should unconfirmed each other


philipbelesky:
I assume it is a relatively rare issue that a team would (through malice, or incompetence) submit on all of their panellists and also the chair. However if we expect feedback only on the oralist, whenever a submission comes through for an adjudicator from a team it should probably unconfirm their previous feedbacks for adjs on that same panel. 

This both removes those 'bad' pieces of feedback from the feedback scores, but also ensures that the team record pages and expected feedback pages will give a better guide of what is actually going on.

czlee:
I see, so if they submit on the (not-rolled) chair and then a panellist, only the latter is confirmed causing it to look like there's no submission on the chair and an unexpected submission on the panellist? (This is the effect it would have.)

philipbelesky:
Yea, that's exactly the result I'd like. The thinking here is that it would hopefully solve two problems at once: the need to 'clean' unexpected feedback from the database and signal to teams that they are doing something wrong. Even if they don't realise what they are doing wrong, hopefully the tracking page at least prompts them to resubmit feedback on the chair one more time and in doing so unconfirmed their previous attempts.

czlee:
Cool. This actually interferes with the current mechanism a little; we might have to implement it at a different layer. Will think on it.

PR #1718, opened 2021-01-06
Decouple adj feedback progress with statistics

tienne-B:
The pages listing participants with the number of feedback to submit does not require loading all the progress objects; just aggregated values. With this, less memory is used as well as fewer database queries.

### DIFF
--- a/tabbycat/adjfeedback/progress.py
+++ b/tabbycat/adjfeedback/progress.py
@@ -12,9 +12,10 @@ adjudicator).
 import logging
 from operator import attrgetter
 
+from django.db.models import Count, F, Q
+
 from adjallocation.models import DebateAdjudicator
 from adjfeedback.models import AdjudicatorFeedback
-from draw.models import DebateTeam
 from results.prefetch import populate_confirmed_ballots
 from tournaments.models import Round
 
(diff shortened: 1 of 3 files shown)


czlee:
I quickly ran this on my test database and it seems to produce negative numbers:
image

It also changed the subtitle at the top of the feedback progress page from
> 624 missing feedback submissions (55.5% returned)
to 
> 330 missing feedback submissions (79.2% returned)

tienne-B:
Negative numbers can happen when there is unexpected feedback, and so it would be good to check those cases. Such extraneous feedback should be automatically unconfirmed however: #473.

czlee:
I checked that, and while it's true that I have a lot of unexpected feedback in my database, the correlation doesn't seem to be one-to-one with the numbers shown.

Though the automatic unconfirmation only happens on save, right? What happens if a user changes the "feedback expected from" setting with feedback already in the system?

czlee:
(I'll dig further into this to try to figure out what's going on some time in the next couple of days, I switched to working on the emails thing for a bit.)

tienne-B:
> Though the automatic unconfirmation only happens on save, right?

That's correct, so existing feedback would not be unconfirmed when changing the setting, and both changing the settings mid-way and unconfirming based on that should be avoided. Maybe have a warning to that effect when changing? (Let me avoid subqueries!)

czlee:
Haha, sure, okay. Or maybe having some sort of function (even if it's just a command-line dj management command) to do that en masse, for the cases where a user really needs to change that setting, might help.

I checked again and the mismatches definitely include cases where there is no unexpected feedback. So, umm, will see what I can find…

czlee:
Okay, I think I got it. The old version filtered for `debate__ballotsubmission__confirmed=True` when counting expected submissions. The new version doesn't, so it tags the current round (before there is a result for it) as "expected".

I don't remember exactly why it was written that way, but I guess I can imagine that if multiple rounds are drawn in advance, you don't want the participant to be shamed until the round has actually happened. Or, even when the draw is rolling, or hasn't yet been released, you don't want to count that round as part of the expected feedback.

About the unexpected feedback thing—happy in principle to use user-based strategies to avoid subqueries but just wondering, could this just be done as additional filters on the `num_submitted` annotation? Like, basically you apply all the same filters as for `num_expected` to the `num_submitted` annotation as well, so that the `num_submitted` records are a subset of the `num_expected` records? Or does that break down somewhere?

czlee:
(On another note, need not be done in this PR but it occurred to me randomly while looking at this that it'd be nice to do it so that the secondary sort order on the table is alphabetically by displayed name. I think doing this by just sorting in the queryset will be good enough, no need to carry through the Vue table, though I think it's admittedly nontrivial because of code names.)

czlee:
This PR is a little hard to verify, sorry, so in order to inspect it I need to set aside quite a lot of time that I don't currently have, and won't have until September. It's hard to verify because of the change in behaviour. With most optimizations, to run a sanity check, I would basically just take my existing working database, run the new code with it, and check that relevant pages look the same. In this case, because of the change in treatment of unexpected feedback, there's no quick way for me to verify that it doesn't bring about unexpected changes in behaviour.

One way to speed this up is to write the optimization to behave _exactly_ the same as the current behaviour—then the checks are straightforward.

That said, I'm aware that we've already discussed that in this case, some change in edge case behaviour may be an okay sacrifice for performance. But without an explanation I kind of have to assume every change is a breaking change in some way, which is why it takes time to either satisfy myself that it's not breaking, or figure out precisely why it is breaking. So in this case, what would be helpful is:
- a description of precisely what behaviour is expected to change (and to stay the same)
    - this should include things that might seem inconsequential at first, _e.g._ sort order
    - if possible, this should describe _how_ it changes, so that I (or a reviewer) can verify at sight if the new behaviour is as expected
- why it would be too unwieldy to refactor without changing this behaviour
- some thinking on how this might affect tournaments (or why it won't) and any guidance that might need to be given to users (if any)

PR #1719, opened 2021-01-08
Fix feedback graph hovers with undefined

tienne-B:
The lambdas that create the popovers in the adj feedback graphs had the wrong order of parameters, making all variables show as undefined. The positioning was also off as the location of the dot was relative to the SVG. The text is now within the SVG, removing the CSS though.

Should the text even be kept? Un-i18n'd and counting the location of the hovered dot isn't complicated.


philipbelesky:
Hmm, on `develop` the tooltips work perfectly fine for me, whereas they do not show on your branch. Could you try cleaning out `node_modules` and reinstalling? This might be related to a bump in d3 that changed some of the call methods.

PS: the dots exist partly to explain/reinforce the color scheme and I still think the position of the dots is quite hard to count out - particularly if using a larger score scale.

Screen Shot 2021-01-09 at 12 32 34 pm

PR #1720, opened 2021-01-12
Associate more Venues[Categories] to tournaments

tienne-B:
This commit makes sure to assign tournaments to VenueCategories in the
formset.

It also works on existing Venues and VenueCategories in a migration,
which assigns tournaments to venues with debates associated as well as
VenueCategories with venues associated, splitting them if they are
linked to many tournaments implictly.

For good measure, a second round of finding venues' tournaments is
conducted through VenueCategories.

### DIFF
--- a/tabbycat/venues/views.py
+++ b/tabbycat/venues/views.py
@@ -78,6 +78,7 @@ class VenueCategoriesView(LogActionMixin, AdministratorMixin, TournamentMixin, M
         if self.instances:
             for category in self.instances:
                 category.tournament = self.tournament
+                category.save()
 
             message = ngettext("Saved room category: %(list)s",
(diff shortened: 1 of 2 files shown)


tabbycat/venues/views.py
```diff
@@ -78,6 +78,7 @@ def formset_valid(self, formset):
         if self.instances:
             for category in self.instances:
                 category.tournament = self.tournament
+                category.save()
```
czlee, line 81: Just curious, what was the behaviour before this change? Did this omission effectively render the line above it a no-op, or did it do stuff intermittently?

tienne-B: No-op if I remember.

PR #1723, opened 2021-01-24
Fix room constrainee ID dropdown

### LINKED ISSUE #1722: Room constrainee ID dropdown doesn't list options
This field is meant to show the constrainees that match the constrainee type (i.e., `subject_content_type`), but it doesn't.
image

Thanks to Will Reilley Silberman for bringing this to my attention.

## Diagnosis work so far
- The data list is present and correct, but all of the options in the data list appear always to be disabled, even when you change the constrainee type. Disabled options aren't included in the dropdown, so this means the dropdown is always empty. https://github.com/TabbycatDebate/tabbycat/blob/a72e905a449048c71d97625a4d560579bf80ff7f/tabbycat/venues/templates/venue_constraints_edit.html#L32-L42
- The JavaScript in venue_constraints_edit.html that modifies the data list filter does so by checking if the name of the type (e.g., "team") is case-insensitively contained within the option label (e.g., "Bree HC (Team)").
- However, currently, the content type label is "participants | team" (etc.) rather than just "team". This means that the type label is never contained within constrainee label, so no constrainee passes the filter for any type.
- This appears to be a change in Django that first appeared in Django 3.0: https://github.com/django/django/commit/48c17807a99f7a4341c74db19e16a37b010827c2, https://github.com/django/django/pull/10776. The change is sensible, but unfortunately it breaks our assumptions.
- We first upgraded to Django 3.0 in Nebelung (2.5): https://github.com/TabbycatDebate/tabbycat/commit/becabef7a077e855acd0540a92ed718f76b7a521. So presumably issue is new in that version, which we released just a couple of months ago.

## Unsuccessful fix efforts so far
All we want to do is revert the content type labels to their old behaviour, which is just the `ContentType.name` field.
- Specifying `choices` in the `Select()` constructor in venues/views.py doesn't seem to help, the choices seem to be overridden by the queryset for `ContentType`, regardless of whether `ForeignKey.limit_choices_to` is specified. https://github.com/TabbycatDebate/tabbycat/blob/a72e905a449048c71d97625a4d560579bf80ff7f/tabbycat/venues/views.py#L116-L119
- Is there a method we can override to specify how to get option labels, maybe, so that it'll take `content_type.name` rather than `str(content_type)`?

czlee:
Fixes #1722.

@tienne-B can you think of any cleaner ways to do this?

Also, I wanted to add a simple regression test just to check the label of the field, so that if a package change bites us like this again, we'll know straight away rather than only when it's reported by a user. But I couldn't find any off-the-shelf tests suitable for [checking the label of a field](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.SimpleTestCase.assertFieldOutput), and it's complicated by how the formset and form in question aren't classes in their own right; they're constructed by factories in the view. We could check the HTML output directly, but that's messy, and the `value` attribute of the `<option>` element is database-dependent (since it's the ID of the content type). I guess this is mostly a frontend function (the labels aren't, but the dropdown filtering is), so it's not really cut out for testing by the Django testing tools. The other idea would be to include a unit test for the JavaScript in venues/templates/venue_constraints_edit.html, @philipbelesky is there any (easy) scope for this?

(If adding a test is too hard then we should just merge without it, I just wanted to get takes on the idea first, since this seems like the kind of thing regression testing is for.)

### DIFF
--- a/tabbycat/utils/widgets.py
+++ b/tabbycat/utils/widgets.py
@@ -1,9 +1,0 @@
-from django.forms import TextInput
-
-
-class SelectPrepopulated(TextInput):
-    template_name = 'select_prepopulated_widget.html'
-
-    def __init__(self, data_list, *args, **kwargs):
-        super(SelectPrepopulated, self).__init__(*args, **kwargs)
-        self.data_list = data_list
--- a/tabbycat/utils/fields.py
+++ b/tabbycat/utils/fields.py
@@ -1,5 +1,6 @@
 from django import forms
 from django.contrib.postgres.fields import ArrayField
+from django.db.models import ForeignKey
 
 
 class ChoiceArrayField(ArrayField):
@@ -18,3 +19,19 @@ class ChoiceArrayField(ArrayField):
         # Skip our parent's formfield implementation completely as we don't
         # care for it.
         return super(ArrayField, self).formfield(**defaults)
+
+
+class LabelByNameModelChoiceField(forms.ModelChoiceField):
+    """ModelChoiceField that uses `obj.name` rather than `str(obj)` for labels."""
+
+    def label_from_instance(self, obj):
+        return obj.name
+
+
+class LabelByNameForeignKey(ForeignKey):
+    """ForeignKey that uses `obj.name` rather than `str(obj)` for labels."""
+
+    def formfield(self, **kwargs):
+        defaults = {'form_class': LabelByNameModelChoiceField}
+        defaults.update(kwargs)
+        return super().formfield(**defaults)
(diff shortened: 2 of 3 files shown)


tabbycat/utils/widgets.py
```diff
@@ -1,9 +0,0 @@
-from django.forms import TextInput
-
-
-class SelectPrepopulated(TextInput):
-    template_name = 'select_prepopulated_widget.html'
-
-    def __init__(self, data_list, *args, **kwargs):
-        super(SelectPrepopulated, self).__init__(*args, **kwargs)
-        self.data_list = data_list
```
czlee, line 9: I don't know why this file was still here lol

tienne-B:
Merged in 186c81b518f0a5848e44481d38812d8e2f4ee84a

PR #1752, opened 2021-03-08
Revise email docs and update email configuration

### LINKED ISSUE #1751: Update guidance for email configuration
Creating an issue to track much-discussed updates relating to how emails are configured. This issue is about configuring an email server for Tabbycat. It is *not* about the notifications framework within Tabbycat, though where configuring via Tabbycat settings is a substitute for configuring via environment variables, it may touch on it.

See also: [Posts marked #emails in the Facebook group](https://www.facebook.com/hashtag/emails/).

Current state of discussion:
1. **Automatically provisioning SendGrid is no longer viable.** Two main problems:
    - SendGrid auto-blocks all automatically created accounts (e.g. created via the Heroku add-on), and the only known way to get them unblocked is to contact SendGrid support and wait for a response. While they respond within 24-48 hours, this also sort of renders the concept of automatic provisioning moot.
    - SendGrid spams you _every four minutes_ from the time the account is created to tell you that the brand new account's activity is "in violation of one or more of our policies", despite its having done absolutely nothing. No, really, _every four minutes._ I don't know if SendGrid is aware of the irony in spamming a user who is suspected of spamming despite having sent no emails.
      image
2. **A third-party SendGrid account still works.** That is: User can create their own SendGrid account, possibly via the GitHub Education Pack if they're eligible. Then they configure Heroku to use that SendGrid account using config vars, _without the SendGrid add-on_, as suggested in [current docs](https://tabbycat.readthedocs.io/en/v2.5.7/install/heroku.html#sendgrid-account-details).
3. **No known automatically provisioned alternative via Heroku.** We're not currently aware of any Heroku-based solution that allows us to recreate the seamlessness we had when we introduced SendGrid emails back in 2016.
4. **Several known third-party solutions.**
    - [Calico](https://www.calicotab.com/) (run by @tienne-B) provides the email integration itself, so Calico users are unaffected by this issue.
    - Users can configure (e.g.) a Gmail account directly, so long as their use doesn't fall foul of Gmail's policies.
    - Can use third-party mail merge solutions, see Facebook post above for products users have reported using.

Current plan:
- [ ] Remove SendGrid from automatic provision.
- [ ] Update docs to advise more clearly through personal SendGrid account instructions.
- [ ] Update docs to reference other third-party solutions.
- [x] Investigate [other email apps on Heroku Elements](https://elements.heroku.com/addons/categories/email-sms).

czlee:
Will close #1751, once it's ready.

Notable changes:
- Replaced config vars `SENDGRID_USERNAME` and `SENDGRID_PASSWORD` with just `SENDGRID_API_KEY`—this is more aligned with how personal SendGrid accounts should be used, and since we're dropping automatic provisioning of the SendGrid add-on, there's no need for us to follow their convention. Plan is to check for the old setting and for it still to function, but to show a warning on the tournament overview page.

More to come—may add more notes on other options to docs. Thoughts welcome while still being drafted.

### DIFF
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -121,7 +121,10 @@ elif environ.get('SENDGRID_API_KEY', ''):
     EMAIL_USE_TLS = True
 
 elif environ.get('SENDGRID_USERNAME', ''):
-    # these settings are deprecated as of Tabbycat 2.6.0 (Ocicat)
+    # These settings are deprecated as of Tabbycat 2.6.0 (Ocicat).
+    # When removing, also remove utils.mixins.WarnAboutLegacySendgridConfigVarsMixin and
+    # templates/errors/legacy_sendgrid_warning.html (and references thereto).
+    USING_LEGACY_SENDGRID_CONFIG_VARS = True
     SERVER_EMAIL = environ['SENDGRID_USERNAME']
     DEFAULT_FROM_EMAIL = environ.get('DEFAULT_FROM_EMAIL', environ['SENDGRID_USERNAME'])
     EMAIL_HOST = 'smtp.sendgrid.net'
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -109,7 +109,7 @@ if environ.get('EMAIL_HOST', ''):
     EMAIL_HOST_USER = environ['EMAIL_HOST_USER']
     EMAIL_HOST_PASSWORD = environ['EMAIL_HOST_PASSWORD']
     EMAIL_PORT = int(environ.get('EMAIL_PORT', 587))
-    EMAIL_USE_TLS = environ.get('EMAIL_USE_TLS', '').lower() == 'true'
+    EMAIL_USE_TLS = environ.get('EMAIL_USE_TLS', 'true').lower() == 'true'
 
 elif environ.get('SENDGRID_API_KEY', ''):
     SERVER_EMAIL = environ.get('DEFAULT_FROM_EMAIL', 'root@localhost')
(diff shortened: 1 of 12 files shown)


tabbycat/settings/heroku.py
```diff
@@ -109,8 +109,19 @@
     EMAIL_HOST_USER = environ['EMAIL_HOST_USER']
     EMAIL_HOST_PASSWORD = environ['EMAIL_HOST_PASSWORD']
     EMAIL_PORT = int(environ.get('EMAIL_PORT', 587))
-    EMAIL_USE_TLS = bool(environ.get('EMAIL_USE_TLS', True))
+    EMAIL_USE_TLS = environ.get('EMAIL_USE_TLS', '').lower() == 'true'
```
tienne-B, line 112: Should default to true

czlee: welp nice catch

czlee:
Quick note while I remember:
- [ ] This PR should add a deprecation notice that is displayed to users who are still using `SENDGRID_USERNAME` and `SENDGRID_PASSWORD`, since support for that will eventually be removed, and it is indicative of users who are using the Heroku add-on.

czlee:
Random thought: I think we can control the from email address from source code (as opposed to from `DEFAULT_FROM_EMAIL`), so it's probably not a bad idea to make this a proper Tabbycat setting and run some more intelligent checks/error messages when sending fails due to Sender Identity Verification or something like that.

tienne-B:
> I think we can control the from email address from source code

Yes we can.

czlee:
Current status:
- This removes the auto-provision SendGrid, and changes the expected config var to `SENDGRID_API_KEY`, as indicated earlier.
- It shows a deprecation message if it detects that `SENDGRID_USERNAME` is in use.
- Docs are updated accordingly.
- It makes an effort to catch the SendGrid Sender Identity error.

Caveats:
- I don't know whether the Sender Identity error catch actually works, and I'm not sure how to test it because short of creating a new account I'm not sure how to generate the error. I _think_ it's probably an `SMTPResponseException` but I couldn't say for sure. I've added `logger.warning(..., exc_info=True)` calls throughout the code catching test email errors to try to give us better visibility.
- I haven't created the "from email" preference, because I realized as I was doing it that I wasn't sure whether to make it a site-wide preference or a tournament preference. My first instinct was tournament preference, but that would break the test email facility, which is a site-wide page. I'm inclined to leave this for a future PR/commit unless other devs think the answer here is obvious.

PR #1769, opened 2021-04-18
Add preference to exclude ghosts from team scores

tienne-B:
As pointed out recently, the WUDC and EUDC rules show that team scores must not include scores from iron-men (except for their highest). This is 34.9.3.2 in WUDC. The current behaviour was to keep ghost scores, and is expected now. As such, a new preference is added to toggle between these calculations, but is set correctly in the preset.

### DIFF
--- a/tabbycat/results/result.py
+++ b/tabbycat/results/result.py
@@ -888,6 +888,8 @@ class ConsensusDebateResultWithScores(DebateResultWithScoresMixin, ConsensusDeba
     get_score = speakerscore_field_score
 
     def teamscore_field_score(self, side):
+        if not self.tournament.pref('teamscore_includes_ghosts'):
+            return sum(self.get_score(side, pos) for pos in self.positions if not self.get_ghost(side, pos))
         return self.scoresheet.get_total(side)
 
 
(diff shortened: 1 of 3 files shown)


tabbycat/results/result.py
```diff
@@ -944,13 +946,18 @@ def teamscorebyadj_field_margin(self, adj, side):
     def teamscorebyadj_field_score(self, adj, side):
         return self.scoresheets[adj].get_total(side)
 
+    def _teamscore_score_component(self, adj, side):
+        if self.tournament.pref('teamscore_includes_ghosts'):
```
czlee, line 950: Australs results seem flipped, is this line missing a `not`?

czlee (commented):
It's a bit weird now that the totalling is done in `Scoresheet` normally, but done in `ConsensusDebateResultWithScores` or `DebateResultByAdjudicatorWithScores` if ghosts are excluded. But I thought about it for quite a while and I don't have an obviously better solution 🤷🏻‍♂️

Looks good, just check up on the reversal in line 950?

tabbycat/results/result.py
tienne-B: Fixed. Prefer avoiding a "if not, else" construction, so swapped the returns.

czlee:
We should probably add unit tests for this kind of thing, lol

PR #1811, opened 2021-06-08
Move user-related templates to users app

tienne-B:


philipbelesky:
Is this finished? The templates whose paths have changed now seem to redirect to their defaults.

PR #1828, opened 2021-06-19
Generalize the controlled access mixin

tienne-B:
As it is also used with global preferences, thus without a tournament.

The tournament argument for `is_page_enabled` is also un-needed as the
object's tournament property is available there.

### DIFF
--- a/tabbycat/adjfeedback/views.py
+++ b/tabbycat/adjfeedback/views.py
@@ -431,8 +431,8 @@ class PublicAddFeedbackIndexView(PublicTournamentPageMixin, BaseAddFeedbackIndex
     template_name = 'public_add_feedback.html'
     tabroom = False
 
-    def is_page_enabled(self, tournament):
-        return tournament.pref('participant_feedback') == 'public'
+    def is_page_enabled(self):
+        return self.tournament.pref('participant_feedback') == 'public'
 
     def get_from_adj_link(self, team):
         return reverse_tournament('adjfeedback-public-add-from-adjudicator-pk',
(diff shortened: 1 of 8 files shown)


czlee:
Huh, I wonder why I had a `tournament` argument in `is_page_enabled` originally 😅

czlee (changes requested):
Currently in this PR, going to the account sign-up page with an incorrect key yields a 500 with

> Exception Type: AttributeError at /accounts/signup/sad/
> Exception Value: 'SignUpView' object has no attribute '_user_role'

`_user_role` is currently set on `PersonalizablePublicTournamentPageMixin` and `OptionalAssistantTournamentPageMixin`. I think in principle `_user_role` is a property of the page, not the user (user permissions are checked in a different mechanism), so I assume this should just be a simple matter of setting `_user_role` to `None` in the new base class (or similar). So I tried that, and then:

> Exception Type: AttributeError at /accounts/signup/sad/
> Exception Value: 'SignUpView' object has no attribute 'template_403_name'

This raises the question: What template should it be going to? Currently, we have three 403 templates:
- the general `403.html` is shown when the user's access level does not extend to the page sought
- `assistant_403.html` is shown when the user is an assistant accessing an assistant page, but it's disabled in tournament settings
- `public_403.html` is shown when the user is a public user accessing a public page, but it's disabled in tournament settings

The current behaviour is to go to the general 403 (thanks to the `SignUpView.dispatch()` override).

But I feel like an incorrect signup key should be going to none of these. Strictly speaking, since the key is _part of the URL_, when someone tries to access the page with the wrong key, they are trying to access a URL that doesn't exist. So I think the response should be a 404, not a 403, like we do for private URLs. Then rather than go through `is_page_enabled()` and generate a template response, it can just `raise Http404()` as soon as it figures out that it's rejecting the key.

I'm on board with the other refactoring here, even though I think this particular use case ends up not being a great example.

(Status of review is "request changes" because it currently gives a 500. The expected behaviour, which I think should be changed to a 404, can be up for discussion.)

PR #1847, opened 2021-07-21
Check for icon cells in table CSV conversion

tienne-B:


philipbelesky:
Hey, thanks for catching this!

PR #1874, opened 2021-09-27
Make Calico deployment more prominent

tienne-B:
This commit moves the paragraph about Calico over Heroku and adds a similar "Deploy" button. It also fixes a broken anchor.

Calico should be prioritized for a few reasons, namely that it is the most effective tool to enforce the licence, and that it requires the least effort for the users to deploy and maintain, with fewer problems arising.

The link to create a site in the footer is also replaced with the README as to present Calico.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -15,7 +15,7 @@
 
 Tabbycat is a draw tabulation system for British Parliamentary and a variety of two-team formats. It was used at Australs 2010 and 2012–2019, EUDC 2018, WUDC 2019–2020 and many other tournaments of all sizes and formats. To see an example of a post-tournament website, have a look at the [WUDC 2020 tab website](https://wudc2020.herokuapp.com).
 
-**Want to try it out?** The best way to trial Tabbycat is just to launch a new site, as described [in our user guide](https://tabbycat.readthedocs.io/en/stable/install/heroku.html) (or [below](#installation-and-user-guide)). It takes just a few clicks, costs nothing, requires no technical background, and you can always deploy a fresh copy when you're ready to run your tournament.
+**Want to try it out?** The best way to trial Tabbycat is just to launch a new site, as described [below](#%EF%B8%8F-installation)). It takes just a few clicks, requires no technical background, and you can always deploy a fresh copy when you're ready to run your tournament.
 
 ## 🔍 Features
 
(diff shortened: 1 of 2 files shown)


.github/README.md
```diff
@@ -35,11 +35,15 @@ Our user guide is at [tabbycat.readthedocs.io](http://tabbycat.readthedocs.io/).
 
 Tabbycat can be deployed in a number of ways. While you can set it up to [run on your own computer](https://tabbycat.readthedocs.io/en/stable/install/local.html) most users will want to run it as a website.
 
-Tabbycat is configured for a '1 click' install to the [Heroku](https://www.heroku.com) web platform. Click this button to begin:
+[Calico](https://calicotab.com/) is Tabbycat's managed hosting service, offering websites for a flat fee and handling all the technical aspects for you. Click this button to deploy to Calico:
```
philipbelesky, line 38: I'd prefer something more like "Calico is a managed hosting service run by one of Tabbycat's developers. For a flat fee, it will host tab websites, automatically manage their setup and performance, and provide ongoing access to the released tab."

(the second half of the sentence I don't feel strongly about, but thought it might be worth spelling out the benefits more precisely)

.github/README.md
```diff
@@ -35,11 +35,15 @@ Our user guide is at [tabbycat.readthedocs.io](http://tabbycat.readthedocs.io/).
...
+
+[![Deploy](https://raw.githubusercontent.com/gist/tienne-B/fc04ecd3c11a38424b642b4bba60e8d9/raw/b2c71d7d6a0d368d3e9dfd8002af729d155ad09b/calicodeploy.svg)](https://calicotab.com/tournaments/new/)
```
philipbelesky, line 40: I've had a go at making a more Heroku-like button (attached below). Would you prefer to use that? 

[button.svg.zip](https://github.com/TabbycatDebate/tabbycat/files/7242662/button.svg.zip)

philipbelesky (changes requested):
This looks good!

.github/README.md
tienne-B: I'm partial to it. The logo is too similar (like a copy-cat) and the font becomes avec serif if Inter isn't installed (or with a different name)

philipbelesky: Ah, I can change the text to render as an outline rather than with the font. Do you have any ideas for a logo substitute? Maybe just the arrow without the hexagon?

tienne-B: That'd be good.

czlee (approved):
Looks good from my end once you two have figured out the button 😊

.github/README.md
philipbelesky: Ok, updated. If the new version works for you feel free to merge the PR as a whole!

[button.svg.zip](https://github.com/TabbycatDebate/tabbycat/files/7270893/button.svg.zip)

PR #1898, opened 2021-11-14
Update instructions for backups using Postgres GUI

thbsp:
This commit reflects the current details page for the postgres add-on.


tienne-B:
My, I never knew the old interface! I've only ever had the new one.

The change is good; but if possible, change the commit message to use the imperative mood in the subject line, and manually wrap the body to <100 chars/line

PR #1899, opened 2021-11-14
Change instances of "iron man" to "iron person" in draw table and elsewhere

dcorks:
Elsewhere in the Tabbycat interface, the phrase "iron man" has been replaced with "iron person". 
I noticed it's still present in these popovers. I'm proposing this change for the sake of 
consistency.

### DIFF
--- a/docs/features/data-entry.rst
+++ b/docs/features/data-entry.rst
@@ -45,7 +45,7 @@ Duplicate/Swing Speeches
 
 .. image:: images/iron-speakers.png
 
-When entering the ballots there is a toggle label *'Iron' speeches*. When set to "yes" this allows you to have the same speaker deliver multiple speeches provided their extra speeches are labelled on the form as 'duplicates'. Typically, most tournaments require that lesser 'iron man' speech is discarded from the tab, which would mean that you would mark the lower speaker of the two scores as the duplicate (note that this may require you to check each score's average across a panel).
+When entering the ballots there is a toggle label *'Iron' speeches*. When set to "yes" this allows you to have the same speaker deliver multiple speeches provided their extra speeches are labelled on the form as 'duplicates'. Typically, most tournaments require that the lesser 'iron-person' speech is discarded from the tab, which would mean that you would mark the lower speaker of the two scores as the duplicate (note that this may require you to check each score's average across a panel).
 
 Speeches marked as duplicates are not included in the speaker tab. This means that they can also be used to exclude swing speakers from the tab as needed; even if they do not actually speak twice. To do so, change the name of the swing speaker to be that of an existing team member and ensure that that speech is marked as a duplicate.
 
--- a/docs/guide/tournament-logistics.rst
+++ b/docs/guide/tournament-logistics.rst
@@ -330,7 +330,7 @@ If at all feasible you want to train that volunteers acting as runners and/or da
   - Run through this in the actual tab room; illustrating examples with actual ballots and going through the roles in the actual spots which they will occur.
   - Run through how the seating/table/box arrangement works and the types of roles at different positions.
   - Emphasise that in general, any ambiguities should be raised with the tab directors/assistants; i.e. that you should never guess about ballots but instead always delegate resolving issues to someone else.
-  - Run through the different edge cases and things to check during entry. For example Iron Person speeches, mismatched totals, entering the wrong ballot for the wrong panellist, etc (see section below). Be sure to also go through what happens when the validation step fails; i.e. when a ballot needs to be re-entered.
+  - Run through the different edge cases and things to check during entry. For example, iron-person speeches, mismatched totals, entering the wrong ballot for the wrong panellist, etc. (see section below). Be sure to also go through what happens when the validation step fails; i.e. when a ballot needs to be re-entered.
 
 Training the adjudication core
 ------------------------------
--- a/tabbycat/results/templates/BallotEntryHeader.vue
+++ b/tabbycat/results/templates/BallotEntryHeader.vue
@@ -70,7 +70,7 @@
         </div>
         <div class="alert alert-info mb-0" v-if="showDuplicates">
           Speeches marked as 'duplicates' are hidden from the speaker tab and often need to be
-          tracked in order to determine break eligibility. If a speaker is 'iron-manning' you would
+          tracked in order to determine break eligibility. If a speaker is 'iron-personing' you would
           typically set their lowest-scoring speech as the 'duplicate'.
         </div>
       </div>
--- a/tabbycat/results/templates/ballot/ballot_speaks.html
+++ b/tabbycat/results/templates/ballot/ballot_speaks.html
@@ -16,7 +16,7 @@
 
         <div class="iron-person small pt-0 m-0 form-check form-check-inline" style="display: none;"
                data-toggle="tooltip" data-placement="bottom"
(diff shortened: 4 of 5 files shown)


tienne-B (commented):
Only one "n" in "[personed](https://en.wiktionary.org/wiki/personed)".

Same deal with the commit message. How about "Change 'iron-manned' to 'iron-personed' in draw table"?

Did a quick grep:
* 9 matches "iron-man"
* 1 match "iron man"
* 2 matches "iron person" (case insensitive)
* 15 matches "iron-person"

Could be good to replace them all now, not just a file at a time.

czlee:
Oh wow, that's crazy, that spelling makes no sense phonically. I checked the reputable (non-wiki) dictionaries and none of them list "person" as a verb, but Oxford lists "[personed](https://www.lexico.com/definition/personed)" with the same pronunciation in a related (but different) context, and Collins lists "[personned](https://www.collinsdictionary.com/dictionary/english/personned)" as the adjective. I guess that's English for you, huh 🤷🏻‍♂️

Personally, I preferred the verb "to iron", as in, "the team _ironed_ the last round", but much to my chagrin I never managed to popularise that while I was on the circuit 😂 (and its past tense/participle also only has one 'n' 😑)

thbsp:
> Did a quick grep:
> * 9 matches "iron-man"
> * 1 match "iron man"
> * 2 matches "iron person" (case insensitive)
> * 15 matches "iron-person"
> Could be good to replace them all now, not just a file at a time.

I'm happy to change the instances you mentioned but I don't know how to find them.
I tried to do a grep myself to find those instances but I found just the affected lines in tables.py (already in the commit) and one in line 359 of models.py which explains what it means to "iron-person" (seems reasonable to keep). (I limited my search to just in the .py files because when I didn't I got hundreds of results from the various locale folders.)

tienne-B:
I like the term "iron"; it's two fewer syllables!

You can exclude `.po` and `.mo` files in grep, and that's where you'd get the results I found. I somewhat prefer having the hyphen if we choose to append "person"; what would be important is keeping it consistent though.

czlee:
If it helps:
```
$ git grep -i "iron(-|\w)man" ':(exclude)*.po' ':(exclude)*.mo'
.github/CHANGELOG.rst:748:- Added the ability to mark speeches as duplicates when entering ballots so that they will not show in speaker tabs, intended for use with 'iron-man' speeches and swing speakers
docs/api-schema.yml:4060:                                  known as iron-person or iron-man speeches).
tabbycat/results/migrations/0001_initial.py:53:                ('ghost', models.BooleanField(default=False, help_text='If checked, this score does not count towards the speaker tab. This is typically checked for speeches where someone spoke twice to make up for an absent teammate (sometimes known as "iron-person" or "iron-man" speeches).', verbose_name='ghost')),
tabbycat/results/models.py:359:            "\"iron-man\" speeches)."))
tabbycat/results/templates/BallotEntryHeader.vue:73:          tracked in order to determine break eligibility. If a speaker is 'iron-manning' you would
tabbycat/results/templates/ballot/ballot_speaks.html:19:               title="{% trans "Duplicate speeches are hidden from the speaker tab. If a speaker is 'iron-manning' you would typically mark only the lesser of their scores as a duplicate." %}">
tabbycat/utils/tables.py:927:                        popover_text = _("Team iron-manned this round and the last.")
tabbycat/utils/tables.py:930:                        popover_text = _("Team iron-manned this round.")
tabbycat/utils/tables.py:933:                        popover_text = _("Team iron-manned last round.")
```

I'd leave the changelog and migrations alone. The instances in docs/api-schema.yml:4060 and results/models.py:359 (the latter pointed out by @thbsp) offer both so should be left as-is.  That just leaves results/templates/BallotEntryHeader.vue and results/templates/ballot/ballot_speaker.html, both of which are not Python files so I think you'd be forgiven for not finding them 😸 (though they are relevant files that should be updated).

Thanks for your help!! Definitely great to tidy this up.

thbsp:
I've added the changes suggested, so I think I've caught everything.
I changed all the instances I found of '`iron man`', '`iron-man`' and '`iron person`' to '`iron-person`': with hyphen, no capitals, and without a doubled consonant.
I only edited things in user-facing text or docs, and left instances where '`iron-man`' and '`iron-person`' appear together.
Thanks for @tienne-B and @czlee for the help with grep. Let me know if I missed anything.

tienne-B:
I think there are two terminological decisions to be made for consistent application:
* Whether the term should be avoided (like with the formulations in `admin_results.html` which may seem cluncky);
* Whether quotation marks should be used around the whole term (`results/models.py`), just around "iron" (`ballot_entry_base.html`) or not at all (`utils/tables.py`).

I would suggest using the term rather than eschewing it for the first point, and not including in quotes for the second.

Further, the help-text of the `ghost` field in `results/models.py` could be changed to remove "iron-man" (don't worry about a migration).

czlee:
My 2¢:

> Whether the term should be avoided (like with the formulations in `admin_results.html` which may seem cluncky)

My general guide with terminology (which we're not very consistent about) is to use the "standard" term if one exists, and to choose formal terms ("preliminary round") over colloquial terms ("inrounds"). Sometimes this doesn't go very well, because every region uses a different term ("conflicts"/"clashes"/"scratches", though in this case we're just bound by our Australs origins).

As far as I know, "iron-person" is probably the "most standard" term there is, but you can kind of see in existing text that, when we implemented this feature, we didn't consider it standard _enough_ that everyone should just know what it means (in contrast to, _e.g._, "break").

I guess I'm in the "reluctant default to this term" camp, since while I don't like this term / consider it too colloquial, I'm not sure what other name we'd give it. But if there's less cryptic term available, I'm sure it wouldn't take much to convince me.

> Whether quotation marks should be used around the whole term (`results/models.py`), just around "iron" (`ballot_entry_base.html`) or not at all (`utils/tables.py`).

This kind of relates to what I said above, I guess—when we added this feature, we didn't consider it to be "standard" or "clear" enough to use without explanation. (I'm sure the term has caught on more widely since I was on the circuit, but I still worry that it's not standard among circuits not closely tied to WUDC.) 

I'm not sure if it was me or @philipbelesky who wrote the text, but I'm guessing we used quotation marks where there was enough space to explain what it is/what's going on, and omitted quotation marks where space was constrained and we were using the term by itself (_e.g._ in those debate detail popovers). I'd be inclined to do that: I don't think it's inconsistent to introduce a term using quotation marks, and then to use the term without quotation marks thereafter, though I admit that "thereafter" is a more arguable concept in software than in prose.

We should change "'iron' person" to "'iron-person'" though.

philipbelesky:
Agree with CZ regarding changing `"iron"` to `iron-personed`. In ideal world there would be a relatively concise way to describe the situation as I think jargon like this is non-ideal from a translation/global perspective. But, "had a speaker who spoke twice" is the best I can come up with and that probably is excessive for tooltips etc.

czlee:
Brainstorming: How about "repeat speaker"?

I guess our terminology is kind of all over the place, iirc we describe the checkboxes as "duplicate speeches", which makes sense because you check the speech that is "not counted", but idk if it might be confusing… Partly the problem here is that there is meant to be some sort of "abnormal" connotation going on here—as in, in general, teams should not be doing this, and something like "repeat speaker" might not convey that.

Or "fill-in speaker"? Though that doesn't reflect that someone spoke twice.

tienne-B:
Maybe at this point we make a FB post for how best to describe/name it; we also use "ghost", so it is inconsistent everywhere.

tienne-B:
Sorry for the delay in merging this; it escaped our radar. I think we're all fine with the terminology and it seems to be in wide use anyways.

PR #1901, opened 2021-11-18
Order debates in private URL ballots column by seq

### LINKED ISSUE #1900: 'View ballot' links in judges' private URL pages point to wrong ballot
I noticed a bug with the 'view ballot' links in the judges' private URL pages.
My tournament has had three prelim rounds so far.
For any judge who was present for all three rounds, on their private landing page the 'view ballot' links on the landing page for their private URL for R2 and R3 point to the wrong place. The link for R2's ballot actually points to the R3 ballot, and the link for R3's ballot points to the R2 ballot.
All other permutations seem to have no problems. (Pages for judges available for only two rounds so far or just one round so far are fine.)
Status as chair, panel or trainee judge seems to have no effect.

Here's the link to the public version of my page: https://fall2021sdc.herokuapp.com/fall2021sdc/
The issue can be seen on the private URL pages for the following judges:
"We... Th..."
"Pa... Jo..."
"Re... Ha..."

(I don't want to share the judges' full names here, and I don't want to share the private links here either since the tournament is ongoing. I'm guessing the devs have backdoor access to the database. If so, the above info will be enough to find the right pages. If not, I can send the private links through a email or something.)

tienne-B:
The superseding debates queryset when adding the ballots link column to the table is not sorted, unlike the squashed debates list. This sets the ordering to be like its invocation, by round sequence.

Fixes #1900


czlee:
Oh, this is super interesting. Is this the root problem? In theory, participants/tables.py:80 should be passing a list of debates to `table.add_debate_ballot_link_column(debates)`, which should just add the rows _in the same order_ as the sequence `debates` that is was passed. It seems like it's not doing that: (existing code, not PR)
https://github.com/TabbycatDebate/tabbycat/blob/bdeacfe4b13727bb41fef89455ea5e581db7feeb/tabbycat/utils/tables.py#L848-L854
Line 849 overwrites `debates` with a new QuerySet, then line 854 loops through this new `debates`. But the new `debates` isn't guaranteed to have the same order as the original `debates`, which I'm guessing is where the mismatch arises? So I feel like the proper solution is to guarantee that the loop in lines 854 runs in the same order as the original `debates`.

One way (not necessarily the best way) would be to put the results of the line-849 query into a dict by debate ID, which the loop then reads while looping through the original `debates`.

I worry a little that resolving this by dictating the `order_by` of both queries would still miss edge cases, for example, if an adjudicator is assigned to more than one debate in a round (then ordering isn't unique), or if for whatever reason (_e.g._ a future change) the query in utils/tables.py:849 doesn't include all the debates in its result. Better to code the direct intent.

tabbycat/utils/tables.py
```diff
@@ -846,15 +845,13 @@ def add_debate_ballot_link_column(self, debates, show_ballot=False):
...
-                Prefetch('ballotsubmission_set', queryset=BallotSubmission.objects.exclude(discarded=True), to_attr='nondiscard_ballots'))
+            populate_debate_has_ballots(debates)
             ballot_links_data = []
             for debate in debates:
-                if not debate.has_ballot:
+                dummy_bs = BallotSubmission(debate=debate, single_adj=self.tournament.pref('individual_ballots'))
+                if not debate._has_ballot:
                     ballot_links_data.append(_("No ballot"))
-                elif not get_result_class(debate.nondiscard_ballots[0], debate.round, self.tournament).uses_speakers:
+                elif not get_result_class(dummy_bs, debate.round, self.tournament).uses_speakers:
```
czlee, line 854: Would it be cleaner to refactor `get_result_class` so that passing in a `BallotSubmission` isn't mandatory if there's another way to get the required information? _e.g._ permit passing in `single_adj` instead of `ballotsub` (but still generally expect `ballotsub`)? (We should probably add a docstring to `get_result_class`, I guess.)

PR #1904, opened 2021-11-25
Filter private URL ballot link individually

tienne-B:
When using non-conferral formats, it would be important that the adjs do not see each other's ballots through their private URL. For that, a new filter is added for their latest individual single-adj ballot, rather than just non-discarded.

### DIFF
--- a/tabbycat/results/views.py
+++ b/tabbycat/results/views.py
@@ -733,13 +733,19 @@ class AdjudicatorPrivateUrlBallotScoresheetView(RoundMixin, SingleObjectByRandom
     def is_page_enabled(self, tournament):
         return True
 
+    def get_ballotsub(self):
+        filters = Q(discarded=False)
+        if self.tournament.pref('individual_ballots'):
+            filters &= Q(participant_submitter__url_key=self.kwargs.get('url_key'), single_adj=True)
+        return self.object.ballotsubmission_set.filter(filters)
+
     def check_permissions(self):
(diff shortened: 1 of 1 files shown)


czlee:
Oh, interesting. I suppose one could argue that we want their own latest non-discarded ballot before confirmation, and the whole panel once it's confirmed. But I can also see some tournaments wanting judges only to see their own scores (as in this PR). Probably not worth the complication to do the latter; the primary purpose of this feature was to allow judges to check their own scores and (potentially) alert tab about data entry errors.

czlee:
Actually, scratch that.

I first realised that a consequence of this change was that if an adj didn't submit an individual ballot, this page will show a generic "There is no result yet" error message, which could be confusing, and given that we have a semi-elaborate mechanism for detailed error messages via `check_permissions()`, maybe we should provide a different error message in this situation.
image


But then I thought… what if the adj submitted an individual ballot, but then the confirmed aggregated ballot is (for whatever reason) _different_ to their individual submission? Then which ballot should we be showing?

Or… outside the scope of this PR, but definitely related: What if (in non-individual cases) the confirmed ballot is not the most recent submission? Is the _existing_ behaviour of this page wrong, in that it will choose the most recent non-discarded ballot over the one that is actually confirmed, if one exists?

(Then you might wonder why this page doesn't already filter for confirmed ballots… and if my memory serves me right, the answer is, it's not meant to, because it's also meant to permit adjudicators to check their draft ballot submissions, _before_ the tab room confirms them.)

PR #1905, opened 2021-11-28
Prevent confirming single-adj ballots

tienne-B:
This change does two things to prevent single-adj ballots from being confirmed:
* Disabling the confirmed field on single-adj ballots;
* Removing the effect of the auto-confirm preference.

With this adding more complexity, templates have been changed to reflect the value of the confirmed field rather than the preference.

A question to think about is whether assistants may confirm ballots they've merged (currently not).

Fixes BACKEND-6EB


czlee:
Oh, huh. This is definitely something we should do. But with respect to BACKEND-6EB, we probably still need to be able to handle the situation, because it's not a prohibited database state.

czlee (approved):
While reviewing this I noticed a bug in the "bypass double-check" (`disable_ballot_confirms`) option that existed before this PR, will raise separately.

PR #1906, opened 2021-11-29
Fix assistant form behaviour when checks bypassed

czlee:
Discovered when reviewing #1905. The assistant form does weird things when `disable_ballot_confirms` (bypass double-checking) is enabled:
image

I think it's just disabling the button on a weaker condition than it should be applying, as indicated by the `not new` condition a few lines above it (for displaying the "can't confirm your own entry" error message).

Also, I don't know why there's a condition on the `btn-success` class… I assume it was to stylize the button differently when it's disabled, but our CSS already does that.

### DIFF
--- a/tabbycat/results/templates/assistant_enter_results.html
+++ b/tabbycat/results/templates/assistant_enter_results.html
@@ -43,7 +43,7 @@
           <input id="id_confirmed" type="hidden" name="confirmed" value="True" />
           <div class="row">
             <div class="col">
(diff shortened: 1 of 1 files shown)


czlee:
(in principle any second pair of eyes will do, but @tienne-B might find it easier since he just looked at this same chunk of code in #1905)

tienne-B:
It was actually merged as 73559113062345c8c7e0235e1af1e6296d3648f1

PR #1926, opened 2021-12-04
Bump channels to 3.0.4

czlee:
This upgrade is slightly nontrivial; it involves consequential compatibility changes:
https://channels.readthedocs.io/en/stable/releases/3.0.0.html

So just looking to get a second pair of eyes to make sure I haven't done anything stupid.


tienne-B:
I was looking at this in my work making Websocket API views!

Issue #1929, opened 2021-12-27
Docker compose issue on MacOS


Joshuah143:
When I downloaded release 2.6.4 and ran the `osx_docker_start.command` file I got this error: `Error response from daemon: pull access denied for tabbycat, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

It seems to be some docker config error, should I have added the DB and redis details first?

I did my best to follow the instructions [posted to the docs](https://tabbycat.readthedocs.io/en/stable/install/docker.html) but, for some reason, it didn't work. I'm a newbie to docker so this is likely a pretty dumb mistake. Thanks for your help in advance!

Steps taken:
1. Download from GitHub
2. Download Docker
3. Open Docker
4. Run `docker login`
2. Navigate to file and run `osx_docker_start.command`

Versions:
- OSX = 12.0.1
- Docker = 20.10.11
- TabbyCat = 2.6.4

Command line:

> USER@MACHINE tabbycat-2.6.4 % /Users/USER/Downloads/tabbycat-2.6.4/bin/osx_docker_start.command
> [+] Running 0/4
>  ⠿ redis Error                                                                                                                                                                                         2.1s
>  ⠿ worker Error                                                                                                                                                                                        2.1s
>  ⠿ db Error                                                                                                                                                                                            2.1s
>  ⠿ web Error                                                                                                                                                                                           2.1s
> Error response from daemon: pull access denied for tabbycat, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

Also, just to note, the reason that I am trying to use docker instead of a traditional local install is that I'm hoping to be able to put the container on Google Cloud. So any advice specific to that is very welcome!

Joshuah143:
Update:

Fixed by replacing `image: Tabbycat` with `build: .` in the docker-compose.yaml file for both worker and web services.

Should I submit a PR for this or just leave it as a one-off in my setup? I'm more than happy to submit one.

tienne-B:
Mind submitting a PR?

Joshuah143:
Created pull request, worked for me after making the change

tienne-B:
Closed with #1931

PR #1931, opened 2021-12-27
Update docker-compose.yml

Joshuah143:
Removed image and replaced with build from local to prevent permissions errors. See #1929

Noticeably, after testing, there were login issues for me. I would chalk those up to just my machine. Though might be worth double-checking. Seems entirely unconnected to my small change.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,6 @@ services:
 
   web:
     build: .
-    image: tabbycat
     # Hack to wait until Postgres is up before running things
     command: ["./bin/docker-wait.sh", "--timeout=0", "db:5432", "--", "./bin/docker-run-honcho.sh"]
     depends_on:
@@ -46,7 +45,7 @@ services:
     working_dir: /tcd
 
   worker:
-    image: tabbycat
+    build: .
     # Hack to wait until migration is done before running things
     command: ["./bin/docker-wait.sh", "--timeout=0", "web:8000", "--", "./bin/docker-run-worker.sh"]
     depends_on:


philipbelesky:
Hey, thanks for the detailed report and for the PR. I'm on the same build environment and this change works with both the `.command` and the normal compose command (although they are basically the same thing). 

The login issues you mentioned, is that related to logging into the app itself? And is the issue that it won't accept a username/login, or that the interface state doesn't match your logged-in status?

Joshuah143:
They are pretty much identical, but for some reason, the original just wouldn't work, kept getting: `Error response from daemon: pull access denied for tabbycat, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`.

This issue is that the interface doesn't match the logged-in status and keeps asking me to log in again.

Joshuah143:
Having re-composed the docker image and looked at the logs I see:

> tabby-web-1     | 22:13:41 asgi.1  | 2021-12-28 09:13:41,348 - daphne.ws_protocol - DEBUG - WebSocket closed for ['172.21.0.1', 0]
> tabby-web-1     | 22:13:48 asgi.1  | 2021-12-28 09:13:48,965 - daphne.http_protocol - DEBUG - Upgraded connection ['172.21.0.1', 0] to WebSocket
> tabby-web-1     | 22:13:48 asgi.1  | 2021-12-28 09:13:48,971 - daphne.server - INFO - failing WebSocket opening handshake ('Access denied')
> tabby-web-1     | 22:13:48 asgi.1  | 2021-12-28 09:13:48,972 - daphne.server - WARNING - dropping connection to peer unix:None with abort=False: Access denied
> tabby-web-1     | 22:13:48 asgi.1  | 2021-12-28 09:13:48,972 - daphne.ws_protocol - DEBUG - WebSocket ['172.21.0.1', 0] rejected by application
> tabby-web-1     | 22:13:48 nginx.1 | 172.21.0.1 - - [27/Dec/2021:22:13:48 +0000] "GET /ws/bp88team/action_logs/ HTTP/1.1" 403 5 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.1 Safari/605.1.15"

Super strange, any ideas? I'll create a separate issue entirely for that.

Seems unconnected to the PR, worth moving to 'under review' (or just merging)?

Joshuah143:
I also added a second commit, which should just be some clean-up of the tabby/adjudication/views.py file to move the method `add_message` to the parent class `BaseAdjudicatorConflictsView`. 

Had some issues running tests (see above comments), but should be nothing more than superficial.

tienne-B:
> I also added a second commit, which should just be some clean-up of the tabby/adjudication/views.py file to move the method `add_message` to the parent class `BaseAdjudicatorConflictsView`.

If you want to add that message, it should raise a `NotImplementedError` as the message depends on which model is conflicted, and isn't/shouldn't be specified in the base class.

Joshuah143:
Sorry my bad, reverted back to just the original changes.

philipbelesky:
Hmm, to my eyes the logs point towards something websocket related rather than database related. Maybe the handshake fails as it wants to be running over https? Anyway, let's explore on a separate issue. Thanks again for the PR!

PR #1951, opened 2022-01-01
Enable Redis in Docker compose file

### LINKED ISSUE #1932: Docker web socket/login issues
After completing the setup of the develop branch on docker using the `docker compose up` command. I experienced an issue with login, it would take my credentials but not authenticate them and put me back to the page accessed immediately before the login request in a logged-out state. 

The logs are as follows:
> tabby-web-1 | 22:13:41 asgi.1 | 2021-12-28 09:13:41,348 - daphne.ws_protocol - DEBUG - WebSocket closed for ['172.21.0.1', 0]
> tabby-web-1 | 22:13:48 asgi.1 | 2021-12-28 09:13:48,965 - daphne.http_protocol - DEBUG - Upgraded connection ['172.21.0.1', 0] to WebSocket
> tabby-web-1 | 22:13:48 asgi.1 | 2021-12-28 09:13:48,971 - daphne.server - INFO - failing WebSocket opening handshake ('Access denied')
> tabby-web-1 | 22:13:48 asgi.1 | 2021-12-28 09:13:48,972 - daphne.server - WARNING - dropping connection to peer unix:None with abort=False: Access denied
> tabby-web-1 | 22:13:48 asgi.1 | 2021-12-28 09:13:48,972 - daphne.ws_protocol - DEBUG - WebSocket ['172.21.0.1', 0] rejected by application
> tabby-web-1 | 22:13:48 nginx.1 | 172.21.0.1 - - [27/Dec/2021:22:13:48 +0000] "GET /ws/bp88team/action_logs/ HTTP/1.1" 403 5 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.1 Safari/605.1.15"

Notably, this is running with HTTP, not HTTPS.

Versions:
Docker: 1.4.12
Host: MacOS 12.0.1
Tabbycat: 2.7.0

czlee:
Fixes #1932.

This sets the `DOCKER_REDIS` environment variable in docker-compose.yml, so that Redis will be set up as in settings/docker.py:
https://github.com/TabbycatDebate/tabbycat/blob/5c68911acd5fed04103493e2e60a81c2049a5a73/tabbycat/settings/docker.py#L20-L41

Redis seems to be necessary for the `"django.contrib.sessions.backends.cache"` session engine to work. Or at least, I think it is… I'm not sure if we fall back to another cache if Redis isn't running?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,7 @@ services:
       - DEBUG=1
       - IN_DOCKER=1
       - DISABLE_SENTRY=1
+      - DOCKER_REDIS=1
     ports:
       - "127.0.0.1:8000:8000"
     volumes:
@@ -55,6 +56,7 @@ services:
       - DEBUG=1
       - IN_DOCKER=1
       - DISABLE_SENTRY=1
+      - DOCKER_REDIS=1
     volumes:
       - ./tabbycat/settings:/tcd/tabbycat/settings
     working_dir: /tcd


docker-compose.yml
```diff
@@ -55,6 +56,7 @@ services:
       - DEBUG=1
       - IN_DOCKER=1
       - DISABLE_SENTRY=1
+      - DOCKER_REDIS=1
```
czlee, line 59: does this need to be enabled in the worker as well?

philipbelesky: I don't think so - all the websocket and async functionality works (for me) as-is

philipbelesky (approved):
Ah, turns out that I had redis set to run on launch, hence why I didn't come across this.

docker-compose.yml
czlee: Just tried it out, adj and room allocations don't seem to work without it, though side editing and checkins work fine.

PR #1952, opened 2022-01-03
Upgrade to Python 3.9 and Django 4.0

czlee:
This doesn't work yet.

Two sticking points:

1. The `django-jet` library is incompatible in several places with Django 4.0. I've had a crack to see if the fixes are trivial in [czlee/django-jet](https://github.com/czlee/django-jet), and might submit a PR to [RamsesMartinez/django-jet](https://github.com/RamsesMartinez/django-jet) if they work out. But otherwise we might need to find another solution again—see #1507 for the last time we discussed this.

2. `channels` requires `asgiref<3.4`, but Django 4.0 requires `asgiref>=3.4.1`. This is a known issue, see django/channels#1713.

   We can't stay on `channels==2.4.0`, because it's also not compatible with Django 4.0 (because `LANGUAGE_SESSION_KEY` was removed; this has been fixed in `channels`). So we're currently blocked by django/channels 1713.

This includes:
- #1926
- #1935
- #1938
- #1941

(The tests pass, but `npm run serve` does not.)


czlee:
Okay, I think we're there!

1. [assem-ch/django-jet-reboot](https://github.com/assem-ch/django-jet-reboot) seems to do the trick for `django-jet`.
   - No updates on [RamsesMartinez/django-jet](https://github.com/RamsesMartinez/django-jet).
   - Sadly, [Barukimang/django-jet](https://github.com/Barukimang/django-jet) has a compatibility issue in its changelist handling: Django 4.0 added a `search_help_text` argument to the `ChangeList` constructor in https://github.com/django/django/commit/1143f3bb5ecaa2be58f2cd9077f147040291659d, and this fork hasn't modified [`get_model_queryset()`](https://github.com/Barukimang/django-jet/blob/aa55e032b8f1c9016e1f85ecae8d3610f1959351/jet/utils.py#L238-L257) to reflect this. assem-ch/django-jet-reboot has [updated the lines here](https://github.com/assem-ch/django-jet-reboot/blob/f37d8a775308604fc2b1db6d7704462dff03c989/jet/utils.py#L238-L244).
2. https://github.com/django/channels/issues/1713 has been resolved in asgiref 3.5.0. We could just remove that line from requirements.txt, but I've opted to lock the version in case this happens again.

PR #1954, opened 2022-01-15
Add button for direct preformed allocation

entinglee:
When allocating adjudicators, additional button for "Direct Allocation" which would allow preformed panels to be directly allocated to the draw instead of being allocated via Hungarian allocation


tienne-B:
Looks good! Made a little fix that wouldn't be evident (Tabbycat thinking all the settings should be applied tournament-wide, even for the allocation method, which isn't set as a preference), and linting (style problems). Good work!

PR #1955, opened 2022-01-15
Add button for direct preformed allocation

tienne-B:
When allocating adjudicators, additional button for "Direct Allocation" which would allow preformed panels to be directly allocated to the draw instead of being allocated via Hungarian allocation.


philipbelesky:
Hey, thanks for working on this @entinglee and @tienne-B, I can see how this would be super-useful for particular adj cores.

I've taken this as a chance to do a slight redesign of that modal window, in light of having an additional option to present/explain. This is the current draft:

Screen Shot 2022-01-15 at 6 08 36 pm

Any feedback would be great!

I renamed `allocateWithInfill` to `allocateIndividualAdjs` as the name initially made me think we had implemented an auto-allocation method that would fill gaps without wiping existing panels - iirc this is a feature we intend to do at some point. 

I also altered some of the showing/hiding logic:

1. So that the option to use the traditional allocator is always shown
2. To always show the preformed panel functionality (even if panels are not made). 
3. I wasn't sure why `notUsingPreformed` exists so remove it (let me know if that was a mistake)

A couple of things still to work on/address:

- [ ] This modal window is also shared with `EditDebateVenuesContainer.vue`; will need to check nothing breaks there
- [ ] I haven't fully tested the different combinations of the showing/hiding logic
- [ ] We should probably explain this new feature in the docs?

(feel free to tackle any of the above!)

tienne-B:
The modal seems too big now, and the preformed panels section would be useless if there are no preformed panels, which is the use of the `notUsingPreformed` property. Might also want to be clearer in the explanations of the options.

czlee:
Direct allocate doesn't take priority into account at all, right? And the point of smart allocate is that it also does conflict avoidance? So for descriptions, how about something like:
> **Smart Allocate:** Allocates preformed panels to debates of similar priority level, while avoiding conflicts
> **Direct Allocate:** Allocates panels in exact order going from top to bottom (ignoring priority/conflicts)

(Seems like adj cores prefer the overhead in post-allocation fixed, than pre-allocation priority assignment, which is fine—just trying to be as helpfully descriptive as possible, though if anyone feels this is too biased in favour of smart allocate, feel free to edit to correct for it.)

philipbelesky:
> The modal seems too big now, and the preformed panels section would be useless if there are no preformed panels, which is the use of the notUsingPreformed property. Might also want to be clearer in the explanations of the options.

The panel hides the buttons when no panels have been formed. Then, it just notes how to make the panels. I think this is worthwhile to include, both to highlight the feature exists and highlight where to find it in the UI (if it is ever used). I'm not too worried about the overall size of the modal unless it doesn't fit within a standard screen resolution. 

> Direct allocate doesn't take priority into account at all, right? And the point of smart allocate is that it also does conflict avoidance? So for descriptions, how about something like:

These sound good, they'll be in the next commit.

Issue #353, opened 2016-08-02
Add swappable panels to allocation pages


philipbelesky:
Requires only front-end changes; shouldn't be too difficult. Off limited use though, so low-priority for now.

philipbelesky:
Ok, this has landed in ed3ccae9fceba048c1aa1d374ccc69661c6b3d74. I got the hoverable conflicts going, so you should be able to see any potential conflicts before a swap is made. I didn't get the conflicts panel (e.g. the top overlay) going as I think there would just be too much information there to parse. 

I've had a check of the functionality and the over drag/drop pages for any issues. At first glance they seem fine, but it would be good to test these changes a bit before they are released.

czlee:
Woohoo! 🎉 Always exciting to see a three-digit PR get closed as completed!

PR #1957, opened 2022-01-17
Add swappable panels (#353)

philipbelesky:
Just getting the ball rolling...


philipbelesky (commented):
This should be reverted - it breaks dragging adjs between debates. I think what's happening is that the behaviour should stay in the mixin, rather than the template, because we want to prevent duplicate drag-registrations from when there are nested components that trigger at each level of nesting (e.g. adjs as draggable items within a panel as a draggable collection).

PR #1958, opened 2022-01-18
Update app.json

theminh05:


tienne-B:
This is the official repository, so `app.json` should link to here, not a fork.

PR #1973, opened 2022-02-05
Update app.json

LnhSchools:


tienne-B:
I think this is accidental. This repo remains the canonical source, so we aren't going to change `app.json`.

Issue #1667, opened 2020-11-24
Create Hungarian-based two-team draw generator


tienne-B:
To have a more powerful conflict avoidance than one-up-one-down, a Hungarian draw generator similar to the one for BP could be created using penalties as costs. The penalties could then be balanced with the allocation method (slide, fold, etc.) if necessary.

czlee:
Ah, this would be quite the project 😅 but I guess it's not too dissimilar in principle to the BP draw generator.

austinredmond:
Thanks for opening an issue. I have some done some hand draws assisted by whatever algorithm chutabs uses. What materials should I familiarize myself with if I wanted to attempt some work on this project?

tienne-B:
ChuTabs uses a similar algorithm to the one we currently use to resolve conflicts, but continues trying to swap until a pairing is found without conflict. In that, there is no concept of "penalties" in ChuTabs and will attempt to resolve any conflict. The code in ChuTabs is in `lib/inc.pairing.cp.php` if you want. (I just downloaded it from [SourceForge](https://sourceforge.net/projects/chutabs/); hadn't looked at its code until now) To implement such a system in Tabbycat, you'd need to adapt [tabbycat/draw/generator/one_up_one_down.py](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/draw/generator/one_up_one_down.py) (in a new generator) to want to swap to the closest "unencumbered" team (or where the conflict is strictly less than in the current matchup, although that'd be more hybrid). This would be simple enough to implement, with the swapping mechanism already there, and lots of comments to explain what the code does. You'll then want to add the feature in [tabbycat/draw/generator/powerpair.py](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/draw/generator/powerpair.py) (with "avoid_conflicts"), add a few flags to [tabbycat/draw/generator/__init__.py](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/draw/generator/__init__.py) and finish it off by adding it as a preference (and document, of course):
https://github.com/TabbycatDebate/tabbycat/blob/fde12c67b626e44c6eb8d0fda100326c02ba909c/tabbycat/options/preferences.py#L223-L234

For random draws where ranks are less important, you could take a look here: https://github.com/TabbycatDebate/tabbycat/blob/fde12c67b626e44c6eb8d0fda100326c02ba909c/tabbycat/draw/generator/random.py#L47-L70

[...]

PR #1998, opened 2022-03-20
Remove rank annotators from draw generation

### LINKED ISSUE #1997: ValueError when generating draw with only one team standings metric
Sentry Issue: [BACKEND-77T](https://sentry.io/organizations/tabbycat/issues/3116046380/?referrer=github_integration)

```
ValueError: ExpressionList requires at least one expression.
(10 additional frame(s) were not displayed)
...
  File "standings/base.py", line 366, in generate_from_queryset
    queryset = annotator.get_annotated_queryset(queryset, self.queryset_metric_annotators, *self.options["rank_filter"])
  File "standings/ranking.py", line 73, in get_annotated_queryset
    self.key          : self.get_annotation(annotators, min_field, min_rounds),
  File "standings/ranking.py", line 165, in get_annotation
    return Window(
  File "django/db/models/expressions.py", line 1287, in __init__
    self.order_by = ExpressionList(*self.order_by)
  File "django/db/models/expressions.py", line 896, in __init__
    raise ValueError('%s requires at least one expression.' % self.__class__.__name__)
```

czlee:
Fixes #1997

The subrank annotator crashes when there is only one team standings metric (because there's nothing to subrank on). But I don't think the ranking annotators are necessary for draw generation anyway, as it just works on the ordering returned, not the rank annotations. So this just removes them.

See discussion in #1997 about whether this should be a hotfix.

### DIFF
--- a/tabbycat/draw/manager.py
+++ b/tabbycat/draw/manager.py
@@ -197,7 +197,7 @@ class PowerPairedDrawManager(BaseDrawManager):
         metrics = self.round.tournament.pref('team_standings_precedence')
         pullup_metric = PowerPairedDrawGenerator.PULLUP_RESTRICTION_METRICS[self.round.tournament.pref('draw_pullup_restriction')]
 
-        generator = TeamStandingsGenerator(metrics, ('rank', 'subrank'), tiebreak="random",
+        generator = TeamStandingsGenerator(metrics, (), tiebreak="random",
             extra_metrics=(pullup_metric,) if pullup_metric and pullup_metric not in metrics else ())
         standings = generator.generate(teams, round=self.round.prev)
 


tienne-B:
Ranking annotators are what gets ordered; without them the order would be arbitrary. A better solution in my view would be to skip subrank annotators if unapplicable.

czlee:
Does `Standings.sort` not just sort the teams based on the provided metrics? Or did you change this when you incorporated rankings into the SQL queries using window functions?

Originally my thinking on design here was:
- the team standings precedence is the thing that decides team order; teams in a standings object are always sorted by team standings precedence
- ranking annotators are just there to provide easy access to a "ranking" field, in particular one that marks equally ranked teams correctly (as team order wouldn't do this)

Is there are reason not to think about it like this? Like, do we ever want to generate standings with non-ranked order, or have them in an order that is not dictated by team standings precedence?

tienne-B:
The queryset pathway is in parallel to the non-query path, used when all the metrics can be combined in one query. As such, `.sort()` still sorts, but would be bypassed in most metric combinations for draw generation. I see the ranking annotators not just as a different `enumerate()` generator, but as explicit calculation for the ordering of the teams. The ranking is the semantic information, not the metrics themselves.

czlee:
Would it break anything to sort by the metrics as specified in the team standings precedence? Like, would it impact any behaviour, other than to make the output more predictable when no ranking annotators are specified, or would it unduly complicate code paths or cause something to have to be treated as a special case?

tienne-B:
You'd have to change `base.py` to order by the metrics rather than by ranking, which would complicate the PR. However that doesn't really respond to my conception of the ranking as essential/semantic information as draws are constructed based on their ranking, and by removing it in favour of by ordering, it becomes just implicit.

tienne-B:
Revisiting this, as this PR would be incompatible with work on #1667, which would use the ranking annotators (specifically `subrank` as a factor in penalty calculations to avoid major distortions in brackets.

PR #2016, opened 2022-04-16
Migrate to Pipenv

czlee:
This removes our `requirements.txt` structure in favour of Pipenv's `Pipfile` and `Pipfile.lock`.

The primary motivation is to make it easier to update lots of dependencies simultaneously.

This does come with some costs:
- Pipfile requires a single Python version, so we'll need to stick to supporting just one Python version. (Though, this is probably a good thing.)
- We have to install `pipenv` in all our environments in order to install or deploy anything. In environments that don't require virtual environments, like Heroku, docker and Django CI, we can use `pipenv install --system` to install the packages using the system `pip`.
- [x] All developers and local-installers will need to use `pipenv`, so the docs need to be updated accordingly. (Still working on this.)

I've checked:
- Docker on my (Windows) machine, if I fully destroy all containers, images and volumes, and start from scratch
- The `deploy_heroku.py` script

I haven't checked the Deploy to Heroku URL because as far as we know that refuses to run anyway (well, I just verified that it's still broken).


philipbelesky:
Looks/works good on my end!

philipbelesky (approved):

docs/install/osx.rst
```diff
@@ -124,49 +125,45 @@ b. Copy **settings/local.example** to **settings/local.py**. Find this part in y
 
     TIME_ZONE = 'Australia/Melbourne'
 
-c. Ensure you are in the main Tabbycat directory (not the config folder where **settings_local.py** is and start a new virtual environment. We suggest the name ``venv``, though it can be any name you like::
+c. Ensure you are in the main Tabbycat directory (not the config folder where **settings_local.py** is). IInstall the Python packages specified in the Pipfile using `Pipenv <https://pipenv.pypa.io/en/latest/>`_ (this also creates a virtual environment), and install the Node.js packages specified in package.json using `npm`::
```
tienne-B, line 128:
```suggestion
c. Ensure you are in the main Tabbycat directory (not the config folder where **settings_local.py** is). Install the Python packages specified in the Pipfile using `Pipenv <https://pipenv.pypa.io/en/latest/>`_ (this also creates a virtual environment), and install the Node.js packages specified in package.json using `npm`::
```
Typo

docs/install/osx.rst
```diff
@@ -124,49 +125,45 @@ b. Copy **settings/local.example** to **settings/local.py**. Find this part in y
...
 
-    $ python3 -m venv venv
+    $ pipenv install
+    $ npm install
```
tienne-B, line 131:
```suggestion
    $ npm ci
```
Ditto

docs/install/linux.rst
```diff
@@ -25,23 +25,24 @@ You need to be familiar with command-line interfaces to get through this comfort
...
   createdb mydatabasename
 
 Then create **settings/local.py** as described :ref:`below <local-settings-linux>`, then::
 
-  python3 -m venv venv
-  source venv/bin/activate
-  pip install --upgrade pip
-  pip install -r ./config/requirements_core.txt
+  pipenv install
   npm install
```
tienne-B, line 41: Should be `npm ci` to install from lock rather than install fresh.

docs/install/linux.rst
```diff
@@ -173,23 +179,12 @@ a. Navigate to your Tabbycat directory::
...
-c. Run the ``activate`` script. This puts you "into" the virtual environment::
-
-    $ source venv/bin/activate
-
-d. Install Tabbycat's requirements into your virtual environment::
-
-    $ pip install --upgrade pip
-    $ pip install -r ./config/requirements_core.txt
+    $ pipenv install
     $ npm install
```
tienne-B, line 185: Ditto

docs/install/linux.rst
```diff
@@ -208,25 +203,29 @@ e. Navigate to the **tabbycat/settings** sub folder and copy **local.example** t
...
-  It should show something like this::
+f. Start Tabbycat!
 
-    serving on http://127.0.0.1:8000
+  ::
+
+    (tabbycat-9BkbSRuB) $ npm run serve
 
-h. Open your browser and go to the URL printed above. (In the above example, it's http://127.0.0.1:8000.) It should look something like the screenshot below. If it does, great! You've successfully installed Tabbycat.
+  Lots of text will flow by---this command starts up all of the processes necessary to run Tabbycat. But the app will be at http://127.0.0.1:8000/ or http://localhost:8000/ (not at any of the other addresses that will show).
```
tienne-B, line 226: Link URLs?

docs/install/windows.rst
```diff
@@ -174,29 +174,25 @@ b. Make a copy of **settings\\local.example** and rename it to **settings\\local
...
+c. Install Pipenv. Then install the Python packages specified in the Pipfile using `Pipenv <https://pipenv.pypa.io/en/latest/>`_ (this also creates a virtual environment), and install the Node.js packages specified in package.json using `npm`::
 
-    > python -m venv venv
-
-d. Run the `Activate.ps1` script. This puts you "into" the virtual environment::
-
-    > .\venv\Scripts\Activate.ps1
+    > pip install --user pipenv
+    > pipenv install
+    > npm install
```
tienne-B, line 181:
```suggestion
    > npm ci
```

tienne-B (commented):
Works well; left a couple comments for the instructions.

docs/install/linux.rst
czlee: they autolink 🙂

docs/install/linux.rst
czlee:
```suggestion
  npm ci
```

docs/install/linux.rst
czlee:
```suggestion
    $ npm ci
```

docs/install/linux.rst
czlee: Hmm, if we do this then I guess `pipenv install` should also be `pipenv install --deploy` everywhere it occurs in the instructions, and also in the Dockerfile. That is, all installation instructions should install from lock.

I think I'm on board with this, all things considered. Installation instructions are at least ostensibly aimed at users, not developers, and advanced users who want to update these can do so themselves (and at their own risk). @philipbelesky, thoughts?

philipbelesky: Sounds good. I don't think we've been bitten by any dependency issues from not using the lock files, but makes sense to be consistent/deterministic in the installs.

Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN git config --global url."https://".insteadOf git://
 
 # Install our node/python requirements
 RUN npm install -g npm@6.14.5
```
tienne-B, line 26:
```suggestion
RUN nvm use
```
We should update `.nvmrc` to the current version, and then this command would read from that and install+use that version.

Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN git config --global url."https://".insteadOf git://
...
-RUN pip install -r ./config/requirements_docker.txt
+RUN pip install pipenv
+RUN pipenv install --system
 RUN npm install --only=production
```
tienne-B, line 29:
```suggestion
RUN npm ci --only=production
```
Would this work?

Dockerfile
czlee: It now crashes because it can't find `nvm`; I'm guessing we also need to install `nvm` or something?

tienne-B: Guess so

philipbelesky: I've just removed NVM as I don't think we need to worry about handling multiple node versions in the context of docker deploys or CI.

PR #2017, opened 2022-04-17
Mainline render deployment option

philipbelesky:
To be merged after #2016. Remaining items:

- [x] Fix local Docker builds
- [x] Fix Docker builds on Github CI
- [ ] Check Sentry works on Render
- [x] Check Heroku deploy script still works
- [ ] Check/update install docs RE: psycopg-binary
- [ ] Migrate to gunicorn or unicorn rather than Daphne (might be a separate issue/PR)

Implementation notes:

- Settings imports have changed to be an if/elif type of structure, so that there should be a more strict separation of imports amongst the different environments
- Render doesn't use nginx in any capacity. I'm tempted to remove it from Heroku and Docker as well to streamline things given Calico is the recommended option for scalable/performant hosting
- I swapped to psycopg-binary as Render doesn't seem to have any qualms about using it in production
- The above changes probably break a few things on Calico, sorry @tienne-B !


.github/README.md
```diff
@@ -19,7 +19,7 @@ Tabbycat is a draw tabulation system for British Parliamentary and a variety of
 
 ## 🔍 Features
 
-- Deployable to [Heroku](https://www.heroku.com/) for an easy, fast, and free setup
+- A range of deployment options — Tabbycat will run on Calico (a paid hosted service), Render or Heroku (free hosted services), and local installations (natively or via Docker)
```
tienne-B, line 22: Wouldn't class Render or Heroku as "free hosted services", but something like "self-deployment" as a differentiator is the deployment obligations, not just the price.

.github/README.md
```diff
@@ -33,13 +33,17 @@ Our user guide is at [tabbycat.readthedocs.io](http://tabbycat.readthedocs.io/).
...
+Tabbycat can be deployed in a number of ways.
 
 [Calico](https://calicotab.com/) is a managed hosting service run by one of Tabbycat's developers. For a flat fee, it will host tab websites, automatically manage their setup and performance, and provide ongoing access to the released tab. Click this button to deploy to Calico:
 
 [![Deploy](https://raw.githubusercontent.com/gist/tienne-B/fc04ecd3c11a38424b642b4bba60e8d9/raw/b2c71d7d6a0d368d3e9dfd8002af729d155ad09b/calicodeploy.svg)](https://calicotab.com/tournaments/new/)
 
-Tabbycat is **currently unavailable as a 1-click installation** on the [Heroku](https://www.heroku.com) web platform. If you are unable to use Calico, we suggest [creating a local installation](https://tabbycat.readthedocs.io/en/stable/install/local.html). We hope to address this issue or [develop an alternate installation method](https://tabbycat.readthedocs.io/en/feature-render/index.html) in a future version of Tabbycat.
+Tabbycat is **no longer available as a 1-click Heroku installation** on the Heroku web platform. If you do not want to use Calico, you have a few alternate options:
```
tienne-B, line 42: Giving up on Heroku I see.

philipbelesky: Yea, even if we did have the 1-click install running again it doesn't seem like it has a reliable future. And, using the local-deploy script seems like a support headache.

.github/README.md
philipbelesky: Revised in 9db7cf1008b1f5aa730e4d76135f045967e31dbd, let me know if you'd like any further edits.

philipbelesky:
Manually merged-in to `develop`

Issue #1928, opened 2021-12-18
[Suggestion] Archiving a tournament


franga2000:
Hi again! This is another one of those suggestions that I intend to implement for our own use and want to ask for feedback here so it can hopefully be merged. *And hopefully I actually do it this time*

**The problem:** after a while of use, old tournaments start piling up and it becomes increasingly annoying to find the active ones (both in the sidebar switcher and the frontend). 

**Proposed solution:** An "archive this tournament" switch in the tournament options that:

1. moves the tournament to a collapsed-by-default "archived" section in the sidebar

![image](https://user-images.githubusercontent.com/3891092/146636908-28a68883-f7af-49ee-9268-2fe0a75aaec2.png)

2. moves their links in the frontend below the non-archived ones

![image](https://user-images.githubusercontent.com/3891092/146636992-fcbf30eb-f3a5-43ae-9ac2-aebb3ceb7a5e.png)

*(screenshots are just mockups)*

Having this flag available would also enable potential performance optimizations if they become necessary (like extending the frontend cache time since archived tournaments could be treated as read-only).

Thoughts?

czlee:
Ah, interesting. So, I think we kind of have this. But maybe it's not as prominent as it could be, or maybe it's poorly named.

Tournaments currently have a field called "active", and it's true by default. If you set it to false, then it should disappear from the main list on the home page, and in #1384, @tienne-B added a page for inactive tournaments. Is this basically the functionality you're suggesting? Maybe we should rename "inactive" to "archived" or something?

Someone might correct me, but from memory to set a tournament as inactive you need to do it via Edit Database—if this is frequently used, potentially it could be brought out to an admin area page in the Configuration section or something.

Thanks so much for these suggestions and willingness to contribute!

franga2000:
Oh wow, that's exactly what I had in mind! It is indeed only available from Edit Database, but that seems pretty reasonable to me. 

I'm tempted to support renaming it to "archived" in the frontend, as while "inactive" is entirely correct, at least to me it doesn't convey the same meaning of "this tournament was active, but is now finished". Then again, I can see some use-cases where the "inactive" tag wouldn't be permanent, in which case "archived" would actually be incorrect (like if a split-timezone tournament wanted to indicate which division is currently active). 

So...looks good as it is, I might still send a PR for the admin dropdown one day if it doesn't break anything, and I'll look around a bit harder next time before opening an issue :sweat_smile:

czlee:
[...]

PR #2053, opened 2022-06-10
Add inactive tournaments to main page

tienne-B:
Rather than hide them in another page, it's simpler to just append to the main page, with a new section header.

Works on the main page for #1928.


philipbelesky:
Thanks for adding this. I made a slight tweak to the presentation — let me know if you'd like to revert/discuss. Otherwise it looks good to merge. 

Screen Shot 2022-06-11 at 6 24 05 pm

PR #2057, opened 2022-06-12
Upgrade to Vue 3

tienne-B:
Starting work on #1502.

PR #2058, opened 2022-06-12
Remove dependency on deprecated vue-touch

tienne-B:
This is a little more complicated than anticipated, and would like a 2nd look. Seems that now the popovers are in a different spot. I also don't really see the difference between "tap" and "click."

This PR will merge into the main Vue 3 upgrade PR (#2057), rather than `develop`.

Work on #1502.


tabbycat/templates/tables/Popover.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 
-  <v-touch class="touch-target" ref="container" v-on:tap="togglePopOver">
```
philipbelesky, line 3: I *think* that when this was changed from being a component to a standard div, the `$el` reference in `createPopper(this.$refs.container.$el` was no longer valid.

philipbelesky (approved):
I haven't had a chance to test this on a touch device, but that's something to look out for in future rounds of testing. Eventually, it would be nice to migrate to the 2.X version of Popper, possibly in conjunction with Bootstrap 5.X

PR #2060, opened 2022-06-20
Add API for speaker points by round per speaker

idowolf:


tabbycat/participants/models.py
```diff
@@ -332,6 +332,26 @@ class Meta:
...
+            return int(s.score)
+        except TypeError:
+            return 0
+
+    @property
+    def rounds(self):
+        """Callers using this property for many teams should prefetch them
+        using `populate_win_counts()` in the `participants.prefetch` module.
+        (That's not a typo -- that function populates both `_wins_count` and
+        `_points`.)"""
```
tienne-B, line 347: Don't think it does any of that...
```suggestion
```

tabbycat/participants/models.py
```diff
@@ -332,6 +332,26 @@ class Meta:
 
     def __str__(self):
         return str(self.name)
+    
+    def get_speaker_points(self, s):
+        try:
+            return int(s.score)
```
tienne-B, line 338: This is not necessarily an integer; the scoring could be decimal or the average of many AdjScores could use them.

tienne-B (commented):
There will be a bit to do, but I've left some preliminary comments (not including linting or branches)

Please also update `api-schema.yml` with your new endpoint.

tabbycat/api/views.py
```diff
@@ -440,12 +440,42 @@ class SubstantiveSpeakerStandingsView(BaseStandingsView):
...
+        serializer.save()
+
+    def get_queryset(self):
+        category_prefetch = Prefetch('categories', queryset=SpeakerCategory.objects.all().select_related('tournament'))
+        if not self.request.user or not self.request.user.is_staff:
+            category_prefetch.queryset = category_prefetch.queryset.filter(public=True)
+
+        return super().get_queryset().prefetch_related(category_prefetch)
+
+
```
tienne-B, line 478: Duplicated.

tabbycat/api/views.py
```diff
@@ -440,12 +440,42 @@ class SubstantiveSpeakerStandingsView(BaseStandingsView):
     def get_max_round(self):
         return self.tournament.round_set.last()
 
+class SubstantiveSpeakerStandingsRoundsView(TournamentAPIMixin, TournamentPublicAPIMixin, ModelViewSet):
+    serializer_class = serializers.SpeakerRoundsSerializer
+    tournament_field = "team__tournament"
+    access_preference = 'public_participants'
+
+    def perform_create(self, serializer):
+        serializer.save()
```
tienne-B, line 449: I don't think we'd want to allow editing through this endpoint.

tabbycat/api/views.py
```diff
@@ -440,12 +440,42 @@ class SubstantiveSpeakerStandingsView(BaseStandingsView):
...
+
+    def get_queryset(self):
+        category_prefetch = Prefetch('categories', queryset=SpeakerCategory.objects.all().select_related('tournament'))
+        if not self.request.user or not self.request.user.is_staff:
+            category_prefetch.queryset = category_prefetch.queryset.filter(public=True)
+
+        return super().get_queryset().prefetch_related(category_prefetch)
```
tienne-B, line 456: Are speaker categories included in the endpoint? If not, no need to filter them

tabbycat/api/views.py
```diff
@@ -440,12 +440,42 @@ class SubstantiveSpeakerStandingsView(BaseStandingsView):
     def get_max_round(self):
         return self.tournament.round_set.last()
 
+class SubstantiveSpeakerStandingsRoundsView(TournamentAPIMixin, TournamentPublicAPIMixin, ModelViewSet):
+    serializer_class = serializers.SpeakerRoundsSerializer
+    tournament_field = "team__tournament"
+    access_preference = 'public_participants'
```
tienne-B, line 446: It's not enough to have access to the participants list to get the tab. The logic may be a bit more complicated, but it should allow viewing with `speaker_tab_released` or for completed rounds that aren't silent.

tienne-B (changes requested):
There's still some work to be done here. Check my previous comments as well and why the build failed.

PR #2130, opened 2022-09-26
Create new field to obscure participant names

tienne-B:
This commit creates a new field for speakers and adjudicators so that they may have code names shown in public views. This feature would allow for privacy in school tournaments and to avoid bias based on names.

Old participants will not have a code name assigned, but new People will have a numerical identifier used as code name.


tabbycat/participants/models.py
```diff
@@ -101,6 +101,9 @@ class Person(models.Model):
     anonymous = models.BooleanField(default=False,
         verbose_name=_("anonymous"),
         help_text=_("Anonymous persons will have their name and team redacted on public tab releases"))
+    code_name = models.CharField(max_length=25, blank=True, null=True,
+        verbose_name=_("code name"),
+        help_text=_("Name used to obscure real name on public-facing pages"))
```
czlee, line 106: What's the intended behaviour if `self.code_name` isn't set? Currently it returns `None`, which can cause TypeErrors in places that are expecting a string. Examples (last two lines of stack trace only):

- when loading path bp88team/admin/participants/adjudicator/1/
  ```
  File "/home/czlee/tabbycat/tabbycat/utils/tables.py", line 591, in add_debate_adjudicators_column
    'text': construct_text(adjs_data),
  File "/home/czlee/tabbycat/tabbycat/utils/tables.py", line 537, in construct_text
    adj_str = '<span class="d-inline">' + a['adj'].get_public_name(self.tournament)
  ```
- when loading path bp88team/participants/adjudicator/1/:
  ```
  File "/home/czlee/tabbycat/tabbycat/utils/tables.py", line 913, in add_debate_results_columns
    cell = self._team_cell(team, show_emoji=False, subtext=subtext)
  File "/home/czlee/tabbycat/tabbycat/utils/tables.py", line 287, in _team_cell
    cell['popover']['content'].append({'text': ", ".join(speakers)})
  ```

I don't think we need to support cases where code names aren't properly set of course (or I guess more precisely, we definitely shouldn't fall back to real name!), but would it be easier to leave out `null=True` and just use `blank=True`, so that if a code name isn't set, it reverts to the empty string?

tabbycat/participants/models.py
```diff
@@ -125,6 +128,11 @@ class Meta:
     def __str__(self):
         return str(self.name)
 
+    def get_public_name(self, tournament):
+        if tournament.pref('participant_code_names') == 'off':
+            return self.name
+        return self.code_name
```
czlee, line 134: Passing in `tournament` is just a performance optimization, right? Since it can always just be inferred from `adj.tournament`? If so, are there any calls where this performance optimization doesn't do very much and we could do something like
```suggestion
    def get_public_name(self, tournament=None):
        if tournament is None:
          tournament = self.tournament
        if tournament.pref('participant_code_names') == 'off':
            return self.name
        return self.code_name
```
to simplify some calls?

(Or is it not a valid assumption that `adj.tournament is not None`? I can't remember if we still permit adj sharing - I have some vague recollection of deprecating this feature, and can't find `share_adjs` in the code base, but also `Adjudicator.tournament` is still nullable.)

(Also... I can't remember why I didn't use a function like this for team codes 😂)

tabbycat/utils/templatetags/debate_tags.py
```diff
@@ -183,6 +183,12 @@ def roundurl(parser, token):
     return RoundURLNode(*args)
 
 
+@register.simple_tag
+def person_display_name(person, tournament):
+    # If returned directly from the object it will have to lookup tournament
+    return person.get_public_name(tournament)
```
czlee, line 189: Similarly - is it possible to get the `tournament` from `context`, in the same way that the `tournament_url` template tag does? (I'm not sure what the answer is... I mean, I'm sure there's a way, but I'm not sure how painful it'd be.) I notice that other analogous tags don't do this, again not sure if there's a reason for it or if we just didn't realise it was an idea at the time.)

tabbycat/participants/utils.py
```diff
@@ -48,3 +51,18 @@ def get_side_history(teams, sides, seq):
     ).annotate(
         **annotate_side_count_kwargs(sides, seq))
     return {team.id: [getattr(team, '%s_count' % side) for side in sides] for team in queryset}
+
+
+def populate_code_names(people, length=8, num_attempts=10):
+    """Populates the URL key field for every instance in the given QuerySet."""
```
czlee, line 57: URL key? is this an oversight?

tabbycat/participants/utils.py
```diff
@@ -48,3 +51,18 @@ def get_side_history(teams, sides, seq):
...
+    chars = string.digits
+
+    existing_keys = list(Person.objects.exclude(code_name__isnull=True).values_list('code_name', flat=True))
+    for person in people:
+        for i in range(num_attempts):
+            new_key = generate_identifier_string(chars, length)
+            if new_key not in existing_keys:
+                person.url_key = new_key
```
czlee, line 65: ditto

czlee (approved):
Looks great. Left a few suggestions / comments on possible oversights.

I wonder how long it'll be before someone asks to be able to turn code names for speakers, and code names for adjudicators, on/off independently of each other 😉 

I notice that a lot of the admin pages also apply code names to participants when this is enabled, and that there are only two settings (as opposed to the five for team code names). I guess when we implemented team code names there were a lot of logistical concerns in the community that made us nervous about showing team code names everywhere, so we included options to keep the admin view with real names. What's the current state of thinking on this nowadays?

tabbycat/participants/models.py
tienne-B: It can't really always be inferred by `adj.tournament` as it also affects speakers, who would then be `spk.team.tournament`.

tabbycat/utils/templatetags/debate_tags.py
tienne-B: Yep, with the `takes_context` decorator argument!

tabbycat/participants/models.py
tienne-B: True

tienne-B:
Yes, there are two ways I see this feature being extended:
* Separate Adj and Speaker preferences
* Better admin indication of real names

However I think this works well as an MVP and for what was asked for. I'm currently thinking about how to show real names, especially in places that do not have tooltips, like in ballot views...

tabbycat/participants/models.py
czlee: Ah, but then same idea applies? Either add logic to tell between the two (using duck typing or whatever) or add
``` python
    @property
    def tournament
        return self.team.tournament
```
to `Speaker`.

Anyway, I'm not too fussed about this, I can see how with two cases it becomes a little clunky. Can always add later if it becomes more bothersome.

PR #2145, opened 2022-10-10
Fix dead links and change 2020 to 2022

### LINKED ISSUE #2144: README.md links to dead tab
The first link in `README.md`, which is supposed to go to the WUDC 2020 tab. This tab has, however, gone offline. It should probably be changed to a more recent example.

daankoning:
Closes #2144

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -13,7 +13,7 @@
 
 </div>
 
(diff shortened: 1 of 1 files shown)


tienne-B:
Looks good; thanks!

PR #2168, opened 2022-11-19
Remove "mark debate as status" dropdown actions

tienne-B:
This commit removes the status dropdown actions from the database area for Debates as that field is modifiable more intuitively with the drop-downs available on each row.


czlee (approved):
Haven't checked the implementation but the idea and motivation sound fine to me.

Issue #1464, opened 2020-05-25
Add type hints for the codebase.


teymour-aldridge:
One thing I think would make it substantially easier to refactor code and catch bugs which aren't possible to catch otherwise. Having types makes it possible to use the wealth of code analysis tools which are out there, in addition.
I'd be happy to work on this.

czlee:
Sure, I'd support seeing type hints added where useful. It'll be a lot of work, of course, so it's fine if the transition is gradual, like a lot of refactoring work we've done on the Tabbycat code base.

teymour-aldridge:
I think it's important to have 100% coverage. My justification for this comes from the Reason (OCaml <=> Javascript interop built by Facebook) which puts it more nicely than I could:

> Many folks who come from a gradually typed community (for example, having experienced using a type system on top of JavaScript, Ruby, Python, etc.) don't understand the fuss about Reason's types "soundness". In reality, there's a big difference between these type systems that are "best effort", aka 99.9% correct, and our type system that's "sound", aka 100% correct.

> Many claim that "99% is good enough". While this could be true in other domains, such claim isn't very valid for code. Naive probability: If your code has 100 types, 99.9% probability of correctness per type means that overall, that code's correctness is 99.9% ^ 100 = 90.4%

> Oops. That's potentially lots of false positives in the system's reported types now. Which ones are wrong though? That type system sure isn't telling you ¯\_(ツ)_/¯.

teymour-aldridge:
I've been doing some work on Werkzeug (a Python WSGI framework) recently, and I think it will be possible to generate type annotations from Tabbycat's tests which can then be refined. I'll try and get round to this soon-ish.

daankoning:
@teymour-aldridge still intend to do this? If not I'd be happy to begin a bit of work on this.

czlee:
@daankoning Sorry I haven't had a chance to reply sooner. You're welcome to work on this if you're particularly interested in it, but I should say, I think there are much more egregious areas of technical debt in Tabbycat than type hints, examples:
- Our unit test coverage is currently pretty bad.
- There are some classes that have outgrown themselves and need to be refactored for DRY and separation of concerns. (Our tables generators, in files named tables.py, come to mind.)
- Our action logs are a bit ad hoc and not as useful as they could be.
- Things like #500

Happy to discuss what you're interested in! I just wanted to say it's probably not the most pressing thing on our backlog.

daankoning:
[...]

PR #2201, opened 2022-12-17
Add a couple of type hints

daankoning:
As discussed in #1464. 

This PR only adds hints to the files in `tabbycat/notifications` and `tabbycat/privateurls` but mostly serves as a proof of concept and to ensure everyone agrees with the way I am going about these (such as when to use `Any`).

Unless any objections are raised I'm happy to continue work on the rest of the codebase in similar fashion.

### DIFF
--- a/tabbycat/notifications/admin.py
+++ b/tabbycat/notifications/admin.py
@@ -1,3 +1,5 @@
+from typing import Any, Callable, Type, TYPE_CHECKING
+
 from django.contrib import admin
 from django.utils import timezone
 
@@ -5,8 +7,12 @@ from utils.admin import ModelAdmin, TabbycatModelAdminFieldsMixin
 
 from .models import BulkNotification, EmailStatus, SentMessage
 
+if TYPE_CHECKING:
+    from django.db.models import QuerySet
+    from django.core.handlers.wsgi import WSGIRequest
 
-def precise_timestamp_isoformat(model, field_name):
+
+def precise_timestamp_isoformat(model: Type, field_name: str) -> Callable[[Any], str]:
     @admin.display(description=model._meta.get_field(field_name).verbose_name)
     def precise_timestamp(self, obj):
         return timezone.localtime(getattr(obj, field_name)).isoformat()
@@ -20,7 +26,7 @@ class SentMessageAdmin(TabbycatModelAdminFieldsMixin, ModelAdmin):
     search_fields = ('message_id', 'recipient__name', 'email', 'recipient__email')
     ordering = ('-notification',)
 
-    def get_queryset(self, request):
+    def get_queryset(self, request: 'WSGIRequest') -> 'QuerySet[SentMessage]':
         return super().get_queryset(request).select_related('recipient', 'notification__tournament')
 
     precise_timestamp = precise_timestamp_isoformat(SentMessage, 'timestamp')
@@ -32,7 +38,7 @@ class BulkNotificationAdmin(TabbycatModelAdminFieldsMixin, ModelAdmin):
     list_filter = ('tournament', 'round', 'event')
     ordering = ('-timestamp',)
 
-    def get_queryset(self, request):
+    def get_queryset(self, request: 'WSGIRequest') -> 'QuerySet[BulkNotification]':
         return super().get_queryset(request).select_related('round__tournament', 'tournament')
 
     precise_timestamp = precise_timestamp_isoformat(BulkNotification, 'timestamp')
--- a/tabbycat/notifications/admin.py
+++ b/tabbycat/notifications/admin.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
     from django.http.request import HttpRequest
 
 
-def precise_timestamp_isoformat(model: Type, field_name: str) -> Callable[['Model'], str]:
+def precise_timestamp_isoformat(model: Type['Model'], field_name: str) -> Callable[['Model'], str]:
     @admin.display(description=model._meta.get_field(field_name).verbose_name)
     def precise_timestamp(self, obj):
         return timezone.localtime(getattr(obj, field_name)).isoformat()
--- a/tabbycat/notifications/admin.py
+++ b/tabbycat/notifications/admin.py
@@ -9,7 +9,7 @@ from .models import BulkNotification, EmailStatus, SentMessage
 
 if TYPE_CHECKING:
     from django.db.models import QuerySet
-    from django.core.handlers.wsgi import WSGIRequest
+    from django.http.request import HttpRequest
 
 
 def precise_timestamp_isoformat(model: Type, field_name: str) -> Callable[[Any], str]:
@@ -26,7 +26,7 @@ class SentMessageAdmin(TabbycatModelAdminFieldsMixin, ModelAdmin):
     search_fields = ('message_id', 'recipient__name', 'email', 'recipient__email')
     ordering = ('-notification',)
 
-    def get_queryset(self, request: 'WSGIRequest') -> 'QuerySet[SentMessage]':
+    def get_queryset(self, request: 'HttpRequest') -> 'QuerySet[SentMessage]':
         return super().get_queryset(request).select_related('recipient', 'notification__tournament')
 
     precise_timestamp = precise_timestamp_isoformat(SentMessage, 'timestamp')
@@ -38,7 +38,7 @@ class BulkNotificationAdmin(TabbycatModelAdminFieldsMixin, ModelAdmin):
     list_filter = ('tournament', 'round', 'event')
     ordering = ('-timestamp',)
 
-    def get_queryset(self, request: 'WSGIRequest') -> 'QuerySet[BulkNotification]':
+    def get_queryset(self, request: 'HttpRequest') -> 'QuerySet[BulkNotification]':
         return super().get_queryset(request).select_related('round__tournament', 'tournament')
 
     precise_timestamp = precise_timestamp_isoformat(BulkNotification, 'timestamp')
--- a/tabbycat/notifications/admin.py
+++ b/tabbycat/notifications/admin.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Type, TYPE_CHECKING
+from typing import Callable, Type, TYPE_CHECKING
 
 from django.contrib import admin
 from django.utils import timezone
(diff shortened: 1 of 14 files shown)


tabbycat/privateurls/views.py
```diff
@@ -99,15 +105,15 @@ def get_adjudicators_table(self):
...
+    def get_tables(self) -> List[TabbycatTableBuilder]:
         return [self.get_adjudicators_table(), self.get_speakers_table()]
 
 
 class GenerateRandomisedUrlsView(AdministratorMixin, TournamentMixin, PostOnlyRedirectView):
 
     tournament_redirect_pattern_name = 'privateurls-list'
 
-    def post(self, request, *args, **kwargs):
+    def post(self, request: Any, *args, **kwargs) -> 'HttpResponseRedirect':
```
daankoning, line 116: The hint on `request` conceivably be set to WSGIRequest but given that it goes unused in the function body it doesn't matter that much.

daankoning: A similar thing could be set for most of the usages of bare `Any` hints, they could be set to more specific hints but they tend to go unused anyways.

tabbycat/notifications/utils.py
```diff
@@ -30,23 +37,23 @@
...
+def _check_in_to(pk, to_ids) -> bool:
     try:
         to_ids.remove(pk)
     except KeyError:
         return False
     return True
 
 
-def adjudicator_assignment_email_generator(to, url, round):
+def adjudicator_assignment_email_generator(to: 'QuerySet[Person]', url: str, round: 'Round') -> List[Tuple[Dict[str, str], 'Person']]:
```
tienne-B, line 56: This is ugly... The generators could use dataclasses instead, which would then show what properties (and types) would be present for each.

As an example, the dataclass for this would be like:
```py
from dataclasses import dataclass

@dataclass
class AdjudicatorAssignment:
    ROUND: str
    VENUE: str
    PANEL: str
    DRAW: str
    USER: str
    POSITION: str
    URL: str
```
etc. It would be more consequential for the more complicated generators though.

tienne-B (commented):
Left some inline comments, but there are a few more general comments.

Right now, there isn't much use for these type hints in that they are not being checked. The code just isn't really set up for them. As a prerequisite to adding type hints, checking their validity should be implemented in our build and linter. This would require installing `mypy` or an equivalent and adding it to our pre-commit hook.

Your work here has shown that we used general types (like `dict`) more than making new types. The work here could be used as a jumping off point for using more specific data structures.

tabbycat/privateurls/views.py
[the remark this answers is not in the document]
daankoning: That was my initial hunch as well but prodding around and a couple of `logger.info(type(request)`'s revealed this.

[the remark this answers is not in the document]
tienne-B: `WSGIRequest` is a subclass of `HttpRequest`, and it's possible that pages eventually get served with ASGI so it'd be good to keep it as an agnostic request.

tabbycat/notifications/models.py
```diff
@@ -110,25 +110,24 @@ class EmailStatus(models.Model):
...
-        (EVENT_TYPE_DELIVERED, _("Delivered")),
-        (EVENT_TYPE_BOUNCED, _("Bounced")),
-        (EVENT_TYPE_OPENED, _("Opened")),
-        (EVENT_TYPE_CLICKED, _("Clicked")),
-        (EVENT_TYPE_UNSUBSCRIBED, _("Unsubscribed")),
-        (EVENT_TYPE_SPAM, _("Marked as spam")),
-        (EVENT_TYPE_ASM_UNSUBSCRIBED, _("Unsubscribed from group")),
-        (EVENT_TYPE_ASM_RESUBSCRIBED, _("Resubscribed to group")),
-    )
+    class EventType(models.TextChoices):
```
tienne-B, line 113: Why'd you keep the `EVENT_TYPE_` attributes above? Not needed anymore with this `enum`.

tabbycat/notifications/consumers.py
```diff
@@ -30,17 +38,17 @@ class NotificationQueueConsumer(SyncConsumer):
...
 
-    def _get_from_fields(self, t):
+    def _get_from_fields(self, t: Tournament) -> Tuple[str, Optional[List[str]]]:
         from_email = formataddr((t.short_name, settings.DEFAULT_FROM_EMAIL))
         if t.pref('reply_to_address'):
             return from_email, [formataddr((t.pref('reply_to_name'), t.pref('reply_to_address')))]
         return from_email, None  # Shouldn't have array of None
 
-    def email(self, event):
+    def email(self, event: Dict[str, Union[str, List[int], Dict[str, Union[Round, Tournament, Debate, Any]]]]) -> None:
```
tienne-B, line 51: A `Union` with `Any` is just `Any` anyways. Ugly typing.

tabbycat/notifications/consumers.py
```diff
@@ -17,9 +18,16 @@
...
+if TYPE_CHECKING:
+    from django.db.models import QuerySet, Model
+
 
 class NotificationQueueConsumer(SyncConsumer):
 
+    NOTIFICATION_GENERATORS: Dict[str, Union[
+        Callable[['QuerySet[Person]', str, 'Model'], List[Tuple[Dict[str, str], Person]]],
+        Callable[['QuerySet[Person]', 'Model'], List[Tuple[Dict[str, str], Person]]],
+    ]]
```
tienne-B, line 30: Ugly typing, but also `BulkNotification.EVENT_TYPE_` should also be an `enum` (like email status types).

tabbycat/privateurls/views.py
```diff
@@ -149,12 +156,12 @@ class EmailRandomisedUrlsView(RoleColumnMixin, TournamentTemplateEmailCreateView
 
     tournament_redirect_pattern_name = 'privateurls-list'
 
-    def get_extra(self):
+    def get_extra(self) -> Dict[str, Union[int, str]]:
```
tienne-B, line 159: I think I'd use `Dict[str, Any]` here as we don't care that it's only `int` or `str`.

tabbycat/privateurls/utils.py
```diff
@@ -24,6 +28,6 @@ def populate_url_keys(people, length=8, num_attempts=10):
     Person.objects.bulk_update(people, ['url_key'])
 
 
-def delete_url_keys(queryset):
+def delete_url_keys(queryset: 'QuerySet') -> None:
```
tienne-B, line 31: URL keys are only on `Person`, so the type for `QuerySet` could be made more specific.

tabbycat/notifications/views.py
```diff
@@ -343,21 +349,21 @@ def form_valid(self, form):
 
 class TournamentTemplateEmailCreateView(TemplateEmailCreateView):
 
-    def get_default_send_queryset(self):
+    def get_default_send_queryset(self) -> 'QuerySet[Person]':
         return super().get_default_send_queryset().exclude(
             sentmessage__notification__event=self.event, sentmessage__notification__tournament=self.tournament)
 
-    def get_extra(self):
+    def get_extra(self) -> Dict[str, int]:
```
tienne-B, line 356: As used elsewhere, think `Any` should be used. It's odd to have an extending method (`EmailRandomisedUrlsView.get_extra()`) with a different return type.

Also for `RoundTemplateEmailCreateView`.

tabbycat/notifications/views.py
```diff
@@ -283,7 +289,7 @@ def get_table(self, mixed_participants=True):
 
         return table
 
-    def get_context_data(self, **kwargs):
+    def get_context_data(self, **kwargs) -> Dict[str, Union[str, bool, List, BasicEmailForm, 'CustomEmailCreateView']]:
```
tienne-B, line 292: `Dict[str, Any]` here (as for all `get_context_data()` methods)

tienne-B:
Feel free to mark comments as resolved once addressed (make the page less long!)

tabbycat/notifications/consumers.py
tienne-B: Did a bit of research; a `Protocol` might be in order for this:
```py
from typing import Any, List, Protocol, Tuple

class NotificationGenerator(Protocol):
    def __call__(self, *vals: Any) -> List[Tuple[EmailContextData, Person]]: 
```

Then, the type annotation for `NOTIFICATION_GENERATORS` could be (including my other suggestions):

```py
NOTIFICATION_GENERATORS: Dict[BulkNotification.EventType, NotificationGenerator]
```
which is much clearer yet flexible.

tabbycat/notifications/admin.py
```diff
@@ -1,12 +1,18 @@
...
 
 from .models import BulkNotification, EmailStatus, SentMessage
 
+if TYPE_CHECKING:
+    from django.db.models import QuerySet
+    from django.http.request import HttpRequest
+
 
-def precise_timestamp_isoformat(model, field_name):
+def precise_timestamp_isoformat(model: Type, field_name: str) -> Callable[[Any], str]:
```
tienne-B, line 15: Would the type for `model` be a `Model`? Unsure about using `Type`. Also, `obj` from line 17 would be a model object, so why `Any`?

daankoning: My reasoning is that `precise_timestamp_isoformat` doesn't take a specific model instance but rather a type itself and then spits out a function based on that. For example, on line 44 it gets called like this: `precise_timestamp_isoformat(BulkNotification, 'timestamp')`, with the `BulkNotification` class, not an instance of `BulkNotification`.

tienne-B (commented):
Forgot to post these

tabbycat/notifications/consumers.py
```diff
@@ -91,6 +92,8 @@ def email(self, event: Dict[str, Union[str, List[int], Dict[str, Any]]]) -> None
         messages = []
         records = []
         for instance, recipient in data:
+            instance = asdict(instance)
```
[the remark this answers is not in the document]
daankoning: In general, I agree, but in this specific case I feel it makes sense to cast it to a dictionary.  None of `instance`'s attributes are ever accessed. It is only ever passed to other constructors, which expect dictionaries. As such, all that not casting it to a `dict` would achieve is having two instances of `asdict(instance)` instead of one as is the case now.

tienne-B:
I think I'll explore this locally a bit to concretize what'd I be looking for with type hints, because it may be taken a bit too far here making logical changes just for a non-semantic syntax feature. I also urge you to include `mypy` in a PR (or this one) so that they can be tested within the build.

daankoning:
I'd love to but the problem (which might very well be solveable but which I am not knowledgeable enough to solve) is that django's typing is somewhat bad and poorly type hinted which makes it very hard for mypy to work as django either provides no stubs or has functions that can return anything such as `filter`. 

This means that another dependency such as [django stubs](https://github.com/typeddjango/django-stubs) would be necessary. I woud be happy to do that but I'd like your thoughts before introducing external dependencies to a project that is not mine.

tienne-B:
Good point about mypy; we can hold off on that. Also, don't hesitate to add libraries if you feel that'd be appropriate! The project is all of ours, not really much "ownership."

I've added a couple commits that try to make it a little neater. What do you think?

daankoning:
Looks good to me!

PR #2205, opened 2022-12-21
Change in displaying panel (again)

linh-trant:
Adjusting the "," position

### DIFF
--- a/tabbycat/templates/allocations/DragAndDropDebate.vue
+++ b/tabbycat/templates/allocations/DragAndDropDebate.vue
@@ -69,9 +69,9 @@
     </slot>
     <slot name="adjudicators">
       <div class="flex-16 align-self-center p-2 small">
-        <span v-for="adj in debateOrPanel.adjudicators.C">{{ adj.name }} Ⓒ, </span>
-        <span v-for="adj in debateOrPanel.adjudicators.P">{{ adj.name }}, </span>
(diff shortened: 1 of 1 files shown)


philipbelesky:
Hey, if we are making/reviewing changes that affect the UI, can we please make sure to verify the output? This causes a regression the Edit Venues screen that I would have thought would be pretty noticeable:

CleanShot 2022-12-24 at 10 12 33@2x

I'm not sure what the exact issue was that this was trying to resolve, I assume it was trailing commas with solo chairs? I've made ae6ccb0811913f0b21197fe99d51fe669babbc80 that should check for the different combinations more effectively.

tienne-B:
> Hey, if we are making/reviewing changes that affect the UI, can we please make sure to verify the output? This causes a regression the Edit Venues screen that I would have thought would be pretty noticeable:

Feel that you're being excessively harsh here; it was better than the status quo where there's a trailing comma at the end of all rows (not just solo chairs). Further, it is not an unreasonable assumption that adjudicators are (near-)finalized when on the Edit Venues screen, and so would all have chairs.

philipbelesky:
That is fair, and I'd like to apologise to the both of you. I think I had way under-estimated the obviousness of the edge-cases here. And, regardless, issues with QA should be directed at maintainers and definitely not towards new contributors!

PR #2215, opened 2023-01-06
Add site timezone to the API

### LINKED ISSUE #2091: Expose site timezone in API
Could be at the root level.

daankoning:
I currently have it set to be read only (in addition to the view only having a get method), is this the intended behaviour?

Closes #2091

### DIFF
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -141,6 +141,7 @@ class V1RootSerializer(serializers.Serializer):
         institutions = serializers.HyperlinkedIdentityField(view_name='api-global-institution-list')
 
     _links = V1LinksSerializer(source='*', read_only=True)
+    site_timezone = serializers.CharField(allow_blank=False, trim_whitespace=True, read_only=True)
 
 
 class CheckinSerializer(serializers.Serializer):
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -132,6 +132,7 @@ class RootSerializer(serializers.Serializer):
         v1 = serializers.HyperlinkedIdentityField(view_name='api-v1-root')
 
     _links = RootLinksSerializer(source='*', read_only=True)
+    timezone = serializers.CharField(allow_blank=False, read_only=True)
     version = serializers.CharField()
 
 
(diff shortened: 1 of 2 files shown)


tabbycat/api/serializers.py
```diff
@@ -141,6 +141,7 @@ class V1LinksSerializer(serializers.Serializer):
         institutions = serializers.HyperlinkedIdentityField(view_name='api-global-institution-list')
 
     _links = V1LinksSerializer(source='*', read_only=True)
+    site_timezone = serializers.CharField(allow_blank=False, trim_whitespace=True, read_only=True)
```
tienne-B, line 144: How about name it just `timezone`? Also, why use `trim_whitespace=True`? It's also only used for schema generation...

tabbycat/api/views.py
```diff
@@ -80,11 +80,13 @@ def get(self, request, format=None):
         """Entrypoint for version 1 of the API"""
         tournaments_create_url = reverse('api-tournament-list', request=request, format=format)
         institution_create_url = reverse('api-global-institution-list', request=request, format=format)
+        timezone = settings.TIME_ZONE
         return Response({
             "_links": {
                 "tournaments": tournaments_create_url,
                 "institutions": institution_create_url,
             },
+            "site_timezone": timezone,
```
tienne-B, line 89:
```suggestion
            "timezone": settings.TIME_ZONE,
```
I'd be fine with combining those two lines as they are short.

tabbycat/api/views.py
```diff
@@ -80,11 +80,13 @@ def get(self, request, format=None):
         """Entrypoint for version 1 of the API"""
         tournaments_create_url = reverse('api-tournament-list', request=request, format=format)
         institution_create_url = reverse('api-global-institution-list', request=request, format=format)
+        timezone = settings.TIME_ZONE
```
tienne-B, line 83:
```suggestion
```

tienne-B (approved):

tienne-B:
> I currently have it set to be read only (in addition to the view only having a get method), is this the intended behaviour?

Yes, as the endpoint as a whole is read-only. The serializer object is only used by DRF to generate the schema. The timezone is a Django setting filled by an environment variable, so it cannot be modified from within the app.

PR #2217, opened 2023-01-08
Escape adjudicator name in allocation

linh-trant:


tienne-B (approved):
I'd want to make a better way than adding a method field, but it doesn't seem obvious in DRF...

PR #2218, opened 2023-01-08
Hide "Enter ballot" button when adjudicator isn't adjudicating that round.

### LINKED ISSUE #2096: Private URL: Don't show "Enter Ballot" for rounds where adj not judging

daankoning:
Closes #2096

### DIFF
--- a/tabbycat/privateurls/templates/public_url_landing.html
+++ b/tabbycat/privateurls/templates/public_url_landing.html
@@ -58,15 +58,10 @@
     {% endif %}
 
     {% if object.adjudicator and pref.participant_ballots == 'private-urls' %}
-      {% for round in tournament.current_rounds %}
-          {% if object.adjudicator in round.used_adjudicators %}
-            {% roundurl 'results-public-ballotset-new-randomised' round url_key as url %}
-            {% blocktrans trimmed with round=round.name asvar text %}Submit Ballot for {{ round }}{% endblocktrans %}
-            {% include "components/item-action.html" %}
-          {% endif %}
-      {% empty %}
-        {% trans "There are no current rounds" as text %}
-        {% include "components/item-info.html" with type='info' %}
+      {% for dadj in debateadjudications %}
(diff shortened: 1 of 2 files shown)


tabbycat/privateurls/templates/public_url_landing.html
```diff
@@ -57,7 +57,7 @@
       {% endif %}
     {% endif %}
 
-    {% if object.adjudicator and pref.participant_ballots == 'private-urls' %}
+    {% if object.adjudicator and pref.participant_ballots == 'private-urls' and debateadjudications|length > 0 %}
```
tienne-B, line 60: Would it be possible to only show the current rounds for which they can submit ballots, rather than all current rounds if they're in at least one?

tienne-B (changes requested):
This is not the correct fix; we already pass a list of the current rounds the adjudicator is judging on for the "In this round" box (as `debateadjudications`). With that, we could do:

```
{% if object.adjudicator and pref.participant_ballots == 'private-urls' %}
  {% for dadj in debateadjudications %}
      {% roundurl 'results-public-ballotset-new-randomised' dadj.debate.round url_key as url %}
      {% blocktrans trimmed with round=dadj.debate.round.name asvar text %}Submit Ballot for {{ round }}{% endblocktrans %}
      {% include "components/item-action.html" %}
  {% endfor %}
{% endif %}
```

Note that I've also removed the `empty` clause as it doesn't display well; the message is ambiguous, and it's use is dubious to indicate that no action has to be taken, as the "In this round" box would have a clearer message.

You won't need to modify any other file than the landing page template.

daankoning:
Hi, sorry for taking a while to get back to this. I had exams.

When you say you don't need me to modify any file other than the landing page template does that mean from the PR as it stands or in general (i.e. do you want the `used_adjudicators` function to also be removed)? The latter was my interpretation, please point it out if you meant anything else :smile:.

PR #2220, opened 2023-01-09
Escape adjudicator name in room allocation

linh-trant:


philipbelesky:
Good catch, do you think there are other fields that might need similar treatment? Team or institution names perhaps?

linh-trant:
I think the same thing should be done with Team names, and I can get my hand on that.

tienne-B:
@philipbelesky Yes, but we should maybe find a more sustainable solution than making lots of fields `MethodField`s.

PR #2221, opened 2023-01-09
Warn about releasing unset motions

### LINKED ISSUE #1051: Releasing motion to the public should warn if no motion entered
On 2.3.0A (deployed yesterday) there is no warning if you release a motion without a motion actually being entered for the round under WSDC motion entry settings. (except the default warning on the enter a motion section)

image

image

I think the preferred functionality should either be that it prevents you from releasing the motion if there is no motion entered or that there should be a warning message added to the "Released the motion(s)" message or alternatively to the release draw textbox

daankoning:
There are two warnings: one that appears above the button to release the motions and one that appears after an empty motion set is released.

closes #1051

### DIFF
--- a/tabbycat/draw/templates/draw_display_admin.html
+++ b/tabbycat/draw/templates/draw_display_admin.html
@@ -211,8 +211,11 @@
 
           <form id="releaseMotionsForm" method="POST" hidden
                 action="{% roundurl 'motions-release' %}">{% csrf_token %}</form>
-
-          {% if pref.participant_ballots != 'off' and pref.enable_motions %}
+          {% if not round.motion_set.exists %}
+            {% trans "No motion(s) are currently set for this round." as text %}
+            {% include "components/item-info.html" with nopad=True type="danger" %}
+            {% include "components/item-action.html" with type="success" id="triggerReleaseMotionsForm" text=release_text url="" to_complete=True %}
+          {% elif pref.participant_ballots != 'off' and pref.enable_motions %}
             {% trans "Your tournament is configured to require motions to be selected in ballots and to allow ballots to be submitted by adjudicators. Ensure that you release the motions before debates finish, otherwise ballots will not be able to be submitted." as text %}
             {% include "components/item-info.html" with nopad=True type="danger" %}
             {% include "components/item-action.html" with type="success" id="triggerReleaseMotionsForm" text=release_text url="" to_complete=True %}
@@ -366,7 +369,7 @@
 {% block js %}
   {{ block.super }}
   <script>
-    $(document).ready( function() { // Use the fake submit buttons as real submissoin
+    $(document).ready( function() { // Use the fake submit buttons as real submission
       $("#triggerReleaseDrawForm").click( function() {
         $("#releaseDrawForm").submit();
         return false;
--- a/tabbycat/draw/templates/draw_display_admin.html
+++ b/tabbycat/draw/templates/draw_display_admin.html
@@ -192,15 +192,21 @@
   <div class="col-md col-12">
     <div class="card">
 
-      {% if pref.enable_motions or pref.motion_vetoes_enabled %}
-        {% trans "Release Motions" as release_motions_title %}
-        {% trans "Release motions to public" as release_text %}
-        {% trans "Unrelease motions to public" as unrelease_text %}
-      {% else %}
-        {% trans "Release Motion" as release_motions_title %}
-        {% trans "Release motion to public" as release_text %}
-        {% trans "Unrelease motion to public" as unrelease_text %}
-      {% endif %}
+      {% blocktranslate count counter=round.motion_set|length asvar release_motions_title%}
+        Release Motion
+      {% plural %}
+        Release Motions
+      {% endblocktranslate %}
+      {% blocktranslate count counter=round.motion_set|length asvar release_text%}
+        Release motion to public
+      {% plural %}
+        Release motions to public
+      {% endblocktranslate %}
+      {% blocktranslate count counter=round.motion_set|length asvar unrelease_text%}
+        Unrelease motion to public
+      {% plural %}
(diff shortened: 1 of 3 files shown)


tabbycat/draw/templates/draw_display_admin.html
```diff
@@ -211,8 +211,11 @@ <h5 class="card-title mb-0">{{ release_motions_title }}</h5>
 
           <form id="releaseMotionsForm" method="POST" hidden
                 action="{% roundurl 'motions-release' %}">{% csrf_token %}</form>
-
-          {% if pref.participant_ballots != 'off' and pref.enable_motions %}
+          {% if not round.motion_set.exists %}
+            {% trans "No motion(s) are currently set for this round." as text %}
+            {% include "components/item-info.html" with nopad=True type="warning" %}
+            {% include "components/item-action.html" with type="success" id="triggerReleaseMotionsForm" text=release_text url="" to_complete=True %}
```
tienne-B, line 217: Either we should try to avoid duplicating this button (as it is the same in the `elif`), or it should be in a warning state and indicate the problem (and why it is) in the text.

tabbycat/motions/views.py
```diff
@@ -157,7 +157,15 @@ def post(self, request, *args, **kwargs):
...
-        messages.success(request, self.message_text)
+        if round.motion_set.all().exists():  # Means that the round has at least one motion associated with it
+            messages.success(request, self.message_text)
+        elif self.motions_released:  # The (non-existent) motions were released
+            messages.warning(request, _("No motion(s) are set for this round. Are you sure you intended to already "
+                                        "release the motion(s)?"))
+        else:
+            messages.info(request, _("Returned to previous state. Feel free to enter your motion(s) now, they won't be "
+                                     "publicly displayed until you return to this page and release them."))
+
```
tienne-B, line 168: These messages should be refined (if these are even necessary).

I don't think these are necessary, or should be in addition to the success message, as that shows that the button was successfully pressed, which it was.

tienne-B (commented):
I'm unsure about the warning types and messages...

Also something that could be fixed while you're here is pluralization for i18n, where the separate messages like "Release Motions" and "Release Motion" could be merged using `blocktranslate` and `plural`.

I think a better conception of this issue is that if the motions are released without a motion entered, the motion would be released immediately on save. A warning could be added to the "Edit Motions" page.

The warning that there are no motions should stay even if the motions were released.

tabbycat/motions/views.py
tienne-B: To resolve, please indicate why these messages should remain.

daankoning: To me, it seems somewhat intuitive that if the user has released the motions in spite of the warning shown above the button they have either a) not seen the warning, in which case these extra messages should catch this or b) has purposefully chosen to release the motions in spite of this (for reasons unbeknownst to me) so they must clearly have some reason to want to release unset motions. In this case, I'd personally find the message "Motions released successfully" to be somewhat confusing and hence the need to have the second messages.

tienne-B: So how about appending something like "No motions were set" or "Motions will be released on creation" then? These toast messages should not be interrogative and I feel they too are badly worded. Please check my overall comment for how I'd think it should work.

tienne-B:
What would you think of merely disabling the "Release Motions" button if there are no motions set?

philipbelesky:
^ That would be my preference as well. I think we should in general err on not the UI having to explain the weird states it can get into by instead having the UI prevent those states from happening. The edit database area is always there as an escape hatch for when I things are in a weird state.

czlee:
> What would you think of merely disabling the "Release Motions" button if there are no motions set?

> ^ That would be my preference as well. I think we should in general err on not the UI having to explain the weird states it can get into by instead having the UI prevent those states from happening. The edit database area is always there as an escape hatch for when I things are in a weird state.

Yeah, I was originally going to comment with a couple of paragraphs discussing the role of `django.contrib.messages` messages and how they interact with the user (since imho it's not super obvious). But this is a much better idea, and would render basically that entire discussion moot (and least in this context).

daankoning:
Cool, I'll get around to doing that then. Sorry for not reading the initial message that well :(

tabbycat/motions/templates/motions_edit.html
```diff
@@ -44,6 +44,13 @@
     {% include "components/form-errors.html" with errors=formset.non_form_errors %}
   {% endif %}
 
+  {% if round.motions_released %}
+    {% blocktranslate asvar p1 %}
+      The motion(s) for this round are currently publicly displayed because they were released. Any changes made here will be publicly visible as well as soon as they are saved.
+    {% endblocktranslate %}
+    {% include "components/explainer-card.html" with type="info" %}
```
tienne-B, line 51: Maybe as a warning rather than info.

tienne-B (commented):
Looks good!

tabbycat/draw/templates/draw_display_admin.html
```diff
@@ -192,42 +192,56 @@ <h5 class="card-title mb-0">{% trans "Release Draw" %}</h5>
...
-        {% trans "Release Motion" as release_motions_title %}
-        {% trans "Release motion to public" as release_text %}
-        {% trans "Unrelease motion to public" as unrelease_text %}
-      {% endif %}
+      {% blocktranslate count counter=round.motion_set|length asvar release_motions_title%}
+        Release Motion
+      {% plural %}
+        Release Motions
+      {% endblocktranslate %}
+      {% blocktranslate count counter=round.motion_set|length asvar release_text%}
```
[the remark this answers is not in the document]
tienne-B: Careful; instances remain above and below!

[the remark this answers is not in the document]
daankoning: My bad, just went by all the comments :)

PR #2229, opened 2023-01-18
Add richtext to infoslide

linh-trant:


tabbycat/motions/templates/show.html
```diff
@@ -75,7 +75,13 @@
         {% if infos|length == 1 %}
 
           <div class="modal-body d-flex align-items-center justify-content-center text-center">
-            <small>{{ infos.0.motion.info_slide|linebreaks }}</small>
+            <small>
+              {% if infos.0.motion.info_slide|slice:":3" == "<p>" %}
```
philipbelesky, line 79: Is this for backwards compatibility? Or, could we assume all motions going forward using rich text and can be output without using `|linebreaks`? (I may be totally forgetting how this part of django formatting works).

philipbelesky (approved):
This looks good to me, thanks for adding it.

I did make a small change to force left-alignment of info slide modals viewed from the motions tab. This was a pre-existing issue though, where a `text-center` class further up was being inherited accidentally.

philipbelesky:
Actually, nevermind about the text-alignment commit. I couldn't push it to this branch, so I've just added it to `develop` directly as 430798f5b14348b0cfe1188f6f66e792f5ad00b5

philipbelesky:
This should be merged now, but I did so by cherry picking and merging those last two commits as there seemed to be a few unrelated conflicts with the earlier history. Let me know if I messed anything up!

tabbycat/motions/templates/show.html
tienne-B: Yes, but there wouldn't be a way to determine if the motion was created when rich text was added or not, so we can test the text itself for the tag. Also, the presets don't have rich text motions.

PR #2230, opened 2023-01-28
Show the motion reference when there's only one option

### LINKED ISSUE #2224: Show the motion reference when there's only one option
In the adjudicators' private URL, show the motion reference on the table when there's only one option for motion.
image

linh-trant:
Close #2224

### DIFF
--- a/tabbycat/adjallocation/templates/EditEitherAdjudicatorsSharedMixin.vue
+++ b/tabbycat/adjallocation/templates/EditEitherAdjudicatorsSharedMixin.vue
@@ -75,36 +75,42 @@ export default {
       }
       const allocationChanges = []
       const adjudicatorsSetModified = [dragData.item]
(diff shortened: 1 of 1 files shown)


tienne-B:
I'm not really sure about this... we've been trying to move away from being implicit about debates having the round's (sole) motion and towards always associating the motion with the debate (well, ballot submission). Work in that vein includes motion statistics and [this big migration](https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/results/migrations/0009_ballotsubmission_motions.py) to populate the `motion` field on old ballots.

The scope of this fix would then be if the motion gets added after-the-fact, so ballots are entered without an associated motion. Making a fix here would only treat a symptom of that practice, leaving other parts like motion statistics broken. I think a more robust idea would be that when creating a (single) motion, the motion gets applied to the existing results.

Thoughts?

PR #2231, opened 2023-01-28
Regroup motion release messages to use plural 

tienne-B:
See commit messages.

Continues on @daankoning's work in #2221.


daankoning:
Looks like good work!

That said I noticed there's a lot of other cases where a similar issue pops up in a lot of other places. Using `grep -rn --include "*.po" "(s)" tabbycat` will list them (albeit with a lot of duplication because it lists them in every language).  A lot of these also aren't problems as they are either part of the documentation or reference quantities that Tabbycat doesn't have (direct) access to (e.g. the number of tab directors).  However some of these, like the "Submit ballot(s)" buttons are resolveable.

Would you want to contain that within the scope of this PR or should I open a new issue?

tienne-B:
> Would you want to contain that within the scope of this PR or should I open a new issue?

You may create a new ticket.

PR #2232, opened 2023-01-29
Add extra validations for round break categories

tienne-B:
This commit adds new validations when creating/modifying rounds through the DB to avoid preliminary rounds having an associated break category, and so that the break category has the same tournament as the round.

It also consolidates a couple conditions to use XNOR and XOR.

### DIFF
--- a/tabbycat/tournaments/models.py
+++ b/tabbycat/tournaments/models.py
@@ -362,22 +362,23 @@ class Round(models.Model):
         return "[%s] %s" % (self.tournament, self.name)
 
     def clean(self):
-        errors = {}
+        errors = {
+            'break_category': [],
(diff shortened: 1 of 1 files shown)


philipbelesky:
Mild preferences, feel free to disregard and merge as-is:

- I do prefer conditionals using a more python-ey `and` / `is` construct. But it's not like they are a particularly complicated set of conditions.
- I have a slight preference for the more individualised messages so they can be more tailored and explicit.
- For the break category message, perhaps just "An elimination round must have a break category and preliminary round must not"?

tienne-B:
Merged as d44cded56bc03c0292d3aef52ffcda0feb192dec

PR #2235, opened 2023-01-31
Allow moving whole panel to unallocated

### LINKED ISSUE #2225: Allow dragging panels to unallocated
Adjudicators can be dragged back from a panel to the "unallocated" section. It should be possible to drag entire panels back as well.

See also #1457.

tienne-B:
This commit makes it possible to drag panels from the adjudicator allocation interface to the "unallocated" panel without errors.

Fixes #2225


philipbelesky:
LGTM, thanks for the addition. I should have thought of this in the original implementation!

PR #2248, opened 2023-02-01
fix: modified get_public_name to show redactions on participant team views

### LINKED ISSUE #2249: Redactions not reflected in team views
Whenever a person tries to view an individual team's record, the full names of all speakers are shown, ignoring redactions.

Veyron2121:
@tienne-B anything I should specifically include in this description?

Fixes #2249

### DIFF
--- a/tabbycat/participants/models.py
+++ b/tabbycat/participants/models.py
@@ -7,6 +7,7 @@ from django.db import models
 from django.db.models import Value
 from django.db.models.functions import Coalesce
 from django.utils.functional import cached_property
+from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 
 from utils.managers import LookupByNameFieldsMixin
@@ -129,6 +130,8 @@ class Person(models.Model):
         return str(self.name)
 
     def get_public_name(self, tournament):
+        if self.anonymous:
+            return mark_safe("<em>" + _("Redacted") + "</em>")
         if tournament.pref('participant_code_names') == 'off':
             return self.name
         return self.code_name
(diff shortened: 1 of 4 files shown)


tabbycat/participants/models.py
```diff
@@ -129,6 +130,9 @@ def __str__(self):
         return str(self.name)
 
     def get_public_name(self, tournament):
+        # This allows for jinja html templates to still access redactions from the tab
+        if self.anonymous:
+            return mark_safe("<em>Redacted</em>")
```
tienne-B, line 135: "Redacted" should be marked for translation, like
```suggestion
            return mark_safe("<em>" + _("Redacted") + "</em>")
```
No need for that comment (it's also wrong as we don't use Jinja) :P

tienne-B (commented):
Your PR description is fine; I often just say "look at the commit messages".

philipbelesky:
Quick question, for my own curiosity, is the expectation now that redactions also apply to non-public pages? Originally it was a concession to the public / search-indexed nature of tab releases, but I guess it is now intended to cover all places where names are shown to participants?

tienne-B:
Non-public like private URL-accessible pages? I'd think that keeping a simpler approach of admin vs non-admin would be better than also adding "restricted." We should check for the places that use `get_public_name()` as to not code/anonymize on admin pages.

philipbelesky:
> Non-public like private URL-accessible pages?

Yea, that was what I was thinking. But to be clear, I don't have any preference about what should happen, I was just curious about what user expectations look like these days. The simpler implementation is definitely preferred!

tienne-B:
Yeah, think the expectation is basically "remove all traces I was here"

Issue #2234, opened 2023-01-30
Many instances of (ugly) ambiguous plural "(s)"


daankoning:
For more discussion see #2231 

In short: running `grep -rn --include "*.po" "(s)" tabbycat` will list all the cases where this problem pops up. The cases where this is a problem are listed below (notwithstanding those already adressed in #2231 ).

- [ ] The "Send Message(s)" button on the email page.
- [ ] The "Submit Ballot(s)" button on adjudicator's private URLs as well as on the admin interface.
- [ ] The "Team(s) on %(side)s" text on the page for a debate team error.
- [ ] The text phrase "breaks(s)" in the warning that the breaks have already been published while editing them.

There were many other instances of "(s)" but according to my better judgement it appears that these weren't solvable (for reasons similar to those outlined in the linked PR).

I'd be happy to submit a PR for this issue, but it'd take until the end of the week for me to get time to get around to it so if anyone else wants to address this, be my guest.

tienne-B:
* "Send Message(s)" will be hard to implement as it'll require some JS trickery to count the number of marked checkboxes.
* "Team(s) on %(side)s" could just be "Teams on %(side)s" without worrying for singular as it's a table heading so there wouldn't be a way to count (it's counted per debate in the body)

PR #2250, opened 2023-02-03
Disambiguate plurals

daankoning:
Fixes (part of ) #2234

### DIFF
--- a/tabbycat/breakqual/base.py
+++ b/tabbycat/breakqual/base.py
@@ -5,7 +5,7 @@ import logging
 from itertools import groupby
 
 from django.utils.encoding import force_str
-from django.utils.translation import gettext as _
+from django.utils.translation import ngettext
 
 from breakqual.models import BreakingTeam
 from standings.teams import TeamStandingsGenerator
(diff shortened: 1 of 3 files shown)


tabbycat/results/templates/public_enter_results.html
```diff
@@ -61,7 +61,8 @@
         <input id="id_debate_result_status" type="hidden" name="debate_result_status" value="{{ debate.STATUS_DRAFT }}" />
         <input id="id_discarded" type="hidden" name="discarded" />
         <input id="id_confirmed" type="hidden" name="confirmed" />
-        <input class="save btn btn-success btn-block " type="submit" value="{% trans 'Submit Ballot(s)' %}" tabindex="{{ form.nexttabindex }}"/>
+        {% blocktranslate count counter=form.scoresheets|length asvar button_text %} Submit Ballot {% plural %} Submit Ballots {% endblocktranslate %}
```
tienne-B, line 64:
```suggestion
        {% blocktranslate trimmed count counter=form.scoresheets|length asvar button_text %}
          Submit Ballot
        {% plural %}
          Submit Ballots
        {% endblocktranslate %}
```

tabbycat/breakqual/base.py
```diff
@@ -90,11 +90,23 @@ def _metric_name(metric):
...
+                "The break qualification rule %(rule)s requires the following "
+                "metrics to be in the team standings precedence in order to "
+                "work: %(required)s;",
+                len(self.required_metrics),
+            )
+            missing_metrics_message = ngettext(
+                " and the following is missing: %(missing)s.",
+                " and the following are missing: %(missing)s.",
+                len(missing_metrics),
+            )
```
tienne-B, line 106: I'd rather rephrase the message to only require one list than split it in two, as the ordering of clauses (or even agglutination) could determine how they're combined differently than in English. Maybe something like
> The %(rule)s break qualification rule is missing the following required metric in the team standings precedence: %(missing)s

@czlee ?

czlee: Yeah, this kind of thing was a struggle for me in several places with internationalization. I agree, you can't assume that you can just concatenate sentences like that in other languages (though I can't actually name you a counterexample language off the top of my head for this case).

This error message evidently tried to be detailed enough to be as helpful as possible, but yeah, I agree that dropping the `required` list would be a minimal sacrifice in this context.

While I'm here, thanks so much for doing this @daankoning!!

daankoning:
I thought I'd be able to fix it before the weekend (and went away on a tournament), evidently I was not 😅. 

As @tienne-B points out in #2234  the "Send Message(s)" button requires some JS which I do not have enough experience to figure out. Someone else would have to do that if they deem it important enough to fix.

tabbycat/breakqual/base.py
daankoning: Implemented this!

PR #2267, opened 2023-02-26
Fix crash when running out of emoji

franga2000:
This fixes a critical issue we ran into yesterday, where running out of available emoji makes it impossible to add teams from either the API or the admin site. When all emoji are already used, `pick_unused_emoji` returns `None`, but none of the calling code handles that option. 

And no, we didn't have 488 teams at a tournament, it's just that `pick_unused_emoji` considers all teams on the instance and not just the current tournament (and we have a big archive).

This PR:

- makes `pick_unused_emoji` look only inside one tournament if provided
- provides the tournament when creating teams from the API
- in the unlikely case that there are still no emoji left, simply leaves the field blank

### DIFF
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -668,14 +668,13 @@ class TeamSerializer(serializers.ModelSerializer):
         break_categories = validated_data.pop('break_categories', [])
         venue_constraints = validated_data.pop('venue_constraints', [])
 
-        emoji = pick_unused_emoji(validated_data['tournament'].id)
-        if emoji:
-            if 'emoji' not in validated_data or validated_data.get('emoji') is None:
-                validated_data['emoji'] = emoji[0]
-            if 'code_name' not in validated_data or validated_data.get('code_name') is None:
-                validated_data['code_name'] = emoji[1]
+        emoji, code_name = pick_unused_emoji(validated_data['tournament'].id)
+        if 'emoji' not in validated_data or validated_data.get('emoji') is None:
+            validated_data['emoji'] = emoji
+        if 'code_name' not in validated_data or validated_data.get('code_name') is None:
+            validated_data['code_name'] = code_name
 
-        if validated_data.get('emoji') == '':
+        if validated_data['emoji'] == '':
             validated_data['emoji'] = None  # Must convert to null to avoid uniqueness errors
 
         team = super().create(validated_data)
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -668,14 +668,14 @@ class TeamSerializer(serializers.ModelSerializer):
         break_categories = validated_data.pop('break_categories', [])
         venue_constraints = validated_data.pop('venue_constraints', [])
 
-        emoji = pick_unused_emoji()
+        emoji = pick_unused_emoji(validated_data['tournament'].id)
         if emoji:
             if 'emoji' not in validated_data or validated_data.get('emoji') is None:
                 validated_data['emoji'] = emoji[0]
             if 'code_name' not in validated_data or validated_data.get('code_name') is None:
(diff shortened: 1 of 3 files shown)


tabbycat/api/serializers.py
```diff
@@ -466,13 +466,14 @@ def create(self, validated_data):
...
-
-        if validated_data['emoji'] == '':
+        emoji = pick_unused_emoji(validated_data['tournament'].id)
+        if emoji:
+            if 'emoji' not in validated_data or validated_data.get('emoji') is None:
+                validated_data['emoji'] = emoji[0]
+            if 'code_name' not in validated_data or validated_data.get('code_name') is None:
+                validated_data['code_name'] = emoji[1]
+
+        if validated_data.get('emoji') == '':
```
tienne-B, line 476: I don't see why we have to make all these changes; shouldn't just adding the `tournament` parameter be sufficient?
```suggestion
        emoji, code_name = pick_unused_emoji(self.context.get('tournament'))
        if 'emoji' not in validated_data or validated_data.get('emoji') is None:
            validated_data['emoji'] = emoji
        if 'code_name' not in validated_data or validated_data.get('code_name') is None:
            validated_data['code_name'] = code_name

        if validated_data['emoji'] == '':
```

tabbycat/participants/emoji.py
```diff
@@ -24,12 +24,14 @@ def set_emoji(teams, tournament):
         team.save()
 
 
-def pick_unused_emoji():
+def pick_unused_emoji(tournament_id=None):
```
tienne-B, line 27: Could use the entire `tournament` object as a parameter rather than just `id`.

tienne-B (commented):
Looks good!

tabbycat/api/serializers.py
franga2000: Without the extra check, this code could still crash if `pick_unused_emoji` returned `None`. My logic was that something as "unimportant" as picking an emoji shouldn't even theoretically be able to crash something as critical as adding a team. And if the emoji allocation code changes in the future (like to allow users to edit the list or something), this issue could silently come back.

Do I revert it anyways?

tienne-B: The code here should not crash if that's the case, as both
```py
validated_data['emoji'] = emoji
validated_data['code_name'] = code_name
```
are allowable when `emoji` or `code_name` are `None`.

The proper way to fix it would be for `pick_unused_emoji()` to return `None, None` on an `IndexError`, making the return type `Tuple[Optional[str], Optional[str]]` (rather than `Optional[Tuple[str, str]]`) as what was expected both here and in `participants/admin.py`.

That way, we wouldn't need this extra check for the emoji, and you wouldn't need any changes in `participants/admin.py`.

franga2000: Making it return a tuple was my first thought too, but I didn't want to change the function signature - I only later realized the function is only used twice, so it's really easy. I'll do that.

Re. type hinting, I quite like the policy of adding type hints any time a function gets a new caller - if it's only used once, don't bother, but the first time someone adds an additional call site, type hints should be added in the same PR.    
*Of course I usually forget to enforce this in the projects that I manage, but that's easily solvable with an approval checklist...*

tabbycat/participants/emoji.py
franga2000: The idea was that you don't need an additional query if all you have is the ID (if I remember my Django 101 right, if you have a `Team` object, `team.tournament` would run a query, but `team.tournament_id` would not).   
This, however, doesn't currently happen anywhere in the code and is rather unlikely to in the future, so I can of course change it

tienne-B: This was a minor point, so I merged regardless. Using model objects in queries is done smarter; it realizes it doesn't need to join the `tournaments_tournament` table for it's primary keys.

PR #2269, opened 2023-02-26
Add team seeding for first round

### LINKED ISSUE #2165: Round 1 seeding
In APDA, round 1 pairings/judge allocations are based on seeding. Individual debaters are considered seeded if they've qualified to the national tournament, and teams are either fully-, half-, or un-seeded. Pairings are high-low based on the level of a team's seeding (e.g. fully-seeded teams will tend to face unseeded teams), and judges are similarly allocated (with better judges going to rounds with higher seeds). Seeding does not affect pairings beyond round 1, which are power-paired based on previous rounds at that tournament.

This could be put in manually when pairing/allocating round 1, but would go much faster if integrated as a metric like priority into the auto-allocation feature.

tienne-B:
This commit adds a "seed" attribute to Teams which are used for rankings in the first round if power-pairing is used. In that situation, all teams are placed in the same bracket with the Standings being manually created. It can then use the normal power-pairing draw generator.

To continue this feature, a new conflict avoidance should be made to minimize conflicts between pairings where teams have the same seed, i.e.

A 1 (seed 1) vs A 2 (seed 2)
B 1 (seed 1) vs B 2 (seed 2)

would become:
A 1 vs B 2
B 1 vs A 2

Fixes #2165


tabbycat/draw/manager.py
```diff
@@ -197,9 +198,18 @@ def get_teams(self):
...
+            standings.record_added_metric('points', _("points"), _("Pts"), None, False)
+            standings.record_added_metric('seed', _("seed"), _("Seed"), None, False)
+            for team in teams:
+                standings.add_metric(team, 'points', 0)
+                standings.add_metric(team, 'seed', team.seed)
+            standings.sort(('points', 'seed'), random.shuffle)
+        else:
+            generator = TeamStandingsGenerator(metrics, ('rank', 'subrank'), tiebreak="random",
+                extra_metrics=(pullup_metric,) if pullup_metric and pullup_metric not in metrics else ())
+            standings = generator.generate(teams, round=self.round.prev)
```
czlee, line 212: _[Scratch this comment and replace with the next one below, but I'm leaving this here so that my progression of thinking / brainstorm is open.]_

Would there be less special-case logic if we just appended a `'seed'` team metric to the metrics lists passed into `TeamStandingsGenerator`? So that it kind of becomes another tiebreak? (Though not in the same way—I realise we'd still have to fall back to `random.shuffle` because seeds aren't unique.) This implementation of diving back into low-level functions to generate a "special standings" seems kinda clunky and non-extensible.

I guess in this case we'd need to configure it using a new `Round` property that controls whether to append `'seed'`, since we don't want the seed to be in the actual rankings. But I was thinking we don't want to be tied specifically to the first round, if we can use a structure that is both simpler in implementation and more flexible to use.

(Admittedly I sometimes wondered whether the "standings precedence for the draw" and "standings precedence for the published tab" should be separated... but that's another topic.)

_[Again, scratch this comment and replace with the one below.]_

czlee: Actually, a possibly better idea: Rather than implementing logic that treats `self.round_prev is None` as a special case of power-paired draws (since I'd argue it's not really, and implementing it as one is a bit of a kludge), write a new `DrawGenerator` subclass that just does what we want directly.

This new `SeededDrawGenerator` should just strip out everything that's not needed to keep the logic clear—skip all of the complicated odd bracket resolution logic, skip the `points` field, maybe even skip the pairing function options. It should just implement the seed pairing (as desired by APDA) directly. (Or possibly extract common logic to a common parent class with `PowerPairedDrawGenerator`.)

To select it, add a new `Round.draw_type` option, so the first round would have draw type "seeded" rather than "random".

Usability downside: APDA tournaments would need to edit the first round to choose this option explicitly (or we'd need to make some UI to make it easy, _e.g._ by allowing the user to select it on the availability page.

tabbycat/availability/templates/availability_index.html
```diff
@@ -132,15 +132,20 @@
...
+      <div class="row align-items-center">
+        <div class="col-auto pr-1">
+          <i data-feather="alert-circle"></i>
+        </div>
+        <div class="col pl-0 pr-0">
+          {% blocktrans trimmed %}
+            A power-paired first round will use the teams' seeds for ranking.
+            Did you intend for it to be <strong>Random</strong> instead? You can
+            <a href="{{ change_round_url }}" class="alert-link">edit this
+            round's draw type in the Edit Database area.</a>
```
czlee, line 145: Reading this new message was in part what got me thinking about implementing this as a new draw type / draw generator altogether. I think it's confusing for "power-paired" to mean different things in different rounds, and suggest that it'd be easier conceptually if "power-paired first round" was just always (probably) a mistake.

czlee (commented):
Models and import forms look good, but I thought the draw logic would be better spun out to a new generator altogether that only deals with logic for this (new) draw type, rather than seeing it as a special case of the much more complicated power-pairing generator.

tabbycat/draw/manager.py
tienne-B: I would make a new draw manager for seeds, but I wouldn't want to make a new draw generator, as the complexity it would add between the new mixins and inherited classes would overweigh the convenience of the generator having implemented extra steps (that would be skipped anyhow) for a more general case. It's already somewhat hard to follow between generators and managers, so subdividing even more wouldn't help.

czlee: _[comment written before reviewing new code]_

Hmm, yeah, when I decided to rely on mixins and subclasses to mix and match functionality, I was mostly inspired by the patterns they use for class-based views in Django. I'd be open to the idea that this architecture is poorly suited for this application.

Brainstorm idea: Would it make sense to conceptualise draw generation as a series of steps (like, transforms), that are built into a sequence? So then each step would be in its own class, and some sort of composite class would know how to take in a list of steps and apply each one in turn. Examples of steps would be each existing option for each existing step:

- pull up teams to resolve odd brackets
- create intermediate brackets to resolve odd brackets
- rank by seed within brackets
- rank by points within brackets
- pair teams by folding within brackets

Major challenge would be defining a (common) interface for steps, because the input and output of each step don't "naturally" look the same. Some will output lists of teams in brackets, some will guarantee that brackets are aff/neg balanced, others will output an actual draw.

This might be too major a refactor to be worth it in this PR... but I wanted to put the thought up for discussion somewhere. Or another architecture, open to suggestions.

tabbycat/tournaments/models.py
```diff
@@ -275,38 +275,24 @@ def get_queryset(self):
...
-        (STATUS_RELEASED, _('Released')),
-    )
+
+    class DrawType(models.TextChoices):
+        RANDOM = 'R', _('Random')
+        MANUAL = 'M', _('Manual')
+        ROUNDROBIN = 'D', _('Round-robin')
+        POWERPAIRED = 'P', _('Power-paired')
+        ELIMINATION = 'E', _('Elimination')
+        SEEDED = 'S', _('Seeded')
```
czlee, line 285: Does adding this value to the enum require a migration?

czlee (approved):

PR #2271, opened 2023-02-26
Implement per-debate speaker ranks

### LINKED ISSUE #1033: Add fields for speaker rankings
In the APDA format, all speakers are given both a speaking score _and_ a ranking to compare them from the others in the debate. Most of these could be determined by their score, but would be needed for the case of ties (see #643 for draws in team scores).

The ranking could be in a new `SpeakerScore` field (Arrow prevents rankings by-adj), and these rankings could be used in standings similarly to BP points or rankings count (1sts, etc.).

tienne-B:
Fixes #1033


tabbycat/results/forms.py
```diff
@@ -488,6 +494,9 @@ def create_participant_fields(self):
             self.fields[self._fieldname_speaker(side, pos)] = forms.ModelChoiceField(
                 queryset=queryset, required=True)
 
+            if self.tournament.pref('speaker_ranks') != 'none':
+                self.fields[self._fieldname_srank(side, pos)] = forms.IntegerField(required=True, min_value=1, max_value=nspeeches, step_size=1)
+
```
czlee, line 499: I think these would be better thought of as score fields, and the logic moved to `SingleBallotSetForm.create_score_fields` and (possibly) `PerAdjudicatorBallotSetForm.create_score_fields`.

I guess the reason `self.create_score_fields()` is extract to its own method is because the score fields differ depending on whether it's voting or consensus ballots. I think in principle this is also true for speaker ranks—if adjudicators are scoring speakers independently, they should also be ranking speakers independently(?)

If APDA never does voting ballots, I'm fine with only implementing the logic for consensus ballots, and ignoring the speaker ranks setting in voting ballots (_i.e._ only putting the logic in `SingleBallotSetForm`). As things currently stand, when it uses voting ballots, our form logic creates only one set of speaker rank fields, but renders those same fields once per adjudicator, which leads to weird behaviours caused by duplicate fields.

image

czlee (commented):
This looks great! Very exciting to see something that's been on the backlog for so long coming through. But I think it should be thought of as structurally analogous to speaker score fields, rather than structually analogous to the speaker identity fields. It's fine in my opinion not to implement for voting ballots (I guess it might be incoherent?). But in that case, the logic should be in `SingleBallotSetForm` so that it shows up for consensus ballots, but voting ballots don't even know that the concept exists. See comment at results/forms.py:499.

czlee:
I realised we'd also need to add provision for speaker ranks here:
https://github.com/TabbycatDebate/tabbycat/blob/9191bbd23da9ee20557e58a9fe76c9c413143d95/tabbycat/results/prefetch.py#L124-L131

Something like
``` python
            result.speaker_ranks[ss.debate_team.side][ss.position] = ss.rank
```
after line 128.

(Figuring out how to implement efficient prefetches for results without either duplicated or overly convoluted logic was a problem I never solved.)

tabbycat/results/result.py
czlee, line 1: This took me a while to figure out, but other than `speakerscore_fields` on line 627, I think all of the logic relating to speaker ranks should be moved to results/scoresheet.py, for the same reason that we made it analogous to speaker scores not speaker identities in results/forms.py.

So then `DebateResultWithScoresMixin.speakers_as_dicts` would delegate to `sheet.get_speaker_rank` rather than `self.get_speaker_rank`.

Also, `DebateResultWithScoresMixin.identical` wouldn't need to worry about speaker ranks (I think this is currently a missed spot), because that would be the responsibility of the scoresheet classes.

Sadly, this would involve another bunch of mixins and classes in scoresheet.py, since the treatment would be analogous to declared winners, which are encapsulated in `DeclaredWinnersMixin`. But if that's too unwieldy I think I'd be okay with using control flow (if statements) to switch between including and excluding ranks. It might just take some finesse—ideally to avoid confusion, speaker ranks are never presented to other components (_e.g._ passed to the template via the `as_dicts()` ) if they're not relevant.

Hope that made sense, lmk if not?

czlee (commented):
Mostly the restructure suggested in results/result.py, involving moving most logic to results/scoresheet.py 🙊

czlee:
Also, I feel like my unit tests probably aren't doing what they're meant to do… a lot of the minor errors I picked up are consequences of crazy configuration combinations not being hit. Though I guess most of them are in results/forms.py, which doesn't have any unit tests on it. It's not really practical to manually test every case anymore, with how many combinations we have, so it's not too surprising some are getting through.

On this note, I guess it'd be nice to add unit tests for these… but I suspect these unit tests need a deeper look more generally, so I don't think unit tests should block this PR.

tabbycat/results/result.py
tienne-B: I would just put it in `ScoresMixin`. I'm not concerned with the values of the ranks if disabled.

PR #2289, opened 2023-03-15
Create list of actionlog entries in API

### LINKED ISSUE #1993: Create API endpoint for actionlog
Would be simple GET list/detail endpoints for action logs.

daankoning:
Closes #1993 

Right now this leans quite heavily on generic fields. Would it be better to add a serializer for, for example, the `user` field?

I'm also not too sure on the current information given, `"type"` and `"object_id"` might be quite unclear?

### DIFF
--- a/tabbycat/api/serializers.py
+++ b/tabbycat/api/serializers.py
@@ -1351,7 +1351,43 @@ class TeamRoundScoresSerializer(serializers.ModelSerializer):
 
 
 class ActionLogSerializer(serializers.ModelSerializer):
+    object = serializers.SerializerMethodField()
 
     class Meta:
         model = ActionLogEntry
-        fields = ('id', 'user', 'ip_address', 'timestamp', 'type', 'object_id')
+        fields = ('id', 'user', 'ip_address', 'timestamp', 'object')
+
+    def get_object(self, obj: ActionLogEntry):
+        url_format = {
+            'ba': '/rounds',  # TODO
+            'fb': '/feedback/%(id)s',
+            'ts': '/feedback/%(id)s',
+            'dr': '/rounds/%(round_number)s/pairings',
+            'aa': '/rounds/%(round_number)s/pairings',
+            'mu': '/rounds/%(round_number)s/pairings',
+            'ms': '/rounds/%(round_number)s/pairings',
+            've': '/venues/%(id)s',
+            'mo': '/motions/%(id)s',
+            'db': '',  # TODO
+            'rd': '/rounds/%(round_number)s',
+            'br': '/break_categories/%(id)s',
+            'av': '',  # TODO
+            'op': '/preferences',
+            'se': '/speaker-categories/%(id)s',
+            'si': '',  # TODO
+            'aj': '/adjudicators/%(id)s',
+            'ac': '',  # TODO
+            'ch': '',  # TODO
+            'pp': '/rounds/%(round_number)s/preformed-panels/%(id)s',
+        }[obj.type[:2]]
+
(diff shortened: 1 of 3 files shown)


tabbycat/api/serializers.py
```diff
@@ -1344,3 +1348,17 @@ class Meta:
...
+
+
+class ActionLogSerializer(serializers.ModelSerializer):
+
+    id = serializers.CharField(read_only=True)
+    user = serializers.CharField(read_only=True)
+    ip_address = serializers.IPAddressField(read_only=True)
+    timestamp = serializers.DateTimeField(read_only=True)
+    type = serializers.ChoiceField(choices=ActionLogEntry.ACTION_TYPE_CHOICES, read_only=True)
+    object_id = serializers.CharField(read_only=True)
```
tienne-B, line 1360: Would these be necessary? They are model fields included by being on line 1364. They shouldn't be considered read-only in the serializer itself.

There should be a (this time read-only) URL field for the action-log object.

The real problem to solve for #1993 is to create a polymorphic URL field (called `object` or `target`) that uses `type` and `object_id` to create an API URL to the affected object.

tabbycat/api/urls.py
```diff
@@ -224,6 +224,14 @@
                         views.VenueCategoryViewSet.as_view(detail_methods),
                         name='api-venuecategory-detail'),
                 ])),
+                path('/actionlog', include([
+                    path('',
+                        views.ActionLogViewSet.as_view(list_methods),
```
tienne-B, line 229:
```suggestion
                        views.ActionLogViewSet.as_view({'get': 'list'}),
```
By only keeping the `get` verb, the API views become read-only.

tabbycat/api/urls.py
```diff
@@ -224,6 +224,14 @@
...
+                        name='api-actionlog-list'),
+                    path('/<int:pk>',
+                         views.ActionLogViewSet.as_view(detail_methods),
```
tienne-B, line 232:
```suggestion
                         views.ActionLogViewSet.as_view({'get': 'retrieve'}),
```
Same here. This also prevents deletion.

tienne-B (commented):
I hope I've answered the question in your PR within the comments.

tabbycat/api/serializers.py
daankoning: Sure enough, I have a solution to this figured out. My only problem is what do you think is the best way to handle actionlog entries on objects that don't have corresponding API endpoints (like checkins or preformed panels?

tienne-B: Fair point, but I don't think that's a problem; the API is quite comprehensive (including with how to checkin an object or preformed panels). For checkins, I think we use the checkin status as a state on the corresponding object, so you'd link to the ballot rather than ballot/checkin.

daankoning:
So I have an Idea on how to do the linking to other parts but I am unsure this is a particularly good solution. Additionally I left a couple comments where I have no idea where certain action log events should point to.

 I may have bitten off a bit more than I can chew here, I'm happy to continue to try to figure it out but I recognize that that might also be quite a time investment for you that you may or may not be able/willing to make. 😅

tienne-B:
I'd agree that it's an unsatisfying solution to a hard problem. This is how I made a polymorphic URL field (look specifically at `ParticipantSourceField`):
https://github.com/TabbycatDebate/tabbycat/blob/6e1de4bbae941476dbafc96c7f2416a882af31ee/tabbycat/api/fields.py#L209-L282

It may assist in creating a similar field for this.

PR #2298, opened 2023-04-07
Implement new draw algorithm for 2-team formats

tienne-B:
See commit messages

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -39,6 +39,7 @@ whitenoise = "*"
 drf-spectacular = "*"
 daphne = "*"
 drf-link-header-pagination = "*"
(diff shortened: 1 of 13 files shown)


czlee (approved):
This looks fantastic! Very exciting and very nicely implemented. I didn't know about the blossom algorithm before today.

It crashed on me the first time I tried it, though, see the TypeError below.

High-level things to consider:
- I wonder if using subclasses rather than control flow (`if` statements) to distinguish between Australs-style and APDA-style conflict avoidance would be cleaner, at least in terms of separation of concerns.
- Maybe nice in the draw details screen to see actual costs to help users gauge whether they want to adjust penalties and rerun. But definitely not essential and possibly more confusing than enlightening.

PR #2302, opened 2023-04-26
Automatically generate private URLS and identifiers on participant and venue creation

### LINKED ISSUE #2179: Automatically generate private URL key and checkin identifier on participant import
It would reduce the number of steps required to use the checkins and private URLs features if people's/rooms keys could be automatically generated when importing them.

daankoning:
Closes #2179

### DIFF
--- a/tabbycat/importer/forms.py
+++ b/tabbycat/importer/forms.py
@@ -154,8 +154,8 @@ class VenueDetailsForm(BaseTournamentObjectDetailsForm):
 
     def save(self, *args, **kwargs):
         venue = super().save(*args, **kwargs)
-
         VenueIdentifier.objects.create(venue=venue)
+        return venue
 
 
 class BaseInstitutionObjectDetailsForm(BaseTournamentObjectDetailsForm):
--- a/tabbycat/importer/forms.py
+++ b/tabbycat/importer/forms.py
@@ -152,9 +152,10 @@ class VenueDetailsForm(BaseTournamentObjectDetailsForm):
         model = Venue
         fields = ('name', 'priority')
 
-    def save(self, *args, **kwargs):
-        venue = super().save(*args, **kwargs)
-        VenueIdentifier.objects.create(venue=venue)
+    def save(self, commit=True):
+        venue = super().save(commit=False)
+        if commit:
+            VenueIdentifier.objects.create(venue=venue)
         return venue
 
 
@@ -332,10 +333,11 @@ class AdjudicatorDetailsForm(BaseInstitutionObjectDetailsForm):
     def save(self, commit=True):
         adj = super().save(commit=commit)
 
-        if commit and adj.institution:
-            adj.adjudicatorinstitutionconflict_set.create(institution=adj.institution)
-        PersonIdentifier.objects.create(person=adj)
-        populate_url_keys([adj])
+        if commit:
+            if adj.institution:
+                adj.adjudicatorinstitutionconflict_set.create(institution=adj.institution)
+            PersonIdentifier.objects.create(person=adj)
+            populate_url_keys([adj])
 
         return adj
 
--- a/tabbycat/importer/forms.py
+++ b/tabbycat/importer/forms.py
@@ -8,7 +8,9 @@ from django.core.exceptions import ValidationError
 from django.core.validators import validate_email
 from django.utils.translation import gettext as _, ngettext, ngettext_lazy
 
+from checkins.models import PersonIdentifier, VenueIdentifier
 from participants.models import Adjudicator, Institution, Speaker, Team
+from privateurls.utils import populate_url_keys
 from venues.models import Venue
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


tabbycat/participants/models.py
```diff
@@ -9,6 +9,8 @@
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 
+import privateurls
```
tienne-B, line 12: Rather import just `populate_url_keys` rather than the whole module.

tienne-B (commented):
Sorry for the delay in reviewing this!

I am not a huge proponent of extending models' `save()` method, as it makes the model lose flexibility, like no bulk creation/update. It also requires the creation of these objects which may be often wanted but isn't necessarily...

I would suggest putting these in the saving procedure of the simple importer forms directly rather than in the model.

tabbycat/privateurls/utils.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger(__name__)
 
 
-def populate_url_keys(people: 'QuerySet[Person]', length: int = 8, num_attempts: int = 10) -> None:
+def populate_url_keys(people: 'Iterable[Person]', length: int = 8, num_attempts: int = 10) -> None:
```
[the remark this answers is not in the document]
daankoning: You do in order to make type checkers and intellisense work :)

daankoning:
I had that thought as well but then you run into the issue that it won't work when creating objects through other means (such as the API, database or custom importers). Or do you think it is okay to still require manual identifier creation in those cases?

tabbycat/participants/models.py
daankoning: Then you run into circular imports, for now I've tried this as possibly a better alternative...

tienne-B:
> I had that thought as well but then you run into the issue that it won't work when creating objects through other means (such as the API, database or custom importers). Or do you think it is okay to still require manual identifier creation in those cases?

Yes, especially that being able to create people without URL keys in the API is a feature which required extra code :P
https://github.com/TabbycatDebate/tabbycat/blob/6c93deee3982cc39982b8bbcdffc87f96f3b77c2/tabbycat/api/serializers.py#L474-L476

I'd tend towards assuming that if a user is going through a more complicated way to give them more customizability. Maybe they want to assign keys/IDs manually.

tabbycat/importer/forms.py
tienne-B, line 1: I wonder for all of these if it'd be possible to create all identifiers (URL keys and checkin IDs) at the same time. It should be possible as `checkins.utils.create_identifiers()` and `privateurls.utils.populate_url_keys()` both take an `Iterable` of objects. Thoughts? Might it be more efficient?

daankoning: That seems like it might be less readable than status quo. Primarily because if there were one `generate_identifiers` function it would have to check whether or not is being run on a venue (as then there'd be no need for a private URL) leading to some ugly branching with little shared logic. Secondly, the current implementation here does not use `checkins.utils.create_identifiers` at all for readability's sake as `VenueIdentifier.objects.create(venue=venue)` quite succinctly explains what is happening whereas `generate_identifiers` would not be immediately intuitive and would require opening another file and checking.

tabbycat/importer/forms.py
```diff
@@ -320,8 +331,12 @@ def clean_base_score(self):
 
     def save(self, commit=True):
         adj = super().save(commit=commit)
+
         if commit and adj.institution:
             adj.adjudicatorinstitutionconflict_set.create(institution=adj.institution)
+        PersonIdentifier.objects.create(person=adj)
+        populate_url_keys([adj])
```
tienne-B, line 338:
```suggestion
        if commit:
            if adj.institution:
                adj.adjudicatorinstitutionconflict_set.create(institution=adj.institution)
            PersonIdentifier.objects.create(person=adj)
            populate_url_keys([adj])
```

tabbycat/importer/forms.py
```diff
@@ -150,6 +152,11 @@ class Meta:
         model = Venue
         fields = ('name', 'priority')
 
+    def save(self, *args, **kwargs):
+        venue = super().save(*args, **kwargs)
+        VenueIdentifier.objects.create(venue=venue)
+        return venue
```
tienne-B, line 158: Should use the same signature as the other `save()` methods:
```suggestion
    def save(self, commit=True):
        venue = super().save(commit=False)
        if commit:
            VenueIdentifier.objects.create(venue=venue)
        return venue
```

tabbycat/importer/forms.py
tienne-B: I'm not sure I understand this logic... the `generate_identifiers()` function already exists and is used to generate many checkin IDs at once.

I was not proposing to merge the two methods. What I was suggesting was to use those methods to bulk-create/update the IDs and keys rather than generate them one at a time. This would mean extending the `save()` method for the formsets themselves.

tienne-B (commented):
Really close! Sorry for the delays

tabbycat/importer/forms.py
daankoning: Whoops, you're right my understanding was to merge them :)

Then my question is just how you'd overwrite the `save()` method given that these formsets are generated by `modelformset_factory()`.

tienne-B: Fair; I'll merge when it passes the checks. Thank you!

PR #2305, opened 2023-05-27
Make last third of rounds silent by default

CreatedBySeb:
Make the last third of rounds auto-generated by the setup wizard silent by default. This is more inline with common tournament configurations, such as:

* 5 rounds: Rounds 1 - 3 are open, Rounds 4 & 5 are silent
* 9 rounds: Rounds 1 - 6 are open, Rounds 7 - 9 are silent

While this may not fit all tournament configurations, especially ones with fewer/no silent rounds, forgetting to mark a round as not silent is less of an issue than forgetting to mark a rounds as silent, so it is probably a preferable behaviour. An ideal solution in the future may be to have a way to pick which rounds are silent in the setup wizard.

### DIFF
--- a/tabbycat/tournaments/utils.py
+++ b/tabbycat/tournaments/utils.py
@@ -68,9 +68,11 @@ BP_SIDE_NAMES = {  # stop-gap before this system gets refactored
 
 def auto_make_rounds(tournament, num_rounds):
     """Makes the number of rounds specified. The first one is random and the
-    rest are all power-paired. The last one is silent. This is intended as a
-    convenience function. For anything more complicated, a more advanced import
-    method should be used."""
+    rest are all power-paired. The last third of rounds (rounded down) are silent.
+    This is intended as a convenience function. For anything more complicated,
+    a more advanced import method should be used."""
+    silent_threshold = num_rounds * 2 / 3
+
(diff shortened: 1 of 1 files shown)


tienne-B:
I'm a little conflicted between having Round 4 out of 5 be silent by default; I've had it both ways, and it seems to be more with scheduling rather than a normal practice, like for the last round.

Agree about having a better way to set up rounds to be silent. Maybe the "Set Round Weights" page could be adapted.

czlee:
For historical context: I think we defaulted to last round silent mostly because it was the most common system on the New Zealand circuit at the time. I don't remember what was common in Australs at the time, but that would've been the other model we would've likely followed. (This was long before we added support for BP.)

I'm open to changes to whatever is easiest for the current circuits. If there's a lot of diversity with configurations it might be best to require users to specify explicitly every time, or maybe even ask only when it becomes relevant… it'd risk being annoying, but maybe that's less bad, idk.

philipbelesky:
Just a quick note that I agree with @CreatedBySeb that we should err on the side of being cautious, and default to the greatest plausible number of silent rounds. Better to potentially avoid a very negative experience for novice tab directors than add an extra step or two for everyone else.

PR #2308, opened 2023-05-28
User permissions

### LINKED ISSUE #131: Implement more graduated permissions
Currently there are pretty much three levels of access: no-login (public), login (monkey) and superuser. In general this works fine. But there are probably some superuser functions which should not actually grant access to all other functions too. For example:
- Ballot check-in should be its own type of thing.
- Displaying the draw (before public release) should be its own thing too, lest someone accidentally display confidential info to the projector screen in GA.
- Adjudication cores should probably have less permission than tab directors (other than staff access). For example, they probably shouldn't be able to edit ballots in the same way that tab directors can.

Groups are built into Django, but obviously we'd need to standardize them and have some sort of framework for them. This isn't really high priority, it's not as if it's created headaches in previous tournaments.

### LINKED ISSUE #1621: User that can only access feedback
This is more of a feature request than anything else, but would it be possible to create a user type that can only access the view feedback page? This would be especially useful for members of the Adj Core that are not experienced with Tabbycat to avoid any accidents.

### LINKED ISSUE #2173: Catalog distinct permissions for actions
This could use the list of actions from `actionlog` as a guide. We'd want to create permissions that grant permission to view or interact with specified pages. This task is to make a list of them and create a template for these permissions.

### LINKED ISSUE #2174: Create groups of permissions
Using the permissions created in #2173, for the ease of use for the permissions, a couple "groups" of user can be created, corresponding to roles within a tournament. These could be:
* Tab
  * All perms
* CA
  * Generate draw
  * Allocations
  * Motions
* Equity
  * View participants
  * View/modify conflicts/constraints

Once grouped, it could be necessary for the groups to be created for each new tournament, in the tournament creation process.

### LINKED ISSUE #2175: Restrict views and actions based on tournament permissions
This task is to apply the permissions created in #2173 to views and user actions. This is the widest-reaching task, but should be straightforward. The way to verify permissions is similar to Django itself, and is described [in the manual](https://django-guardian.readthedocs.io/en/stable/userguide/check.html).

tienne-B:
Starting work on #131. Trying to somewhat emulate the built-in Django[-guardian] framework, but as it doesn't work well with related objects, re-implementing it somewhat. More details in the commits.

Closes #131
Closes #1621
Closes #2173
Closes #2174
Closes #2175


tabbycat/users/models.py
```diff
@@ -0,0 +1,46 @@
+from django.conf import settings
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from utils.fields import ChoiceArrayField
+
+from .permissions import Permission
+
+
+class UserPermission(models.Model):
```
czlee, line 10: Should there be some hook here that expires the cached permissions upon save? I think the caching might be interfering with when a permission is revoked? Or more precisely it wouldn't take effect until the cache next expires, would that be okay with tab directors?

tabbycat/users/permissions.py
```diff
@@ -0,0 +1,206 @@
...
+            user._permissions[tournament.slug].add(permission)
+        return cached_perm
+
+    perm = (
+        user.userpermission_set.filter(permission=permission, tournament=tournament).exists() or
+        user.membership_set.filter(group__permissions__contains=[permission], group__tournament=tournament).exists()
+    )
+    if perm:
+        user._permissions[tournament.slug].add(permission)
+        cache.set("user_%d_%s_%s_permission" % (user.pk, tournament.slug, str(permission)), perm)
```
czlee, line 189: Just checking if this is supposed to fall back to the default timeout, which if I'm not mistaken is [5 minutes](https://docs.djangoproject.com/en/5.0/topics/cache/#cache-arguments)? This is fine imo, just wondering if you might've wanted to set the timeout here explicitly.

czlee (approved):
Looks fantastic. Mammoth effort, excellent work imho.

I wonder if we'd want to disable submit buttons where users don't have edit permissions, and disable menu items they're not allowed to access, but... way too much effort for an MVP, probably best left for future work.

PR #2309, opened 2023-05-31
Add ability to disable team-judge feedback and judge-judge feedback independently

### LINKED ISSUE #1683: Option to make teams unable to submit feedback
Is there currently a way to make it so teams can't submit feedback while adjudicators can? It'd be useful to be able to send everyone their Private URLs while restricting feedback to adjudicators.

### LINKED ISSUE #1873: No obvious way to prevent judges from submitting feedback, but allow teams to.
At UADC, as I understand it, teams can submit feedback on judges, but judges cannot submit feedback on judges.

There does not seem to be an obvious way in tabbycat to set this up. 

Things I (and the Tab Director) tried:

> - playing around in Setup > Configuration but there doesn't seem to be that flexibility 
> - playing around in the database with limiting the feedback questions to only debaters, but that still leaves the default question that I can't seem to turn off for the judges

It does not seem possible to only allow feedback from teams, and not from judges. This may be a useful feature in the future.

Fortunately, there is a workaround. However, it is truly the definition of turning a bug into a feature.

To prevent people from submitting feedback, however, it is possible to create an impossible question by:
 - Creating a question for judges only
 - Setting it to be an `integer (textbox)` question, with a minimum *larger* than the maximum possible value.

An example is below:
image

This makes it impossible to submit judge feedback.

Hopefully there's a better way to do this, but if not I wanted to share the workaround until a proper fix can be added. Unfortunately I've retired from debating, so I'm unlikely to be able to patch TabbyCat.

daankoning:
closes #1683 and closes #1873

### DIFF
--- a/tabbycat/adjfeedback/models.py
+++ b/tabbycat/adjfeedback/models.py
@@ -256,7 +256,7 @@ class AdjudicatorFeedback(Submission):
 
     def _unique_unconfirm_args(self):
         kwargs = super()._unique_unconfirm_args()
-        if self.source_team is not None and self.source_team.debate.round.tournament.pref('feedback_from_teams') != 'all-adjs':
+        if self.source_team is not None and self.source_team.debate.round.tournament.pref('feedback_from_teams') == 'orallist':
             kwargs.pop('adjudicator')
         return kwargs
 
--- a/tabbycat/adjfeedback/dbutils.py
+++ b/tabbycat/adjfeedback/dbutils.py
@@ -100,8 +100,10 @@ def add_feedback(debate, submitter_type, user, probability=1.0, discarded=False,
 
     if debate.round.tournament.pref('feedback_from_teams') == 'all-adjs':
         sources_and_subjects = [(team, adj) for team in debate.teams for adj in debate.adjudicators.all()]
-    else:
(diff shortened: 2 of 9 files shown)


tienne-B:
Oh, sorry for ignoring this! I'm somewhat far from home right now. I think we're on the right path, but the preference should also affect the "Feedback Progress" pages, and public/assistant/admin feedback submission pages.

I also don't see how this resolves #1683, as that would require a new setting for the `feedback_from_teams` preference (right below the adjudicator one).

tabbycat/adjfeedback/models.py
```diff
@@ -256,7 +256,7 @@ def __str__(self):
 
     def _unique_unconfirm_args(self):
         kwargs = super()._unique_unconfirm_args()
-        if self.source_team is not None and self.source_team.debate.round.tournament.pref('feedback_from_teams') != 'all-adjs':
+        if self.source_team is not None and self.source_team.debate.round.tournament.pref('feedback_from_teams') == 'orallist':
```
tienne-B, line 259: Not quite sure about this... if the setting is for no one, then we shouldn't even reach here as no options should appear.

tienne-B (approved):
Sorry for the long wait!

PR #2310, opened 2023-06-11
Correct ballot stats widths in future rounds

linh-trant:
This commit fixes the ballot stats graphs in result entry for rounds with no draw. The graphs had stubs for each ballot status, due to a faulty calculation when there are no debates to count from. Now, the graphs will not have any status shown.


tienne-B:
Superseded by #2311

PR #2312, opened 2023-06-12
Update django-channels to the latest version

tienne-B:
This commit updates the django-channels npm package to 2.1.3 (or non- breaking higher versions) to suppress an npm warning that it gives.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "": {
       "name": "tabbycat",
       "version": "3.18.1",
-      "license": "GNU Affero General Public License v3.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@babel/cli": "^7.16.0",
         "@babel/core": "^7.16.0",
@@ -24,7 +24,7 @@
         "core-js": "^3.19.2",
         "cpx": "1.5.x",
         "d3": "^6.3.1",
-        "django-channels": "1.1.8",
+        "django-channels": "^2.1.3",
         "feather-icons": "^4.28.0",
         "inter-ui": "^3.15.0",
         "jquery": "^3.6.0",
@@ -45,10 +45,6 @@
       },
       "devDependencies": {
         "eslint-plugin-html": "^6.2.0"
-      },
-      "engines": {
-        "node": "16.15.x",
-        "npm": "8.11.x"
       }
     },
     "node_modules/@achrinza/node-ipc": {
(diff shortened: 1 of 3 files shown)


philipbelesky:
It doesn't look like anything has changed since I last tried bumping this, which is that our use of channels doesn't seem compatible with the 2.X of these package. It might not be too much work to patch, or write our own equivalent, but I think that time would probably be better spent just shifting off of channels and instead adopting the async methods inbuilt to django proper. 

Issues I noticed:

If I release a draw I get:

`Broadcasting notification of ActionLogEntryConsumer
2023-06-17 16:41:57,523 - asyncio - ERROR - Task exception was never retrieved
future: <Task finished name='Task-116' coro=<Connection.disconnect() done, defined at /Users/philip/.local/share/virtualenvs/tc_dev-ToVGcadc/lib/python3.9/site-packages/redis/asyncio/connection.py:723> exception=RuntimeError('Event loop is closed')>
  File "/Users/philip/.local/share/virtualenvs/tc_dev-ToVGcadc/lib/python3.9/site-packages/redis/asyncio/connection.py", line 732, in disconnect
    self._writer.close()  # type: ignore[union-attr]
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/streams.py", line 353, in close
    return self._transport.close()
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/selector_events.py", line 698, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 751, in call_soon
    self._check_closed()
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed`

The Latest Actions area will show just `send_json` rather than the action details (and won't persist after a reload).

Changes to Edit Adjudicators will save/persist, but won't live update across separate browser instances.

tienne-B:
> It doesn't look like anything has changed since I last tried bumping this, which is that our use of channels doesn't seem compatible with the 2.X of these package. It might not be too much work to patch, or write our own equivalent, but I think that time would probably be better spent just shifting off of channels and instead adopting the async methods inbuilt to django proper.
>
> If I release a draw I get: [...]

Is this related to the backend `channels` package? Updating this frontend package has no incidence on how messages are sent from the backend.

> The Latest Actions area will show just send_json rather than the action details (and won't persist after a reload).
>
> Changes to Edit Adjudicators will save/persist, but won't live update across separate browser instances.

Fixed.

philipbelesky:
> Is this related to the backend channels package? Updating this frontend package has no incidence on how messages are sent from the backend.

Yea, from memory, this package was the blessed solution for receiving the django channels updates on the front-end. It's probably not doing anything particularly complex, but I didn't have much experience working with sockets and the like at the time.

I've confirmed those fixes and will approve. It might be worth keeping an eye on things on Calico whenever this change lands there — I'm not sure if I recall all of the issues that meant this bump hadn't been made before.

philipbelesky (approved):

PR #2313, opened 2023-06-17
Add Private URL preset to tournament configure

CreatedBySeb:
Added the existing Private URLs preset to the tournament configure page similar to Enable Public Features, since they are common settings that many tournaments use. As part of this, I removed the `public_checkins_submit` (allow participant self-check in permission) since this doesn't seem in line with the description of the preset and is also generally not a preferred behaviour (most tournaments online or in person want to manage check ins to avoid people checking in when they are not actually present).

I also swapped the settings on the configure page to use `save_preset` rather than re-implementing the logic in that helper function to reduce the duplication and make it a bit more readable. The removal of a space before the equals in the preset is just to bring the equals in line with every other equals in that file (previously they were slightly offset which just bugged me a little).

### DIFF
--- a/tabbycat/options/presets.py
+++ b/tabbycat/options/presets.py
@@ -29,8 +29,14 @@ def presets_for_form():
 
 
 def public_presets_for_form():
-    return [(_('Public Information Options'), _('Enable Public Information')),
+    return [(True, _('Enable Public Information')),
             (False, _('Disable Public Information'))]
+
+
+def data_entry_presets_for_form():
+    return [(False, _('Disabled (tab staff only)')),
+            ("private-urls", _('Use private URLs')),
+            ("public", _('Use publicly accessible form'))]
 
 
 def get_preferences_data(selected_preset, tournament):
@@ -363,6 +369,14 @@ class PrivateURLS(PreferencesPreset):
     show_in_list = False
     description = _("Enables participant data entry through private URLs.")
 
-    data_entry__participant_ballots             = 'private-urls'
-    data_entry__participant_feedback            = 'private-urls'
-    data_entry__public_checkins_submit          = True
+    data_entry__participant_ballots            = 'private-urls'
+    data_entry__participant_feedback           = 'private-urls'
+
+
+class PublicForms(PreferencesPreset):
+    name = _("Use Public Forms")
+    show_in_list = False
+    description = _("Enables participant data entry through public forms.")
+
+    data_entry__participant_ballots            = 'public'
+    data_entry__participant_feedback           = 'public'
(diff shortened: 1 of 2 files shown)


tabbycat/tournaments/forms.py
```diff
@@ -92,6 +92,11 @@ class Meta:
         help_text=_("Show non-sensitive information on the public-facing side of this site, "
             "like draws (once released) and the motions of previous rounds"))
 
+    private_urls = ChoiceField(
+        choices=private_url_presets_for_form(), # Tuple with (Present_Index, Preset_Name)
+        label=_("Use Private URLs"),
+        help_text=_("Enables participant data entry through private URLs."))
+
```
tienne-B, line 99: I think we should generalize this field, to answer the question "How will submissions be made for the tournament" for which there are 3 answers which would form the options:
* Tab team only
* Private URLs
* Public forms

Tab team only is the default, so would be the default, and `False` as the preset "key". Then, private URLs which you have implemented. Public forms should have a preset as well to complete these options. I'd rename this field and its attributes...

BTW, typo in the comment and it's not really true anyways, as the tuple is (preset name, name for the form)

tienne-B (changes requested):

tabbycat/options/presets.py
```diff
@@ -33,6 +33,11 @@ def public_presets_for_form():
             (False, _('Disable Public Information'))]
 
 
+def private_url_presets_for_form():
+    return [(_('Use Private URLs'), _('Enable Private URL Submissions')),
+            (False, _('Disable Private URL Submissions'))]
```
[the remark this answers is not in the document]
CreatedBySeb: Yes I agree, I found it pretty confusing as an implementation but wanted to stay consistent. Happy to try to fix both parts instead.

tabbycat/tournaments/forms.py
CreatedBySeb: This makes sense, I was sticking with the preset, but instead I could add another preset for the public forms, and then implement it as three choices. The comment was just carried over from the public_info code which I duplicated, happy to remove this when resolving the first comment

tabbycat/options/presets.py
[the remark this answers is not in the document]
CreatedBySeb: This is fixed up now for both settings

tabbycat/tournaments/forms.py
CreatedBySeb: Additional setting has been implemented and the names have been changed

tabbycat/tournaments/forms.py
```diff
@@ -110,17 +116,19 @@ def save(self):
...
-                t.preferences[preference['key']] = preference['new_value']
+        if do_public:
+            save_presets(t, PublicInformation)
+
+        # Apply data entry method preset
+        data_entry_method = self.cleaned_data["data_entry"]
+        if data_entry_method == "private-urls":
+            save_presets(t, PrivateURLS)
+        elif data_entry_method == "public":
+            save_presets(t, PublicForms)
```
tienne-B, line 131:
```suggestion
        if data_entry_method:
            save_presets(t, {"private-urls": PrivateURLS, "public": PublicForms}[data_entry_method])
```
Easier to add new keys than to make a new `if`.

PR #2316, opened 2023-06-17
Add silent toggle to complete round page

CreatedBySeb:
Adds a button to the top of the complete round which allows you to toggle the silent status of a round without having to venture into the database.


tabbycat/tournaments/views.py
```diff
@@ -130,9 +130,23 @@ def get_context_data(self, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        self.round.silent = not self.round.silent
+        self.round.save()
+
+        if self.round.silent:
+            messages.success(request, _("%(round)s has been marked as silent.") % {'round': self.round.name})
+        else:
+            messages.success(request, _("%(round)s has been unmarked as silent.") % {'round': self.round.name})
+
+        return redirect_round('tournament-complete-round-check', self.round)
```
tienne-B, line 147: Can set `round_redirect_pattern_name = 'tournament-complete-round-check'` as a class attribute, then return `super` rather than the redirect directly.

tienne-B:
Looks good! Could have a warning like "The round was already marked as _x_" if the state doesn't match the object, but that could be a follow-up.

PR #2325, opened 2023-07-01
Refactor parts of options

tienne-B:
Big one, but split into a couple commits.


tabbycat/options/templates/preset_edit.html
czlee, line 1: imho we should add some explanatory text at the top of the page to explain what's going on, like (but using better phrasing than the following lol)
- applying the preset will modify some preferences
- you can review and modify the modified preferences on this page before saving
- not all preferences are affected by the preset and the ones that aren't aren't listed on this page
- presets are **provided as convenience only**, **please double-check everything** to ensure it matches your tournament's needs

czlee (approved):
Love it. Looks great.

Happy to defer to you and @philipbelesky on whether it's okay to remove the old values from the confirmation page.

tabbycat/options/tests/management/commands/test_applypreset.py
```diff
@@ -0,0 +1,45 @@
...
+    scoring__score_min = 70
+    scoring__score_max = 80
+
+
+class ApplyPresetTests(TestCase):
+    @patch('options.management.commands.applypreset.all_presets', return_value=[TestPreset])
+    def test_preset_options(self, mock_all_presets):
+        with self.assertRaises(CommandError) as cm:
+            Command().create_parser('', '').parse_args(['notvalidpreset'])
+            self.assertEqual(str(cm), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
```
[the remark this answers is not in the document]
tienne-B: No, the `assertRaises` acts as a `try/catch`, and we need the `assertEqual` to have the contents of the exception, which would only be available in the `with` context.

tabbycat/options/utils.py
```diff
@@ -29,6 +33,13 @@ def use_team_code_names_data_entry(tournament, tabroom):
...
+
+
+def validate_metric_duplicates(generator, value):
+    # Check that non-repeatable metrics aren't listed twice
+    classes = [generator.metric_annotator_classes[metric] for metric in value]
+    duplicates = [c for c in classes if not c.repeatable and classes.count(c) > 1]
+    if duplicates:
+        duplicates_str = ", ".join(list(set(force_str(c.name) for c in duplicates)))
+        raise ValidationError(_("The following metrics can't be listed twice: "
+                "%(duplicates)s") % {'duplicates': duplicates_str})
```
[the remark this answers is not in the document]
tienne-B: I'll leave this for now, as it should actually be a part of #2319, as it can interact between settings.

tabbycat/options/templates/preset_edit.html
tienne-B: Reaching that template requires going through the Configuration index which has this message about presets:
> These show changes to your configuration for common situations and debate formats. You may apply these settings or modify them for your tournament directly on the preset's page.

Would you have a better phrasing here? On the preset page itself I find the first 3 points very clear, by having the settings organised (with headers "will be modified") and editable.

tabbycat/options/tests/management/commands/test_applypreset.py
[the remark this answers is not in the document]
czlee: I don't think that's how it works? The example here definitely uses `cm` outside the context:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

``` python
with self.assertRaises(SomeException) as cm:
    do_something()

the_exception = cm.exception
self.assertEqual(the_exception.error_code, 3)
```

Otherwise where would the equivalent `try ... except` block go? The context manager can't tell the difference between anything inside the block—it can't differentiate between the first and second lines, in particular, because the `yield` statement can only go through the whole block. So it would have to translate to
``` python
try:
    Command().create_parser('', '').parse_args(['notvalidpreset'])
    # execution never reaches this line if the above line raises an error
    self.assertEqual(str(cm), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
except CommandError as e:
    # what would it do with the error???
```
My understanding is that the `assertRaises()` context manager is equivalent to
``` python
try:
    Command().create_parser('', '').parse_args(['notvalidpreset'])
    # I think this is the wrong place...
    # self.assertEqual(str(cm), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
except CommandError as cm:
    cm.exception = e

# Now you can use cm.exception
self.assertEqual(str(cm.exception), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
```

----------------------

If you don't believe me, the easiest way to test this is to modify the expected error message to be wrong:
``` suggestion
            self.assertEqual(str(cm), "This is not the expected error message")
```
Try that, and you'll find that both tests (this and the one below) incorrectly pass.

Try unindenting this line:
```suggestion
        self.assertEqual(str(cm), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
```
and I get this:
```
AssertionError: '<unittest.case._AssertRaisesContext obje[17 chars]af0>' != "Error: argument preset: invalid choice: [39 chars]et')"
- <unittest.case._AssertRaisesContext object at 0x7fc3e3bd8af0>
+ Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')
```

I think this is because you're meant to use `cm.exception`:
```suggestion
            self.assertEqual(str(cm.exception), "Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')")
```
This passes.

And to be sure, change it to the wrong message:
```suggestion
            self.assertEqual(str(cm), "This is not the expected error message")
```
And this fails with:
```
AssertionError: "Error: argument preset: invalid choice: [39 chars]et')" != 'This is not the expected error message'
- Error: argument preset: invalid choice: 'notvalidpreset' (choose from 'testpreset')
+ This is not the expected error message
```
which is the expected failure.

tabbycat/options/templates/preset_edit.html
czlee: I think I'm just working on more of a "some new tab directors will need things really spelt out for them" basis. I agree that the first two can be argued to be clear, though I think for something like this it adds some security to have the page explain what you're supposed to do with it. The third one isn't as clear—how do you know that the presented list isn't exhaustive?

But it's the fourth point that I really worry about. It's basically a disclaimer, "don't rely on the software to do your job for you, check what you need yourself". Something like:
> Note: Presets are provided **as a convenience only** and **may not be up to date** for your tournament. **Please double-check everything** to ensure it matches your tournament's needs.

I mean, you'd hope a disclaimer like that wouldn't be necessary, because common sense would say this. But... sometimes... y'know...

philipbelesky: Slight tweak to clarify what 'up to date' might mean?

> Note: Presets are provided as a convenience only and may not be up to date with a debating format's rules or the expectations of your circuit. Please double-check everything to ensure it matches your needs.

PR #2327, opened 2023-07-02
Overhaul merging ballots

tienne-B:
This commit implements a merging strategy for consensus ballots, making sure that speaker order, ghosts, as well as declared winners and speaker scores, are identical, lest an error be raised.

Now, rather than aborting the merge when an error is detected, the error(s) are applied to the new ballot, blanking the fields that were errored. It also shows more general errors of the detected divergences, rather than pointing out where it thinks the error was made. That wasn't needed anymore as the specific fields are now obvious.

### DIFF
--- a/tabbycat/results/result.py
+++ b/tabbycat/results/result.py
@@ -42,6 +42,7 @@ import logging
 from functools import wraps
 from itertools import product
 from statistics import mean
+from typing import TYPE_CHECKING
 
 from adjallocation.allocation import AdjudicatorAllocation
 from adjallocation.models import DebateAdjudicator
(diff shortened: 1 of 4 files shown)


tabbycat/results/result.py
```diff
@@ -680,6 +698,23 @@ def identical(self, other):
...
+            if cur_speaker is None:
+                self.set_speaker(side, pos, result.get_speaker(side, pos))
+            elif result.get_speaker(side, pos) != cur_speaker:
+                errors.append(ResultError("Inconsistent speaker order", "speaker", side, pos))
+                continue  # Don't care about setting ghost/score if can't attribute to the correct speaker
+
+            if not self.get_ghost(side, pos):
+                self.set_ghost(side, pos, result.get_ghost(side, pos))
+            elif not result.get_ghost(side, pos):
+                errors.append(ResultError("Inconsistent ghost order", "ghost", side, pos))
```
czlee, line 714: If `self.get_ghost(side, pos) == False` and `result.get_ghost(side, pos) == True`, then this will have effect of setting the ghost to `True`, not logging the mismatch as an error. Is this the intended behaviour?

czlee (commented):
haven't had a complete look yet, but a couple of things I noticed

tabbycat/results/result.py
tienne-B: I'm okay with it, as iron policy is mostly set by the tournament, is minor, and is the most often wrongly set, so that's the field that would be easiest to correct.

PR #2349, opened 2023-09-04
Implement team byes in draws

### LINKED ISSUE #2282: Byes for 2-team formats
This is a brief description of how to determine which team gets the bye in APDA tournaments: the work here should probably work for any 2v2 system.

**Byes**

This should be an optional setting: some tournaments might opt to give out a Bye, while others pair in a swing team into their tournament to fix an odd number of teams in the draw. 
The bye is allocated to the bottom team in tab, provided it has not gotten a bye before. If it has, we go to the second-lowest team, and so on. If there are multiple teams on the same record, speaks, and ranks, tied for the bottom performing team, then it is allocated randomly to any one of them. 
The team that gets the bye should be assigned their average speaks and ranks for the round they missed. So, if they get the bye in R3, they receive the average of R1 and R2 for pairing for R4, and the average of R1,R2 and R4 for pairing in R5. 
A team which gets the bye needs to be displayed on the draw, so that they know that they are not debating that round. 

An optional capability could be perhaps to allot the bye randomly in the lowest bracket, or randomly overall: APDA tournaments do not do this, but it seems within the realm of possibility of people wanting to do this.

tienne-B:
This commit implements teams being marked as being given a bye for a round. To support this, placeholder debates and ballots are created for them, which bypass results. Rather, they are treated as not a debate when getting their ballot, and links to them are hidden. In draw tables, their situation is highlighted, and by default moved to the bottom of the table.

The creation of bye teams and debates is handled by the draw manager, which is now given a couple more options from the (draw rules) preferences for how to select byes and how to treat them:
- Bye selection: How teams are chosen to get the bye
  - Random: Any team could get the bye,
  - Lowest: The lowest-ranking team (if supported by the draw) is excluded.
- Bye results: What type of results would we give bye teams
  - None: Skip creating ballots for them, equivalent to a loss
  - Points: Give them a winning TeamScore without speaks

As to not tank bye teams within their bracket, average speaks should be used rather than sum of speaks as a tiebreaker.

Used this opportunity to also refactor the affected model choices into enums.

Closes #2282


tabbycat/draw/manager.py
```diff
@@ -220,7 +265,16 @@ def get_teams(self):
                 setattr(team, pullup_metric, standing.metrics[pullup_metric])
             ranked.append(team)
 
-        return ranked
+        n_byes = self.n_byes(len(ranked))
+        if n_byes:
+            if self.round.tournament.pref('bye_team_selection') == 'random':
+                byes = []
+                for i in range(n_byes):
+                    byes.append(ranked.pop(random.randrange(len(ranked))))
```
czlee, line 273: Another option could be to use `random.sample(n_byes)` (though you'd still have to loop through the result to `pop` them from `ranked`) (and similarly in `SeededDrawManager`).

tabbycat/draw/manager.py
```diff
@@ -220,7 +265,16 @@ def get_teams(self):
...
+                return ranked, byes
+            return ranked[:-n_byes], ranked[-n_byes:]
```
czlee, line 275: I'd recommend writing the `elif` case explicitly and raising an error to fail fast if the preference isn't recognised (so that if in future someone adds a fourth option they don't miss this spot here).

tabbycat/draw/tables.py
czlee, line 1: I get a `NoDebateTeamFoundError` on line 124 just after generating a draw that has a bye in it.

czlee (commented):
Very late to the party, sorry: I think the general approach is good. But this breaks the assumption that every side is populated in every debate, which is made in a lot of views/tables. I notice you've got some handling for this in draw/tables.py, but it crashed on me in a different place, on line 124 of draw/tables.py.

(To be clear, I think we should break this assumption for byes, and handle it everywhere.)

I don't really have a fantastic solution here, but here a couple of ideas (intended as alternatives to each other):
1. Patch up all the spots as you find them. The whole tables framework isn't very elegant or well-abstracted, and it'd be nice to do up but also a lot of work.
2. Rewrite `Debate._populate_teams()` and friends to handle bye debates explicitly, so that `DebateTeam.get_side()` will only raise the `NoDebateTeamFoundError` when it's sensible to do so (_i.e._ when it's not a bye debate), otherwise it returns some sort of vacuous object (like `None`). The catch here is that all the views/tables will then have to learn to handle said vacuous object. Maybe this is okay but it could just mean a ton of `AttributeError`s.

Fantastic stuff 🎉

PR #2364, opened 2023-11-18
Tests: Properly exclude later rounds from metrics

### LINKED ISSUE #1466: Draw strength counts opponents without confirmed results
It's not clear to me what the behaviour should be, just something I noticed when writing unit tests.

If there is a debate with only unconfirmed ballots, then this debate _doesn't_ count towards the second-order metrics used in calculating draw strength, but the fact that it is drawn _does_ count for the purposes of determining who a team's opponents have been when calculating draw strength. This is arguably a strange inconsistency—we are counting their opponents' record in a team's draw strength, even when it isn't fully accounted for that they faced the opponent.

For now, I've marked these as expected failures in 78ca17f1a765d062dff9be97fda98b3704829cfc.

tienne-B:
This commit fixes the improper counting of draw strength statistics in the currently ignored tests. The problem was that the test did not exclude the results of Round 3, created to be ignored.

The stats do not count unconfirmed results, but do include the teams they've gone against. That should be the expected outcome, like if power-pairing is lagged.

Fixes #1466

### DIFF
--- a/tabbycat/standings/tests/test_standings.py
+++ b/tabbycat/standings/tests/test_standings.py
@@ -1,5 +1,4 @@
 import logging
-from unittest import expectedFailure
 
 from django.test import TestCase
 
@@ -253,13 +252,15 @@ class TestStandingsWithUnconfirmedBallotSubmission(IgnorableDebateMixin, TestTri
         debate = self.set_up_ignorable_debate()
         debate.ballotsubmission_set.update(confirmed=False)
 
-    @expectedFailure
     def test_draw_strength(self):
-        super().test_draw_strength()
+        # The ignorable debate still counts as an opponent (because they definitely faced each
+        # other) but does not count as a win (because it's unconfirmed). So the losing team has
+        # faced the winning team 3 times (including ignorable debate), and the winning team has 2
+        # wins (excluding ignorable debate), for a losing team draw strength of 3 * 2 = 6.
+        self._base_metric_test({'draw_strength': [0, 6]})
 
-    @expectedFailure
     def test_draw_strength_speaks(self):
-        super().test_draw_strength_speaks()
+        self._base_metric_test({'draw_strength_speaks': [591, 609]})
 
 
 class TestBasicStandings(TestCase):


tabbycat/standings/tests/test_standings.py
```diff
@@ -45,7 +44,7 @@ def tearDown(self):
 
     def get_standings(self, generator):
         with suppress_logs('standings.metrics', logging.INFO):
-            standings = generator.generate(self.tournament.team_set.all())
+            standings = generator.generate(self.tournament.team_set.all(), round=Round.objects.get(seq=2))
```
czlee, line 47: Oh, interesting. But I think this would make all tests with the ignorable debate vacuously pass—in particular, `TestStandingsWithEliminationRound` (the other subclass of `IgnorableDebateMixin`) would pass solely because of this `round=` argument. Even if the elimination round filter failed, the `round=` filter would make it look like it had worked.

Similarly, all the tests in `TestStandingsWithUnconfirmedBallotSubmission` are specifically intended to test that unconfirmed ballots are excluded _because they're unconfirmed_, and this would make all other tests inherited from `TestTrivialStandings` vacuously pass too.

I had to think for a while (again!) about whether #1446 is expected behaviour, and I went back and forth three times 😂 but I think you're right? The unconfirmed ballot should not count in draw strength, but the confirmed results of their opponents from all rounds should count, is that what you're saying? And callers should pass in the `round=` argument if they want unfinished rounds to be excluded, as `DrawManager` already does?

If that's the case then I would instead override `test_draw_strength()` and `test_draw_strength_speaks()` with the new expected results, and add a comment to explain why this is the expected result. So (without having actually tried it) for `test_draw_strength()` it would be `[0, 6]`, and for `test_draw_strength_speaks()` it would be `[591, 609]`, I think? Thoughts?

tabbycat/standings/tests/test_standings.py
```diff
@@ -253,11 +252,9 @@ def setUp(self):
         debate = self.set_up_ignorable_debate()
         debate.ballotsubmission_set.update(confirmed=False)
 
-    @expectedFailure
     def test_draw_strength(self):
         super().test_draw_strength()
```
czlee, line 256: So if we overrode this "properly", it would be:
```suggestion
    def test_draw_strength(self):
        # The ignorable debate still counts as an opponent (because they definitely faced each
        # other) but does not count as a win (because it's unconfirmed). So the losing team has
        # faced the winning team 3 times (including ignorable debate), and the winning team has 2
        # wins (excluding ignorable debate), for a losing team draw strength of 3 * 2 = 6.
        self._base_metric_test({'draw_strength': [0, 6]})
```

tabbycat/standings/tests/test_standings.py
```diff
@@ -253,13 +252,15 @@ def setUp(self):
...
+        # The ignorable debate still counts as an opponent (because they definitely faced each
+        # other) but does not count as a win (because it's unconfirmed). So the losing team has
+        # faced the winning team 3 times (including ignorable debate), and the winning team has 2
+        # wins (excluding ignorable debate), for a losing team draw strength of 3 * 2 = 6.
+        self._base_metric_test({'draw_strength': [0, 6]})
 
-    @expectedFailure
     def test_draw_strength_speaks(self):
-        super().test_draw_strength_speaks()
+        self._base_metric_test({'draw_strength_speaks': [591, 609]})
```
czlee, line 263: I'd include a comment here too (even if it's just a reference to the above one, like "Similar logic to `test_draw_strength` above".

czlee (approved):
Awesome thanks!

PR #2365, opened 2023-11-25
Support selecting multiple scroll speeds on draws page (Fixes #2360)

### LINKED ISSUE #2360: The 'Stop Scrolling' button does not work when multiple speeds are selected in the Draws page
Step to reproduce the bug:
1. Generate draws for any round (AP tournament in our case)
2. Go to the draws page from your private url
3. In the button group where there are multiple speeds 
image
4. Click on multiple speeds one after the other. For example press 'Extra slow' then 'Slow'. You can press more also but 2 are enough to show the bug.
5. Now click on the stop scrolling button that appears on the top right
image
6. The button disappears but the page continues to scroll. This is the bug I am talking about. The page continues to scroll and there is apparently no way to stop it without refreshing the page.

Installation: Calicotab
Browser used: Firefox

jalaj711:
This PR fixes #2360 where selecting multiple scroll speeds on the draws page would cause the stop scrolling to not work.

The fix used is to simply call `stopScrolling()` before calling `startScrolling()` whenever a scroll button is clicked. This way the scroll animation is first reset to no scroll and then a new animation is run with the new speed and so stop scrolling button keeps working as if that was the only scroll speed ever selected.

### DIFF
--- a/tabbycat/draw/templates/draw_display_by.html
+++ b/tabbycat/draw/templates/draw_display_by.html
@@ -94,10 +94,6 @@
       $('#stop_scrolling').hide();
       $(".scroll-speeds .btn").click(function(event){
         var speed = $(document).height() / $(this).attr('data-target');
-        
-        // first call stop scrolling so that we stop scrolling with
-        // previosul selected speed (if any) and then start scrolling
-        // with the new speed. (issue #2360)
         stopScrolling()
         startScrolling()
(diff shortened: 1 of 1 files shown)


tienne-B:
Thank you for your PR; it's a good fix! (I removed the comment as it was annoying the linter and it's not worth a comment)

PR #2367, opened 2023-12-05
Added Mailgun installation if statement

abinash2512:
Added some code if installation uses Mailgun. SendGrid is very finnicky with getting new user accounts, for some reason neither me or a few others have been able to. Heroku offers Mailgun as a free/cheap way to send emails. Adding this means that configuring Mailgun for Tabbycat no longer requires editing the codebase.

### DIFF
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -100,13 +100,13 @@ CHANNEL_LAYERS = {
 # Email / SendGrid
 # ==============================================================================
 
-if environ.get('EMAIL_HOST', ''):
-    SERVER_EMAIL = environ['DEFAULT_FROM_EMAIL']
-    DEFAULT_FROM_EMAIL = environ['DEFAULT_FROM_EMAIL']
-    EMAIL_HOST = environ['EMAIL_HOST']
-    EMAIL_HOST_USER = environ['EMAIL_HOST_USER']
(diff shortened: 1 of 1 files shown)


tienne-B:
This PR seems incomplete or on a false pretence. There are so many email providers that choosing one to be used by default just isn't practical, and as we saw with SendGrid, prone to break at any time. The same thing could happen to Mailgun or any other. Instead of specifying one, we made it so whoever deploys Tabbycat can hook up any email provider using the provided environment variables (no need to change the code in any case). Being able to manage env vars is expected knowledge for self-hosting. The goal of this PR is trivially accomplished, seeing that it's just adding new env var keys that correspond exactly to an existing set of keys.

There could be a wider discussion to be had if you'd want Mailgun to be the default and added to Tabbycat on deployment. In that case, you'd also want to adapt the [deploy script](https://github.com/TabbycatDebate/tabbycat/blob/10198142e7e1020017b8f0a7a07787a5aca0f9f3/deploy_heroku.py) to add the add-on, and the current change isn't sufficient. I'm leaning towards opposing adding a default mail provider to let users find whichever one they prefer though, but I'm open to arguments.

tienne-B:
It's been over a month without reply and is unlikely to be merged, so I'll be closing this. Feel free to re-open it with a reply or amendment.

PR #2374, opened 2023-12-14
Update .python-version

gserranof:
The current Python version used is not longer in line with the current Heroku minimum requirements (https://devcenter.heroku.com/articles/python-support) - thus, deployment is not possible. Although Heroku recommends python-3.12.1,upgrading to 3.9.18 guarantees compatibility whith the minimum upgrade possible (as Tabbycat currently uses 3.9.12).

I've implemented it in my own tabs and no problem has arised yet - everything looks fine and deployment has going fine.


tienne-B:
@gserranof Why we aren't going all the way to 3.12? As Python should be backwards compatible, are there any packages that break with such a recent version?

tienne-B:
Feel free to re-open with an answer!

gserranof:
No special reason! Actually, I didn't even know Python back in december (and now just a little), so I had no idea whatsoever that it was backwards compatible.

PR #2385, opened 2024-01-12
Arbitrary team count

tienne-B:


tabbycat/draw/models.py
```diff
@@ -183,27 +187,31 @@ def debateteams_ordered(self):
...
+    oo_team = _team_property('1_team')
+    cg_team = _team_property('2_team')
+    co_team = _team_property('3_team')
+    aff_dt = _team_property('0_dt')
+    neg_dt = _team_property('1_dt')
+    bye_dt = _team_property('-1_dt')
+    og_dt = _team_property('0_dt')
+    oo_dt = _team_property('1_dt')
+    cg_dt = _team_property('2_dt')
+    co_dt = _team_property('3_dt')
```
czlee, line 208: I think originally I wrote `_team_property` to work on the assumption that `attr` (the argument to `_team_property`) would be a valid attribute name. That's obviously no longer true and I guess it doesn't need to be. So if you like you're welcome to change the naming convention to e.g. `team_0`, `dt_0`, _etc._, if you'd find that more intuitive.

tabbycat/motions/statistics.py
czlee, line 1: I suppose it's kind of interesting to me that you chose to convert the attribute names in this context (where in context it presumably always refers to aff/neg or og/oo/cg/og) to `s0`, `s1`, _etc._, rather than have the logic convert the integer into a descriptive side label. I think I would've done it the other way round (just for this part, given where it gets used). But up to you ☺️

tabbycat/options/preferences.py
```diff
@@ -494,16 +494,12 @@ class FeedbackIntroduction(LongStringPreference):
 
 
 @tournament_preferences_registry.register
-class TeamsInDebate(ChoicePreference):
-    help_text = _("Two-team format (e.g. Australs, WSDC) or British Parliamentary")
-    verbose_name = _("Teams in debate")
+class TeamsInDebate(IntegerPreference):
+    help_text = _("How many teams are in each debate (normally 2 for AP or 4 for BP)")
```
czlee, line 502: What does "AP" mean? 😅 

(I'd just say "most debate styles" but maybe that'd be controversial lol)

tabbycat/results/forms.py
```diff
@@ -286,7 +286,7 @@ def get_preferences_options(self):
 
     @staticmethod
     def _fieldname_motion_veto(side):
-        return '%(side)s_motion_veto' % {'side': side}
+        return '%(side)d_motion_veto' % {'side': side}
```
czlee, line 266: I'm guessing you tried it and these don't need to start with an alphabetic character. But I'd maybe prefix with `s` anyway so it's (slightly) less confusing to future developers?

tabbycat/importer/importers/anorak.py
```diff
@@ -53,8 +53,8 @@ class AnorakTournamentDataImporter(BaseTournamentDataImporter):
     })
 
     lookup_team_position = make_lookup("team position", {
-        ("affirmative", "aff", "a"): dm.DebateTeam.Side.AFF,
-        ("negative", "neg", "n"): dm.DebateTeam.Side.NEG,
+        ("affirmative", "aff", "a"): 0,
+        ("negative", "neg", "n"): 1,
```
czlee, line 57: Did you want to use `DebateSide.AFF`, `DebateSide.NEG` here?

tabbycat/standings/tests/test_standings.py
```diff
@@ -29,8 +29,8 @@ def setUp(self):
         for i in [1, 2]:
             rd = Round.objects.create(tournament=self.tournament, seq=i)
             debate = Debate.objects.create(round=rd, flags=['pullup'])
-            dt1 = DebateTeam.objects.create(debate=debate, team=self.team1, side=DebateTeam.Side.AFF, flags=['pullup'])
-            dt2 = DebateTeam.objects.create(debate=debate, team=self.team2, side=DebateTeam.Side.NEG)
+            dt1 = DebateTeam.objects.create(debate=debate, team=self.team1, side=0, flags=['pullup'])
+            dt2 = DebateTeam.objects.create(debate=debate, team=self.team2, side=1)
```
czlee, line 33: `DebateSide.AFF`/`NEG`?

tabbycat/tournaments/tests/test_round_checks.py
```diff
@@ -24,8 +24,8 @@ def setUp(self):
             team1 = Team.objects.create(tournament=self.tournament, reference=f"Team {i}A", use_institution_prefix=False)
             team2 = Team.objects.create(tournament=self.tournament, reference=f"Team {i}B", use_institution_prefix=False)
             adj = Adjudicator.objects.create(tournament=self.tournament, name=f"Adjudicator {i}")
-            DebateTeam.objects.create(debate=debate, team=team1, side=DebateTeam.Side.AFF)
-            DebateTeam.objects.create(debate=debate, team=team2, side=DebateTeam.Side.NEG)
+            DebateTeam.objects.create(debate=debate, team=team1, side=0)
+            DebateTeam.objects.create(debate=debate, team=team2, side=1)
```
czlee, line 28: `DebateSide.AFF`/`NEG`?

tabbycat/draw/generator/bpelimination.py
```diff
@@ -20,7 +20,7 @@ def _four_way_fold(self, teams, start_rank=0):
         pools = (teams[0:n], teams[n:2*n][::-1], teams[2*n:3*n], teams[3*n:4*n][::-1])  # Reverse 1 & 3
         pairings = list()
         for i, ts in enumerate(zip(*pools), start=start_rank+1):
-            pairing = BPPairing(ts, bracket=0, room_rank=i)
+            pairing = PolyPairing(ts, bracket=0, room_rank=i, num_sides=self.teams_in_debate)
```
czlee, line 23: Interesting—in this context wouldn't `num_sides` always be 4? (Since you still have the `len(teams) % 4 == 0` check in line 16?) Why not just hard-code in this case?

tabbycat/draw/generator/common.py
```diff
@@ -170,7 +171,7 @@ class BasePairDrawGenerator(BaseDrawGenerator):
         "avoid_conflicts"       : "off",
     }
 
-    TEAMS_PER_DEBATE = 2
+    teams_in_debate = 2
```
czlee, line 174: I think if you set `self.teams_in_debate` to an instance attribute in the constructor (line 48), that takes precedence over the class attribute (here), right? So this effectively now does nothing? (I see you deleted it from `BaseBPDrawGenerator`.)

tabbycat/draw/generator/common.py
```diff
@@ -40,16 +40,17 @@ class BaseDrawGenerator:
     requires_prev_results = False
     requires_rrseq = False
 
-    def __init__(self, teams, results=None, rrseq=None, **kwargs):
+    def __init__(self, teams, results=None, rrseq=None, teams_in_debate=2, **kwargs):
```
czlee, line 43: [DRAFT COMMENT — this is partially developed thinking, but maybe it's useful or maybe you have thoughts on it already 🙃]

In fact… I don't think I'd recommend setting `teams_in_debate` as a constructor argument/instance attribute.

Given subclasses like `BaseBPDrawGenerator`, `BasePairDrawGenerator`, it seems to me that `teams_in_debate` should be an attribute of the class, not of the constructor.

If there's a need for a variable number of teams in a specific subclass, I'd create a mechanism for setting this in the constructor of that subclass, like:
``` py
class BaseManyTeamsDrawGenerator:
  def __init__(self, ..., teams_in_debate):
    self._teams_in_debate = teams_in_debate

  @property
  def teams_in_debate(self):
    return self._teams_in_debate
```
(or just set `self.teams_in_debate` in the constructor, understanding that it will override the class attribute.)

(If there's not really a need for subclasses anymore, then I guess ditch them?)

tabbycat/draw/generator/pairing.py
```diff
@@ -172,27 +173,24 @@ def winner(self):
...
-    """Pairing class for British Parliamentary."""
-
-    sides = ['og', 'oo', 'cg', 'co']
+class PolyPairing(BasePairing):
+    """Pairing class for British Parliamentary and Public Speaking."""
 
-    def __init__(self, teams, bracket, room_rank, flags=[], team_flags={}):
+    def __init__(self, teams, bracket, room_rank, num_sides=4, flags=[], team_flags={}):
         super().__init__(teams, bracket, room_rank, flags, team_flags)
-        assert len(self.teams) == 4, "There must be four teams in a BPPairing"
```
czlee, line 182: Not worth checking that `len(self.teams) == num_sides`?

czlee (commented):
I notice that some of the time you use hard-coded `0`, `1`, and some of the time you use `DebateSide.AFF`, `DebateSide.NEG`. Did you want to consider always doing the latter?

I've looked at 83 out of 91 files so far… submitting what I have to unblock amendments.

tabbycat/motions/statistics.py
tienne-B: You're right; it makes more sense to disable motion statistics for PS and keeping these the same.

tabbycat/options/preferences.py
tienne-B: "Associated Press" /jk

Tightened up the wording.

tabbycat/draw/generator/pairing.py
tienne-B: No, as the number of sides may actually vary.

tabbycat/draw/models.py
tienne-B: I'd be more inclined to delete these and use indexes in lists of `teams` and `debateteams`.

tabbycat/motions/statistics.py
tienne-B: I remember why I had changed it to be number-based: `DebateSide.OG` is basically just an alias of `.AFF`, so it gives `'AFF'` as its name. I didn't like that, so put it all as "side _n_" to generalize.

tabbycat/draw/models.py
czlee: Fine with me, I think I originally mostly did this whole `_team_property` thing as a backwards compatibility thing, and I also didn't want to go and change every instance of `.aff_team` and `.neg_team` to an index because the code sometimes read better with it phrased that away. This makes less sense when generalising to arbitrary numbers of teams.

PR #2386, opened 2024-01-13
Set docker to reload on code changes

tienne-B:
This commit sets the volumes for web and worker to be the full directory in order for changes made in the code to immediately be reflected in the Docker instance. We can also use the COPY command rather than ADD, which gives a good performance advantage.

Further, the nginx configuration did not work as it was copied from a Ruby (.erb) template, so some syntax wasn't recognized.


philipbelesky (approved):
Thanks for updating this, the code reload works well and should provide a nice streamlined dev setup! The ruby interpretation issue duplicates a fix in b3d4c61 but should merge cleanly.

PR #2390, opened 2024-01-23
Add extra permissions and link permisions with views

linh-trant:


tabbycat/draw/views.py
```diff
@@ -571,6 +574,7 @@ class AdminDrawWithDetailsView(AdminDrawView):
     detailed = True
     page_emoji = '👀'
     use_template_subtitle = False  # Use the "for Round n" subtitle
+    view_permission = Permission.VIEW_ADMIN_DRAW
```
tienne-B, line 577: Wouldn't be necessary as it was defined in the `AdminDrawView` superclass.

tabbycat/users/permissions.py
```diff
@@ -94,17 +100,25 @@ class Permission(TextChoices):
...
+    VIEW_SPEAKER_CATEGORIES = 'view.speakercategories', _("view speaker categories")
+    EDIT_SPEAKER_CATEGORIES = 'edit.speakercategories', _("edit speaker categories")
     VIEW_BREAK_OVERVIEW = 'view.break.overview', _("view break overview")
     VIEW_BREAK = 'view.break', _("view breaks")
     GENERATE_BREAK = 'generate.break', _("generate all breaks")
 
+    VIEW_PRIVATE_URLS = 'view.privateurls', _("view private urls")
+    GENERATE_PRIVATE_URLS = 'generate.privateurls', _("generate private urls")
+    # need to get rid of generate private urls soons
+    SEND_PRIVATE_URLS = 'send.privateurls', _("send private urls")
```
tienne-B, line 120: Could capitalize, like "URLs", as done elsewhere?

tabbycat/draw/views.py
```diff
@@ -828,6 +835,7 @@ def get_table(self):
 
 
 class SideAllocationsView(AdministratorMixin, BaseSideAllocationsView):
+    edit_permission = Permission.EDIT_ALLOCATESIDES
```
tienne-B, line 838: Not defined.

Also, maybe it should become a `view_permission` as the permissions aren't hooked up to Vue pages, and there'd be no need to get to the page if not to edit. Same thing for Adj, Preformed Panel, and Venue allocation.

tabbycat/options/views.py
```diff
@@ -23,6 +24,8 @@
 
 
 class TournamentConfigIndexView(AdministratorMixin, TournamentMixin, TemplateView):
+    view_permission = Permission.VIEW_TOURNAMENTPREFERENCEMODEL
+    edit_permission = Permission.EDIT_TOURNAMENTPREFERENCEMODEL
```
tienne-B, line 28: There is no "editing" on this page, and there is more than just editing preferences on that page. I think we could leave it public, with more precise permissions on the links.

tabbycat/participants/views.py
```diff
@@ -69,6 +70,7 @@ def get_context_data(self, **kwargs):
 
 
 class AdminParticipantsListView(AdministratorMixin, BaseParticipantsListView):
+    view_permission = [Permission.VIEW_TEAMS, Permission.VIEW_ADJS]
```
tienne-B, line 73: I don't remember supporting a list of permissions, but it could be done easily. I'm assuming this is an `and`.

tienne-B (commented):
Looks good!

tabbycat/draw/views.py
linh-trant: I am pretty sure that it is defined in `permissions.py`, but yes; I will change to `view_permission`.

tabbycat/options/views.py
linh-trant: I quite don't understand what you meant by "leave it public", because I assume that the configuration is only available in Admin views. Do I just leave the `view_permission` there?

tabbycat/draw/views.py
tienne-B: Fair, must just have been hidden by the diff

tabbycat/options/views.py
tienne-B: Leave it without assigning any permission (edit nor view)

tabbycat/adjallocation/views.py
```diff
@@ -187,6 +193,9 @@ def formset_valid(self, formset):
 
 class AdjudicatorTeamConflictsView(BaseAdjudicatorConflictsView):
 
+    view_permission = Permission.VIEW_ADJ_TEAM_CONFLICTS
+    edit_permission = Permission.EDIT_ADJ_TEAM_CONFLICTS
+
```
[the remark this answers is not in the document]
tienne-B: This is a prompt for a discussion: do you want to be explicit here (this PR), or keep as-is?

tabbycat/users/groups.py
```diff
@@ -19,13 +19,49 @@ class BaseGroup:
...
+        Permission.EDIT_BASEJUDGESCORES_IND,
+        Permission.EDIT_DEBATEADJUDICATORS,
+        Permission.EDIT_FEEDBACK_CONFIRM,
+        Permission.EDIT_FEEDBACK_IGNORE,
+        Permission.EDIT_FEEDBACK_OVERVIEW,
+        Permission.EDIT_JUDGESCORES_BULK,
+        Permission.EDIT_MOTION,
+        Permission.EDIT_RELEASEDRAW,
+        Permission.EDIT_UNRELEASEDRAW,
+        Permission.EDIT_SETBREAKING,
```
tienne-B, line 45: Do we want them to?

tabbycat/users/groups.py
```diff
@@ -37,4 +73,36 @@ class TabDirector(BaseGroup):
...
+        Permission.VIEW_FEEDBACK_UNSUBMITTED,
+        Permission.EDIT_FEEDBACK_CONFIRM,
+        Permission.EDIT_FEEDBACK_IGNORE,
+        Permission.ADD_FEEDBACK,
+        Permission.EDIT_BREAK_ELIGIBILITY,
+        Permission.EDIT_SPEAKER_CATEGORIES,
+        Permission.VIEW_PRIVATE_URLS,
+        Permission.GENERATE_PRIVATE_URLS,
+        Permission.SEND_PRIVATE_URLS,
+    ]
```
tienne-B, line 100: Screenshot 2024-01-24 at 14 18 12

I think there are way more permissions in this list than what they can actually do currently.

tabbycat/participants/views.py
```diff
@@ -69,6 +70,7 @@ def get_context_data(self, **kwargs):
 
 
 class AdminParticipantsListView(AdministratorMixin, BaseParticipantsListView):
+    view_permission = Permission.VIEW_TEAMS and Permission.VIEW_ADJUDICATORS
```
tienne-B, line 73: This won't work either. To annotate my earlier comment:
> I don't remember supporting a list of permissions, but it could be done easily. I'm assuming this is an `and`.

Meant that you can't have a view be dependent on multiple permissions, but giving a list of permissions could be supported by editing the mixin. You'd also need to specify that _all_ (and) of the listed permissions must be present, rather than merely _any_.

I'd rather you create a new permission to access this page, rather than try to do permission "logic".

tabbycat/privateurls/views.py
```diff
@@ -149,7 +152,7 @@ def post(self, request: 'HttpRequest', *args, **kwargs) -> 'HttpResponseRedirect
 
 class EmailRandomisedUrlsView(RoleColumnMixin, TournamentTemplateEmailCreateView):
     page_subtitle = _("Private URLs")
-
+    edit_permission = Permission.SEND_PRIVATE_URLS
```
tienne-B, line 155: Should also add a view permission here, maybe a new one as this also has email addresses, so more personal details.

tabbycat/users/groups.py
```diff
@@ -37,4 +79,53 @@ class TabDirector(BaseGroup):
...
+class Language(BaseGroup):
+    name = _("Language")
+    permissions = [
+        Permission.EDIT_BREAK_ELIGIBILITY,
+        Permission.EDIT_SPEAKER_CATEGORIES,
+        Permission.VIEW_PARTICIPANTS,
+        Permission.VIEW_TEAMS,
+        Permission.VIEW_ADJUDICATORS,
+        Permission.VIEW_ROOMS,
+        Permission.VIEW_INSTITUTIONS,
```
tienne-B, line 130: Why?

tabbycat/users/groups.py
```diff
@@ -37,4 +79,53 @@ class TabDirector(BaseGroup):
...
+        Permission.VIEW_PARTICIPANTS,
+        Permission.VIEW_ADMIN_DRAW,
+        Permission.VIEW_DEBATES,
+        Permission.RELEASE_DRAW,
+        Permission.UNRELEASE_DRAW,
+        Permission.EDIT_STARTTIME,
```
tienne-B, line 117: What were the changes here to the previous comment?

PR #2391, opened 2024-01-25
Correct calculation of draw strength

tienne-B:
The draw strength annotators erroneously duplicated teams in the list of opponents when there were unconfirmed ballots between the teams. This commit adds the additional filter.

### DIFF
--- a/tabbycat/standings/teams.py
+++ b/tabbycat/standings/teams.py
@@ -189,7 +189,7 @@ class BaseDrawStrengthMetricAnnotator(BaseMetricAnnotator):
         opponents_annotation = ArrayAgg('debateteam__debate__debateteam__team_id',
                 filter=opponents_filter)
         logger.info("Opponents annotation: %s", str(opponents_annotation))
-        teams_with_opponents = queryset.all().annotate(opponent_ids=opponents_annotation)
+        teams_with_opponents = queryset.model.objects.annotate(opponent_ids=opponents_annotation)
         opponents_by_team = {team.id: team.opponent_ids for team in teams_with_opponents}
 
         opp_metric_queryset = self.opponent_annotator().get_annotated_queryset(


tabbycat/standings/teams.py
```diff
@@ -183,7 +183,7 @@ def annotate(self, queryset, standings, round=None):
 
         # Make a copy of teams queryset and annotate with opponents
         opponents_filter = ~Q(debateteam__debate__debateteam__team_id=F('id'))
-        opponents_filter &= Q(debateteam__debate__round__stage=Round.Stage.PRELIMINARY)
+        opponents_filter &= Q(debateteam__debate__round__stage=Round.Stage.PRELIMINARY, debateteam__teamscore__ballot_submission__confirmed=True)
```
czlee, line 186: I haven't looked at this for quite a long time, of course, but this surprises me a bit.

The original queryset is a `Team` queryset, right? But you're saying when a debate has (say) a confirmed and an unconfirmed ballot, it gets counted twice? But this calculation doesn't actually use the ballots, because `opponents_annotation` just returns an array of all the opponents? So why is it joining to the ballot submissions table in the first place?

tienne-B: All the joins are made because the query is repeated from the combinable metrics' query, but you're right, it should be a clean `Team` query.

Doing this, I think there could be a couple further optimizations:
- Use `FilteredRelation` to avoid filtering annotations on the same things
- If the draw strength metric is combinable, include it in the main query to avoid this extra call.

As this would be for a hotfix, I'll think more about the optimizations later.

tabbycat/standings/teams.py
```diff
@@ -189,7 +189,7 @@ def annotate(self, queryset, standings, round=None):
         opponents_annotation = ArrayAgg('debateteam__debate__debateteam__team_id',
                 filter=opponents_filter)
         logger.info("Opponents annotation: %s", str(opponents_annotation))
-        teams_with_opponents = queryset.all().annotate(opponent_ids=opponents_annotation)
+        teams_with_opponents = queryset.model.objects.annotate(opponent_ids=opponents_annotation)
```
czlee, line 192: Ah, that's a good idea. I guess the downside is that we would lose all the earlier filtering, if any, on `queryset`: This query would just return all teams in the database, right? Which is not too consequential, as subsequent lines only pull the relevant teams from it, but it's just suboptimal.

I was thinking ideally we'd have some sort of `distinct` in a subquery or something. But it wasn't obvious to me how to do this (and I haven't tried it myself). (I also thought about applying distinct to the annotation but of course teams can face each other twice so this wouldn't work.)

PR #2400, opened 2024-02-26
hdd-1

HDD-1998:


tienne-B:
Could you explain this change? Why should we have this new workflow?

Also, changes should be on `develop` rather than `master`.

tienne-B:
I haven't heard anything more in over a month, and am unlikely to merge so I'll close this. Feel free to re-open with an answer!

PR #2409, opened 2024-03-26
Add tournament published time field

idowolf:
This field will help knowing when the tab was published for archiving purposes. For example, I use Tabbycat's robust API to automatically add motions to our circuit's motion database, and it would help listing the correct date of the competition during the publish process.


tienne-B:
Sorry I haven't gotten to this until now... I am hesitant to add this feature, especially as a new field. I think there are a few reasons:
* This field is not used in the function of Tabbycat which isn't archival. It then becomes extraneous and a maintenance burden, especially with signals which aren't commonly used. Such information should be collected where it's used, possibly as a date selector on the archival application.
* As a slippery slope, would the existence of an published date mean that we should also store start/end dates or a schedule? These could be useful if we decide to implement a scheduling feature. In that case, merely having a "published date" seems incomplete, and would be better applicable in it's own model.
* Conceptually, I would think of publishing/retracting as an action, which is better represented as an action log entry. The tournament model itself doesn't need to know when something has occurred, and cannot reflect a series of publishing as it's just one value.

As for the code itself, avoid making extra migrations, and it should blank the published date if unreleased.

tienne-B:
Closing this as I think the path forward is with a scheduling feature rather than this graph.

PR #2413, opened 2024-04-02
Install nvm in Dockerfile

### LINKED ISSUE #2410: Tabbycat installation problem in Docker
After installing docker on my PC, when I installed Tabbycat many errors were coming(I don't remember those). But after installing node.js I thought it would be fixed but then this specific error is coming. 
tabby
tabby 2

How should I solve it?

tienne-B:
With the advice from. https://stackoverflow.com/a/57344191, the Dockerfile will install nvm in order to install node. This should fix the styling issue [reported on FB](https://www.facebook.com/groups/tabbycat.debate/posts/2621976821303592/).

Closes #2410

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@
 
 # Grab a python image
 FROM python:3.9
+SHELL ["/bin/bash", "--login", "-c"]
 
 # Just needed for all things python (note this is setting an env variable)
 ENV PYTHONUNBUFFERED 1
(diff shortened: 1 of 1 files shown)


gserranof:
Sorry for bringing this up. I don't have any specific problem with the commit, but I wanted to point out that it didn't fix the styling issue. Yesterday, I installed it locally with Docker using the develop branch (which already includes this commit, so the files were up-to-date), but I encountered the same issue. Just for context, I have very little knowledge of Python and Django—just a basic understanding—so even though I've managed to solve this, I'm not sure it's the proper way at all.

The issue that caused errors for me, as observed in the browser console, were several 404 errors pointing to some files, specifically style.css, djangojs.js, and app.js, from static/css, static/jsi18n, and static/vue, respectively. Since the Tabbycat repository doesn't have any of these files (the css and jsi18n folders are empty, and the vue folder doesn't exist), I tried to find them in a Tabbycat instance that I know works as intended: Calico. I grabbed the respective files from one of my sites, and after placing them in their respective folders for the local instance, I made a fresh start, and now everything works as intended. To be more specific, I saved the following files: vue/js/app.js, vue/js/app.js.map, css/style.css, and jsi18n/es/djangojs.js.

As I said, I'm not sure if what I did is the best way to solve this issue, nor am I certain if those files are public or if they are Étienne's property (as I found them at Calico but not at Tabbycat). However, I just wanted to try to resolve it as best as I could before pointing you to the issue, in case that could help.

gserranof:
I'm afraid that hasn't solved it either - I've deployed a new instance editing that line before, but the styling remains broken. Anyway, it's clear that there must be a way more simple solution than mine. I'm happy that this is known; I'll try to find some other way, but I know this is in good hands, so thank you anyway!

tienne-B:
@gserranof Try using #2420 ?

PR #2434, opened 2024-05-13
Add Speaker Score Criterion

### LINKED ISSUE #771: Individual scores should be deconstructible
There are formats where individual scores are broken down into criteria (for the Canadian National Debating Championships [high school]: Content, Style, and Strategy) _Personally, I dislike having these codified criteria, and think they are contradictory to their instruction to judge "holistically."_ These criteria can have different scoring ranges, but are summed to form the speaker's speech score.

Tabbycat should be able to include these criteria in ballots, possibly being akin to adjudicator feedback questions. This would make Tabbycat more flexible for other formats and be able to hold more granular data, which is often ignored or not included in the tab.

linh-trant:
This PR adds models in order to represent criteria for speaker scores and their values in ballots. There is a composite key between the criterion and speaker score, but an id is auto-generated. We use `SpeakerScore[ByAdj]` as a foreign key to avoid all the similar fields we'd need to associate them with.

Closes #771.


tabbycat/results/models.py
```diff
@@ -376,3 +377,61 @@ def clean(self):
...
+    speaker_score_by_adj = models.ForeignKey(SpeakerScoreByAdj, models.CASCADE,
+        verbose_name="speaker score")
+
+    class Meta:
+        unique_together = [('speaker_score_by_adj', 'criterion')]
+        verbose_name = _("speaker score for criterion by adjudicator")
+        verbose_name_plural = _("speaker scores for criteria by adjudicator")
+
+    def __str__(self):
+        return ("[{0.ballot_submission_id}/{0.id}] {0.score} at {0.position} for "
```
tienne-B, line 436: This one won't work as `ballot_submission_id`, `position`, `debate_team`, and `debate_adjudicator` aren't defined.

tabbycat/results/scoresheet.py
```diff
@@ -76,6 +77,12 @@ def set_speaker_rank(self, side, position, score):
     def get_speaker_rank(self, side: str, position: int) -> int:
         return self.speaker_ranks[side][position]
 
+    def set_criterion_score(self, side: str, position: int, criterion, score):
+        self.criteria_scores[side][position][criterion] = score
+
+    def get_criterion_score(self, side: str, position: int, criterion):
+        return self.criteria_scores[side][position][criterion]
+
     def get_total(self, side):
```
tienne-B, line 86: Here we could have the calculation for the criteria:
```py
if self.criteria_scores[side]:
    return sum([criteron.weight * val for criterion, val in self.criteria_scores[side].items()])
```

tabbycat/templates/js-bundles/enter_results.js
```diff
@@ -27,6 +27,14 @@ function refresh_totals(scoresheet) {
...
     $aff_margin = $('.aff_margin', $scoresheet);
     $neg_margin = $('.neg_margin', $scoresheet);
+    if ($criteria) {
+      for (int i = 0; i < aff_speakers.size(); i++){
+        aff_speakers.indexOf(i) = sum($(.'aff.score .criterion input', $scoresheet))
+      }
+      for (int i = 0; i < neg_speakers.size(); i++){
+        neg_speakers.indexOf(i) = sum($(.'neg.score .criterion input', $scoresheet))
+      }
+    }
```
tienne-B, line 37: Maybe something like:
```js
if ($('.criteria', $scoresheet)) {
  for (const side of ['aff', 'neg']) {
    for (const speaker of [...$(`.{side}.score`, $scoresheet)]) {
      speaker.querySelector('> input').value = sum($('.criterion input', speaker))
    }
  }
}
```
would work

tienne-B (commented):
Almost there!

tabbycat/results/result.py
```diff
@@ -106,6 +106,9 @@ def DebateResult(ballotsub, *args, **kwargs):  # noqa: N802 (factory function)
     r = kwargs.pop('round', ballotsub.debate.round)
     tournament = kwargs.pop('tournament', r.tournament)
     result_class = get_result_class(ballotsub, r, tournament)
+    if result_class.uses_speakers and 'criteria' not in kwargs:
+        from .models import ScoreCriterion
+        kwargs['criteria'] = ScoreCriterion.objects.filter(tournament=tournament)
```
czlee, line 111: Would 
```suggestion
        kwargs['criteria'] = tournament.scorecriterion_objects.all()
```
work?

(I'm guessing the keyword argument is just to allow a performance optimization, and would never be anything other than all the criteria?)

tabbycat/results/scoresheet.py
```diff
@@ -76,6 +79,17 @@ def set_speaker_rank(self, side, position, score):
     def get_speaker_rank(self, side: str, position: int) -> int:
         return self.speaker_ranks[side][position]
 
+    def set_criterion_score(self, side: str, position: int, criterion, score):
+        self.criteria_scores[side][position][criterion] = score
+        weighted_score = float(score) * criterion.weight
+        if self.scores[side][position] is None:
+            self.scores[side][position] = weighted_score
+        else:
+            self.scores[side][position] += weighted_score
```
czlee, line 88: Would this implementation mean that `set_criterion_score` can be set at most once per speaker per criterion, otherwise if (for whatever reason) this is called twice for the same criterion, that score will be double-counted? (In other words, this function is not idempotent?)

If so, I'd either:
1. Rewrite it so that it is idempotent, _i.e._ so that this method _only_ sets the criterion score, and the scores themselves are always computed dynamically. This probably would require spinning out criterion score handling into a new class that does not store `self.scores` or permit `set_score()`, _i.e._ values must only be set via criterion scores. On the one hand, this is probably a good idea(?). On the other hand, it's a lot of work.
2. Check that the criterion score isn't already set, and raise an error (_i.e._ fails fast) if it is not, so that bugs are highlighted quickly rather than becoming time bombs.

czlee (commented):
I haven't gone through the whole thing in details, but I've looked at most files in the results directory and parts of it in detail, and the general approach looks good to me. I had a closer look at results/models.py and I'm happy with the model schema.

Issue #2435, opened 2024-05-18
Automatically set start time on motion release


tienne-B:
We should be able to have a setting to change the amount of prep time, and if set, when the motion is first released.

It should default to being deactivated but 15 minutes for BP.

PR #2437, opened 2024-05-21
Auto prep time on motion release

TrudeauOkech:
### **Description**
This commit adds a setting to modify prep time in Configuraton>Debate Rules. It also calculates round start time, in minutes, and displays it on triggering the release motions action. The default is set to 0 minutes and users can still modify the start time manually as before. The BP preset is set to 15 minutes.

### **Screenshots**
Screenshot from 2024-05-21 23-50-52

### **Additional Context**
This PR is related to issue #2435

### DIFF
--- a/tabbycat/motions/views.py
+++ b/tabbycat/motions/views.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.contrib import messages
 from django.db.models import OuterRef, Prefetch, Q, Subquery
+from django.utils import timezone
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy, ngettext
 from django.views.generic.base import TemplateView
@@ -177,6 +178,15 @@ class ReleaseMotionsView(BaseReleaseMotionsView):
     @property
     def message_text(self):
         return ngettext("Released the motion.", "Released the motions.", self.round.motion_set.count())
+
+    def post(self, request, *args, **kwargs):
+        preparation_time = self.tournament.pref('preparation_time')
+
+        if preparation_time > -1:
+            self.round.starts_at = timezone.now() + timezone.timedelta(minutes=preparation_time)
+            self.round.save()
+
+        return super().post(request, *args, **kwargs)
 
 
 class UnreleaseMotionsView(BaseReleaseMotionsView):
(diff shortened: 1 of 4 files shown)


tabbycat/options/preferences.py
```diff
@@ -614,7 +614,13 @@ class UseSpeakerRanks(ChoicePreference):
...
     )
     default = 'none'
-
+@tournament_preferences_registry.register
+class PreparationTime(IntegerPreference):
+    help_text = _("How long after motion release does the round start(In Minutes)")
+    verbose_name = _("Preparation Time")
+    section = debate_rules
+    name = 'preparation_time'
+    default = 0
```
tienne-B, line 623:
```suggestion
    help_text = _("How long, in minutes, after motion release does the round start (-1 to deactivate)")
    verbose_name = _("Preparation Time")
    section = debate_rules
    name = 'preparation_time'
    default = -1
```

We use the convention of "-1" being an "off switch" for some other preferences; we should use the same convention. "0" could indicate an immediate start.

tabbycat/motions/views.py
```diff
@@ -170,13 +171,31 @@ def post(self, request, *args, **kwargs):
...
+        start_minutes = preparation_time_preference  
+
+        
+        start_time = timezone.now() + timezone.timedelta(minutes=start_minutes)
+        start_time_local = start_time.astimezone(timezone.get_current_timezone())
+        self.round.starts_at = start_time_local
+        self.round.save()
+
+
+        messages.success(request, f"Debate will start at {start_time_local.strftime('%H:%M')} ({settings.TIME_ZONE})")
```
tienne-B, line 197:
```suggestion
        preparation_time = self.tournament.pref('preparation_time')
        
        if preparation_time > -1:
            self.round.starts_at = timezone.now() + timezone.timedelta(minutes=start_minutes)
            self.round.save()

            messages.success(request, f"Debates will start at {self.round.start_time.strftime('%H:%M')} ({settings.TIME_ZONE})")
```

Does this look good/work well? We have a `.pref()` shortcut, and as we're using `timezone.now()` it should already be in the proper TZ.

tienne-B (commented):
Looks good! Should use a bit of clean-up and then it'd be good to go.

As a general note, avoid including the unrelated migrations in this PR, and run linting.

tabbycat/motions/views.py
TrudeauOkech: I've fixed it. Please check that it's all OK now.

PR #2442, opened 2024-06-16
Install black as a pre-commit hook

tienne-B:
This is a huge PR, as it's re-formatting a lot of the code. However only the 1st commit implements `black`. An automatic code formatter is more user-friendly than a linter in that most problems get automatically fixed, so there'd be less friction as to "why can't I commit."

Most of the reformatting is to prefer quotation marks over apostrophes for strings, and expanding lines. These commits can be excluded when bisecting/blaming past this PR, as explained in [Black's docs](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame).

I would want to move towards [Ruff](https://docs.astral.sh/ruff/) to replace both Black and Flake8 if it could have the fixes for long lines.


pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
+line-length = 100
```
czlee, line 2: Can we use 120? I feel that 100 is unduly conservative.

tabbycat/breakqual/tests/test_liveness.py
```diff
@@ -61,60 +61,92 @@ def test_case_australs_2015_open(self):
...
+                17205,
+                9240,
+                4455,
+                1902,
+                705,
+                220,
+                55,
+                10,
+                1,
+            ],  # noqa: E501
```
czlee, line 149: Changes like this seem like clearly a bad idea to me. This had a `# noqa` ignore command for flake8, how do we do the same with black?

tabbycat/breakqual/tests/test_liveness.py
```diff
@@ -150,12 +182,1038 @@ def test_wudc_2018_open(self):
...
-            5: (16, -1, [6, 8, 7, 7, 8, 5, 9, 8, 5, 7, 9, 6, 5, 8, 6, 6, 6, 5, 6, 5, 6, 7, 4, 7, 7, 8, 6, 5, 8, 6, 6, 5, 3, 7, 4, 6, 4, 5, 4, 8, 6, 5, 8, 5, 6, 6, 4, 5, 7, 7, 7, 5, 6, 6, 4, 5, 5, 6, 4, 6, 3, 5, 6, 2, 6, 8, 5, 5, 6, 7, 5, 8, 6, 8, 7, 4, 6, 4, 5, 5, 4, 5, 3, 5, 4, 5, 5, 3, 5, 3, 2, 4, 7, 6, 6, 7, 6, 6, 4, 5, 3, 5, 5, 4, 4, 2, 3, 7, 4, 4, 2, 3, 0, 4, 3, 4, 5, 6, 4, 6, 6, 2, 3, 3, 3, 3, 4, 4, 5, 3, 2, 5, 5, 5, 4, 3, 6, 4, 4, 5, 5, 4, 6, 3, 2, 3, 1, 3, 2, 2, 2, 2, 3, 3, 4, 1, 1, 2, 3, 1, 3, 2, 4, 1, 0, 0]),  # noqa: E501
-            4: (16, -4, [6, 5, 6, 5, 6, 5, 8, 7, 5, 6, 7, 3, 3, 5, 4, 6, 5, 2, 4, 5, 6, 5, 2, 4, 7, 6, 6, 3, 5, 4, 6, 3, 3, 4, 3, 5, 3, 5, 3, 5, 5, 4, 5, 4, 5, 5, 4, 5, 5, 5, 4, 3, 5, 6, 4, 5, 4, 3, 3, 5, 3, 5, 4, 2, 6, 5, 5, 3, 5, 5, 5, 6, 4, 5, 4, 4, 4, 2, 2, 5, 4, 3, 3, 4, 3, 3, 4, 3, 5, 1, 2, 3, 6, 3, 6, 5, 3, 4, 2, 4, 3, 4, 3, 2, 3, 1, 3, 4, 2, 2, 2, 3, 0, 4, 3, 3, 4, 4, 3, 5, 5, 2, 2, 2, 3, 3, 3, 1, 2, 2, 1, 3, 2, 3, 3, 3, 3, 4, 4, 2, 4, 1, 6, 1, 1, 0, 1, 2, 2, 0, 2, 2, 3, 3, 2, 1, 1, 2, 3, 1, 1, 1, 1, 0, 0, 0]),  # noqa: E501
+            9: (
+                16,
+                11,
+                [
```
czlee, line 188: Or this, even worse 🙈

tabbycat/api/permissions.py
```diff
@@ -40,24 +42,26 @@ def get_required_permission(self, view):
...
+                "destroy": getattr(view, "destroy_permission", False),
+                "delete_all": getattr(view, "destroy_permission", False),
+                "add_blank": getattr(view, "create_permission", False),
+                "GET": getattr(view, "list_permission", False),
+                "POST": getattr(view, "update_permission", False),
+                "PUT": getattr(view, "update_permission", False),
+                "PATCH": getattr(view, "update_permission", False),
+                "DELETE": getattr(view, "destroy_permission", False),
+            }
+        ).get(getattr(view, "action", view.request.method), False)
```
czlee, line 61: Is there a way to configure it to permit two opening brackets on the same line, where the same brackets are adjacent on both sides?

czlee:
> Most of the reformatting is to prefer quotation marks over apostrophes for strings, and expanding lines.

Previously we were using double quotation marks for human-readable strings, and single quotation marks for machine-readable strings. (Of course, a linter cannot enforce this, so just has to disable the check.) I have a slight preference for this but if both of you feel strongly about switching to double quotation marks that's fine.

Expanding lines concerns me though. I've highlighted a particularly egregious example above, but I find oftentimes exploding nested brackets into many lines makes code less readable, not more readable. (Not always—in many of the changes here so far, there's not much difference.) There's probably some way to configure it to be more reasonable, though.

teymour-aldridge:
> Expanding lines concerns me though. I've highlighted a particularly egregious example above, but I find oftentimes exploding nested brackets into many lines makes code less readable, not more readable. (Not always—in many of the changes here so far, there's not much difference.) There's probably some way to configure it to be more reasonable, though.

I did a quick web search and I believe Black doesn't support this (https://github.com/psf/black/issues/1281). I agree that it's not an improvement to have these long arrays wrapped on multiple lines; I think the easiest way to use this is to use the `#fmt: on` and `#fmt: off` directives.

PR #2444, opened 2024-06-18
Create Event object for schedules

### LINKED ISSUE #2425: Create schedule Event model
To begin for #2424, we'd need to create a Django model for each "event" within a tournament. For now, they will only be accessible/modifiable within Edit DB. We should be able to optionally link events to specific rounds, and either have a predefined event type or custom text.

Some ideas for predefined events:
* Check-in
* Briefings
* Debates occurring
* Break announcement
* Awards

In `tournaments/models.py`, we can create a model with the following fields:
* _id_
* tournament (fk)
* round (fk, nullable)
* type (text choice, blank)
* title (text, blank)
* start (aware date-time)
* end (aware date-time, nullable)

You'll then have to add the model to `tournaments/admin.py` for it to show up in Edit DB.

- [ ] Create model
- [ ] Create migration (`dj makemigrations tournament`)
- [ ] Add to Edit DB

daankoning:
Closes #2425


tabbycat/tournaments/models.py
```diff
@@ -584,3 +584,38 @@ def motions_good_for_public(self):
...
+
+    class Types(models.TextChoices):
+        CHECK_IN = 'C', _("Check-in")
+        DRAW = 'T', _("Draw/motion announcement")
+        PREP = 'P', _("Preparation time")
+        DEBATE = 'D', _("Debate")
+        ADJUDICATION = 'J', _("Adjudication")
+        BREAK = 'B', _("Break announcement")
+        AWARDS = 'A', _("Award ceremony")
+        OTHER = 'O', _("Other")
```
tienne-B, line 599: Thoughts about adding a couple more types:
* Briefings
* Meals / break (as in rest)

daankoning: Fair enough :)

PR #2459, opened 2024-07-04
Making the ballot view responsive/phone-friendly

linh-trant:
Not all the judges have laptops available all the time for ballots; thus, they submit their ballots through not-so-responsive ballot view on their phones. Given the score criteria release, the layout got messier. This PR is for fixing that problem. 

Here is the view of ballots:
https://drive.google.com/drive/folders/1e0ZhBXCnEYHp7yD6DD9cMDXIHkOcrsSc?usp=sharing

### DIFF
--- a/tabbycat/results/templates/ballot/ballot_scoresheet.html
+++ b/tabbycat/results/templates/ballot/ballot_scoresheet.html
@@ -22,7 +22,7 @@
     <div class="list-group-item">
       <div class="row">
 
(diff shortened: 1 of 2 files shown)


tienne-B (commented):
In "With criterion normal", could we have some alignment between the speaker and the first criterion? Or no?

Screenshot 2024-07-04 at 18 07 37

PR #2462, opened 2024-07-06
Updates coach class in participants/models.py

### LINKED ISSUE #2460: Create new "Coach" person model for institutions
We should create a Person model to associate the people responsible for representing an institution. This would allow them to get an access through URL keys to create the participants for the institution.

To do this task, subclass `Person` in `participants/models.py`. This model only needs one extra field: `institution`. Then, create a migration with `dj makemigrations participants` and register the model in `participants/admin.py`.

abrarshahid:
The Coach class was updated and also added to admin.py

Closes #2460

### DIFF
--- a/tabbycat/participants/admin.py
+++ b/tabbycat/participants/admin.py
@@ -39,6 +39,12 @@ class InstitutionAdmin(ModelAdmin):
     ordering = ('name', )
     search_fields = ('name', )
 
+#Coach
+@admin.register(Coach)
+class CoachAdmin(ModelAdmin):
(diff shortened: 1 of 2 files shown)


tienne-B:
Hey! The main thing missing here is that either we add another foreign key field `tournament`, or implement #2467 and use have a foreign key to that table.

Having that tournament link is important as the goal of having a Coach class is being able to manage the participants of that institution for a tournament.

tienne-B:
This has been done in #2504 (taking your commits including attribution to you)

PR #2470, opened 2024-07-09
Fix qr codes generator

maclask:
Adapt to a new way python-qrcode returns path


tienne-B:
Thanks! Printing often escapes my view.

PR #2482, opened 2024-08-13
Removing mutable default args

elayeek:
This is a common pitfall in Python that causes unexpected behavior.

Python evaluates functions on module load and generates mutable default argument objects once, then persist them whenever the function is called.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments


.gitignore
```diff
@@ -13,6 +13,10 @@ tabbycat/static/jsi18n/
 tabbycat/static/vue/
 tabbycat/static/fonts/
 
+# IDE files
+.idea/
+.vscode/
+
```
tienne-B, line 19: I think these could more efficiently be in your global gitignore, rather than being project-specific.

tienne-B (approved):
I'll wait for a linting commit (and for the CI to run after that) to merge this in.

.gitignore
elayeek: True, will fix that

Issue #2474, opened 2024-07-16
Create reserve team slots for breaks


tienne-B:
A new practice is to announce a couple teams that have not broken, but are asked to show up as reserve teams if a higher-breaking team does not show up. We should support adding reserves in the break.

tienne-B:
I don't think we'd even need a new DB field (but if you choose to go with a new `reserve` field, the rank numbering should not restart). Could add a preference for number of reserve teams, and then in the break generator create extra `BreakingTeam` objects for the next _n_ teams. Then, as the break category size is smaller, the rounds should be generated with just that number of teams. Could even have a break remark called "Reserve."

One thing to be cautious would be to make sure the reserve teams remain eligible for other breaks, if they'd have qualified.

PR #2498, opened 2024-09-11
Add reserve teams to break.

teymour-aldridge:
Should fix https://github.com/TabbycatDebate/tabbycat/issues/2474.

I'm not sure where the tests (if any) for the debate break generator are – I've just tested this manually using the BP 88 team sample setup. If you could point me to where they are, will add some for this.


tabbycat/breakqual/models.py
```diff
@@ -20,8 +20,10 @@ class BreakCategory(models.Model):
         verbose_name=_("sequence number"),
         help_text=_("The order in which the categories are displayed"))
     break_size = models.IntegerField(validators=[MinValueValidator(2)],
-        verbose_name=_("break size"),
-        help_text=_("Number of breaking teams in this category"))
+                                     verbose_name=_("break size"),
+                                     help_text=_("Number of breaking teams in this category"))
+    reserve_size = models.IntegerField(default=0, validators=[MinValueValidator(0)], verbose_name=_("Reserve size"),
```
tienne-B, line 25:
```suggestion
    reserve_size = models.PositiveIntegerField(default=0, verbose_name=_("Reserve size"),
```

tabbycat/breakqual/models.py
```diff
@@ -91,17 +93,19 @@ class BreakingTeam(models.Model):
...
     REMARK_WITHDRAWN = 'w'
+    REMARK_RESERVE = 'R'
     REMARK_CHOICES = (
         (REMARK_CAPPED, _("Capped")),
         (REMARK_INELIGIBLE, _("Ineligible")),
         (REMARK_DIFFERENT_BREAK, _("Different break")),
         (REMARK_DISQUALIFIED, _("Disqualified")),
         (REMARK_LOST_COIN_TOSS, _("Lost coin toss")),
         (REMARK_WITHDRAWN, _("Withdrawn")),
+        (REMARK_RESERVE, _("Reserve")),
```
tienne-B, line 104: This is a related ask, but could we convert these into a `models.TextChoice` enum? I've been doing that when adding new items to an old-style choices list.

Could also be good to add an "Absent" remark.

tabbycat/breakqual/base.py
```diff
@@ -255,9 +265,18 @@ def compute_break(self):
...
         if len(self.eligible_teams) >= self.break_size:
             last_rank = self.eligible_teams[self.break_size-1].get_ranking("rank")
             for tsi in self.eligible_teams[self.break_size:]:
                 if tsi.get_ranking("rank") != last_rank:
                     break
                 self.breaking_teams.append(tsi)
+                tied_teams_count += 1
+
+        # Then add the reserve teams (these are later marked as "reserve" in the
+        # populate database section)
```
tienne-B, line 278: The logic here should be to include reserve teams up to `len(breaking_teams) == break_size + reserve_size`. Losers in a tie would be first to be a reserve, so adding more wouldn't be necessary.

tabbycat/breakqual/base.py
```diff
@@ -215,8 +216,17 @@ def populate_database(self):
...
-                    break_category=self.category, team=tsi.team,
-                    defaults={'rank': rank, 'break_rank': break_rank, 'remark': None},
+                    break_category=self.category,
+                    team=tsi.team,
+                    defaults={
+                        "rank": rank,
+                        "break_rank": break_rank,
+                        "remark": (
+                            BreakingTeam.REMARK_RESERVE
+                            if self.break_size < rank <= self.reserve_size + self.break_size
```
tienne-B, line 226: `x < y < z` is neat! I wonder though if we should do something akin to `exclude_capped_teams()` in `aida.py` for reserve teams rather than have this logic here.

tienne-B:
Hey @teymour-aldridge ! I've added a commit to clean a few things up using the `exclude_reserve_teams()` method. I think this method would be the best as to allow more subclassing (like with AIDA) and separate the concerns of the `compute_break` method.

For AIDA, breaks are capped to a certain number per institution, and this could cause some problems in deciding reserve teams, as a capped reserve may become eligible if the absent team is from their institution. By marking a team as absent though, the break is updated, and could include them, so it shouldn't be a problem.

Issue #2497, opened 2024-09-08
Anonymized speakers are revealed in the public record page


inohan:
When the public record page of teams with anonymized/redacted speakers are accessed, the speaker name is revealed in the `About (Team Name)` section instead of showing up as *Redacted*.
In the image below, Speaker 1A is set to an anonymous/redacted speaker, but the name is revealed.
<img width="1261" alt="image" src="https://github.com/user-attachments/assets/a98b9b3b-d776-477c-a475-11ec0d5f9c37">

This also applies for when ballots are released to public.

tienne-B:
Closed as a duplicate of #2249 .

PR #2499, opened 2024-09-11
Redact names in team results page.

teymour-aldridge:
Should fix https://github.com/TabbycatDebate/tabbycat/issues/2497.

I'm not sure if we want to redact names in the admin section of the site (as admins can view the names in the database anyway) but I guess it doesn't hurt.

~I imagine it would also be sensible to redact speakers from the API.~ (Apologies, I checked against the API of a Tabbycat instance but forgot I was logged in as an admin so could see the speaker details).

### DIFF
--- a/tabbycat/participants/models.py
+++ b/tabbycat/participants/models.py
@@ -134,6 +134,8 @@ class Person(models.Model):
         return str(self.name)
 
     def get_public_name(self, tournament):
+        if self.anonymous:
+            return "Redacted"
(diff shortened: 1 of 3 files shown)


tienne-B:
Hey! If it's fine by you, I'd take a second look at #2248 and remember what was the problem with that one, as they're basically the same solution.

tienne-B:
I've merged in #2248 so this one wouldn't be needed.

PR #2501, opened 2024-09-17
Fix error when running `simulaterounds` command.

teymour-aldridge:
Cast the result to a string before calling `logger.info`.


tienne-B (commented):
Could this error occur on lines 153 or 160?

PR #2504, opened 2024-09-27
Add coach model

### LINKED ISSUE #2460: Create new "Coach" person model for institutions

teymour-aldridge:
Supercedes #2462 and fixes #2460


tienne-B:
Hey! Thanks for this; I'll have a look this weekend hopefully.

PR #2512, opened 2024-10-17
Add missing csrf token to admin area.

teymour-aldridge:


tienne-B:
Thanks! We have to start enforcing CSRF (was getting those kinds of errors when changing the logouts to a POST request)

PR #2515, opened 2024-10-24
Add email validator to "Reply-To address" setting

### LINKED ISSUE #2514: Add email validator to "Reply-To address" setting
In here: https://github.com/TabbycatDebate/tabbycat/blob/cb8a08fbcdefffe2ba396fb42800c58d552d39c5/tabbycat/options/preferences.py#L1268-L1274

We should either add an attribute
```py
    field_class = forms.EmailField
```
(from django form fields) to get Django's email validation, or the validator directly:
```py
    field_kwargs = {'validators': [EmailValidator()]}
```

aamfahim:
Imported `EmailValidator` from django validators and added it to "Reply-To address" setting

Fixes #2514

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -1,6 +1,6 @@
 from decimal import Decimal
 
-from django.core.validators import MinValueValidator, validate_slug, EmailValidator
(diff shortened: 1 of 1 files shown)


tienne-B:
Perfect! I'll merge this in when it passes the build. Thank you!

PR #2520, opened 2024-10-30
Show side abbrevations on the motion statistics page again.

teymour-aldridge:
Tabbycat used to show side names, but at some point this broke. This commit fixes this.

### DIFF
--- a/tabbycat/motions/statistics.py
+++ b/tabbycat/motions/statistics.py
@@ -6,6 +6,7 @@ from django.utils.translation import gettext_lazy
 
 from draw.types import DebateSide
 from tournaments.models import Round
+from tournaments.utils import get_side_name
 
 from .models import Motion, RoundMotion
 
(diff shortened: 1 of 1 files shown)


tabbycat/motions/statistics.py
```diff
@@ -236,16 +244,17 @@ def _collate_prelim_motion_annotations(self):
             motion.counts_by_bench = {'gov': 0, 'opp': 0}
 
             for side in self.tournament.sides:
+                side_abbrv = abbrv_of_bp_side(side)
```
tienne-B, line 247: Should use `tournaments.utils.get_side_name` like:

```py
side_abbrev = get_side_name(side, 'abbr')
```

Then we wouldn't need the new function.

teymour-aldridge: Thanks! I've updated it to use that function.

I don't believe it affects two-team formats (e.g. here this problem doesn't seem to exist https://edudrift.calicotab.com/australs2024/motions/statistics/).

PR #2542, opened 2024-11-29
Allow zero values for integer feedback scales.

### LINKED ISSUE #2541: Adjudicator Feedback Integer Scale is buggy
image

Doesn't let me add an integer scale despite having a maximum and a minimum.

teymour-aldridge:
Fixes #2541

### DIFF
--- a/tabbycat/adjfeedback/admin.py
+++ b/tabbycat/adjfeedback/admin.py
@@ -40,8 +40,10 @@ class QuestionForm(forms.ModelForm):
     def clean(self):
         integer_scale = AdjudicatorFeedbackQuestion.ANSWER_TYPE_INTEGER_SCALE
         if self.cleaned_data.get('answer_type') == integer_scale:
-            if not self.cleaned_data.get('min_value') or not self.cleaned_data.get('max_value'):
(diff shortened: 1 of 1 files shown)


tabbycat/adjfeedback/admin.py
```diff
@@ -40,8 +40,10 @@ class Meta:
     def clean(self):
         integer_scale = AdjudicatorFeedbackQuestion.ANSWER_TYPE_INTEGER_SCALE
         if self.cleaned_data.get('answer_type') == integer_scale:
-            if not self.cleaned_data.get('min_value') or not self.cleaned_data.get('max_value'):
-                raise forms.ValidationError(_("Integer scales must have a minimum and maximum"))
+            if self.cleaned_data.get('min_value') is None:
+                raise forms.ValidationError(_("Error: min_value must be specified for an integer scale."))
```
tienne-B, line 44: If you want to split the messages for the field, you could use [`Form.add_error()`](https://docs.djangoproject.com/en/5.1/ref/forms/api/#django.forms.Form.add_error) so that the error shows for the specific field (and could simplify the message so that specifying the field isn't needed, like "A value is required when using an integer scale")

PR #2556, opened 2024-12-08
Show only breaking participants in check-in screen

DDH13:


DDH13:
I'm having a bit of a trouble reducing the cyclomatic complexity of the get_context_data method

tienne-B (commented):
Hey! Thanks for this PR! Don't need to worry about Codeclimate BTW.

I was expecting more that the filter is added as an option on the top bar of the page, so it would be more of a frontend change than backend. Having it as a frontend filter would be more flexible, and could include adjudicators as well.

Screenshot 2024-12-08 at 11 07 54

However, that would also require a backend change to pass the teams' break(s) to the frontend. Does this sound like a good plan?

DDH13:
> Hey! Thanks for this PR! Don't need to worry about Codeclimate BTW.
> 
> I was expecting more that the filter is added as an option on the top bar of the page, so it would be more of a frontend change than backend. Having it as a frontend filter would be more flexible, and could include adjudicators as well.
> 
> Screenshot 2024-12-08 at 11 07 54
> However, that would also require a backend change to pass the teams' break(s) to the frontend. Does this sound like a good plan?

Yep, will try to do what you said.

tabbycat/checkins/views.py
tienne-B, line 1: We'd have to set `breaking_team_short_names` for `CheckInVenuesStatusView` as well

tabbycat/checkins/templates/CheckInStatusContainer.vue
```diff
@@ -233,6 +254,28 @@ export default {
...
+      }
+      if (this.filterByBreaking.All) {
+        return this.entitiesByPresence // No filter
+      } else if (this.filterByBreaking.Breaking) {
+        console.log('Breaking Teams and Adjudicators')
+        // print each object in the array
+        const temp = this.entitiesByPresence.filter(isBreaking)
+        for (let i = 0; i < temp.length; i++) {
+          console.log(temp[i])
+        }
```
tienne-B, line 274: Avoid using `console`, as it's just for debugging.

tabbycat/checkins/templates/CheckInStatusContainer.vue
```diff
@@ -233,6 +254,28 @@ export default {
       }
       return this.entitiesByTime
     },
+    entitiesByBreaking: function () {
+      const isBreaking = (entityByPresence) => {
+        if (entityByPresence.type === 'Adjudicator') {
+          return entityByPresence.breaking
+        } else if (this.breakingTeamShortNames.includes(entityByPresence.name)) {
```
tienne-B, line 261: Could using team IDs be shorter and avoid needing to escape the strings?

tienne-B (commented):
It's advancing well! Little problem is that I'm getting these errors when trying to filter all (both adj and teams) by breaking:

Screenshot 2024-12-21 at 22 49 55

tabbycat/checkins/templates/CheckInStatusContainer.vue
DDH13: Yep, it's just temporary, will remove

tabbycat/checkins/templates/CheckInStatusContainer.vue
DDH13: Will do

tienne-B:
Hey @DDH13 ! Any update on this?

DDH13:
Hi @tienne-B , sorry was trying to learn Vue.js and got a bit sidetracked with some work stuff. I'll give it a shot next week. The backend changes should be easy enough. Is there a release date that you are targeting?

PR #2572, opened 2025-01-02
Add team metrics to speakers

### LINKED ISSUE #2553: Add team ranking metrics for speakers
Some tournaments really try to avoid ties in the speaker standings, and for that, they often use team metrics, such as "team points", "total speaks", "total firsts", etc. We already have a "team points" metric for speakers, but could make metrics in https://github.com/TabbycatDebate/tabbycat/blob/develop/tabbycat/standings/speakers.py for the others.

TrudeauOkech:
Fixes #2553 

Added the following team metrics to break speaker ties.
Team wins
Number of 1sts
Number of seconds
Number of 3rds
Votes/ballots

### DIFF
--- a/tabbycat/standings/speakers.py
+++ b/tabbycat/standings/speakers.py
@@ -3,6 +3,7 @@
 import logging
 
 from django.db.models import Avg, Case, Count, F, FloatField, Max, Min, Q, StdDev, Sum, When
+from django.db.models.functions import Cast, NullIf
 from django.utils.translation import gettext_lazy as _
 
 from tournaments.models import Round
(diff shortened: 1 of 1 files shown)


tabbycat/standings/speakers.py
```diff
@@ -84,6 +85,123 @@ def get_annotation(self, round):
...
+        """Returns a QuerySet annotated with the number of wins for the team of the speaker."""
+        annotation_filter = Q(
+            team__debateteam__teamscore__ballot_submission__confirmed=True,
+            team__debateteam__debate__round__stage=Round.Stage.PRELIMINARY,
+            team__debateteam__teamscore__win=self.where_value,
+        )
+        if round is not None:
+            annotation_filter &= Q(team__debateteam__debate__round__seq__lte=round.seq)
+
+        return self.function('team__debateteam__teamscore__win', filter=annotation_filter)
```
tienne-B, line 109: We should be able to create a superclass for these that abstracts these out, like what's done for `teams.py`.

tienne-B: Hey! I've pushed up a change to use a dedicated subclass rather than a big `if`. Thoughts, @TrudeauOkech ?

TrudeauOkech: Yup, this works too!

PR #2580, opened 2025-01-24
Add draw strength by rank pullup resolution option to preferences.

teymour-aldridge:
Not sure if a database migration is required here, but when I ran `dj makemigrations` Django created some migrations related to institutional registration, and non related to preferences.


tienne-B:
Has this been superseded by #2581 ?

teymour-aldridge:
Yes I think so

PR #2585, opened 2025-01-28
Add ability to re-send account creation emails.

### LINKED ISSUE #2582: Add ability to re-send account creation invitation emails
One day is often not enough!

teymour-aldridge:
Submitting the "invite user" form a second time will now send a new email.

Fixes #2582

### DIFF
--- a/tabbycat/users/forms.py
+++ b/tabbycat/users/forms.py
@@ -34,7 +34,8 @@ class InviteUserForm(PasswordResetForm):
                 'username': email.split("@")[0],
             },
         )
-        user.membership_set.create(group=self.cleaned_data['role'])
+        if created is True:
+            user.membership_set.create(group=self.cleaned_data['role'])
         return [user]
 
     def save(self, *args, **kwargs):
--- a/tabbycat/users/forms.py
+++ b/tabbycat/users/forms.py
@@ -34,7 +34,7 @@ class InviteUserForm(PasswordResetForm):
                 'username': email.split("@")[0],
             },
         )
-        if created is True:
+        if created:
             user.membership_set.create(group=self.cleaned_data['role'])
         return [user]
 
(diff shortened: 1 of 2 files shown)


tabbycat/users/forms.py
```diff
@@ -34,7 +34,8 @@ def get_users(self, email):
                 'username': email.split("@")[0],
             },
         )
-        user.membership_set.create(group=self.cleaned_data['role'])
+        if created is True:
```
tienne-B, line 37:
```suggestion
        if created:
```
Simpler :)

teymour-aldridge:
For

         if created:
            user.membership_set.create(group=self.cleaned_data['role'])
    
This might cause some confusion if a user is invited once with some role, but the follow-up email is sent with a different role (as the second role will be ignored). However, I think this is more reasonable than changing the users role here (this should be done separately if necessary)?

tienne-B:
@teymour-aldridge That is a good point! A user can have many groups, and different groups between tournaments, so that new user check is problematic.

Have you tried using the [`.add()`](https://docs.djangoproject.com/en/5.1/ref/models/relations/#django.db.models.fields.related.RelatedManager.add) method instead of `.create()`?

I think it'd look something like this:

```py
user.membership_set.add(self.cleaned_data['role'])
```

tabbycat/users/forms.py
```diff
@@ -34,7 +34,8 @@ def get_users(self, email):
                 'username': email.split("@")[0],
             },
         )
-        user.membership_set.create(group=self.cleaned_data['role'])
+        if created:
```
tienne-B, line 37: With the `.add()` method call, do we need this check? It could cause problems if re-inviting a user, but with a new group (like CA + Equity or in another tournament)

PR #2595, opened 2025-02-10
WSDC graph pairing algorithm

teymour-aldridge:
Work in progress

### DIFF
--- a/tabbycat/draw/generator/graph.py
+++ b/tabbycat/draw/generator/graph.py
@@ -8,6 +8,7 @@ from ..types import DebateSide
 
 if TYPE_CHECKING:
     from participants.models import Team
+    from typing import Optional
 
 
 def sign(n: int) -> int:
@@ -23,7 +24,7 @@ class GraphGeneratorMixin:
         """Graph optimisation avoids conflicts, so method is extraneous."""
         pass
 
-    def assignment_cost(self, t1, t2, size, bracket=None):
+    def assignment_cost(self, t1, t2, size, bracket=None) -> Optional[int]:
         if t1 is t2:  # Same team
             return
 
@@ -37,6 +38,10 @@ class GraphGeneratorMixin:
         if self.options["side_allocations"] == "balance" and self.options["side_penalty"] > 0:
             t1_affs, t1_negs = t1.side_history
             t2_affs, t2_negs = t2.side_history
+
+            if self.options["max_times_on_one_side"] > 0:
+                if max(t1_affs, t1_negs, t2_affs, t1_negs) > self.options["max_times_on_one_side"]:
+                    return None
 
             # Only declare an imbalance if both sides have been on the same side more often
             # Affs are positive, negs are negative. If teams have opposite signs, negative imbalance
--- a/tabbycat/draw/generator/graph.py
+++ b/tabbycat/draw/generator/graph.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
 
 import munkres
 import networkx as nx
@@ -8,7 +8,6 @@ from ..types import DebateSide
 
 if TYPE_CHECKING:
     from participants.models import Team
-    from typing import Optional
 
 
 def sign(n: int) -> int:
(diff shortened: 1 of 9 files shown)


tabbycat/draw/generator/powerpair.py
```diff
@@ -285,15 +285,15 @@ def get_n_teams(self, teams: list['Team']) -> int:
...
+    def assignment_cost(self, t1, t2, size, bracket=None) -> Optional[int]:
         penalty = super().assignment_cost(t1, t2, size)
         if penalty is None:
             return None
 
         # Add penalty for seeing the pullup again
         has_pullup = 'pullup' in self.team_flags.get(t1, []) or 'pullup' in self.team_flags.get(t2, [])
         if self.options["pullup_debates_penalty"] and has_pullup:
-            penalty += max(t1.pullup_debates, t2.pullup_debates) * self.options["pullup_debates_penalty"]
+            penalty += (t1.pullup_debates + t2.pullup_debates) * self.options["pullup_debates_penalty"]
```
tienne-B, line 296: This isn't exactly what you're looking for; this isn't a pullup penalty _per se_, but so that a team doesn't get put against the pullup team too often (an APDA rule; don't think there's an equivalent in WSDC)

tabbycat/draw/generator/graph.py
```diff
@@ -38,6 +39,13 @@ def assignment_cost(self, t1, t2, size, bracket=None):
             t1_affs, t1_negs = t1.side_history
             t2_affs, t2_negs = t2.side_history
 
+            if self.options["max_times_on_one_side"] > 0:
+                if t1_affs > self.options["max_times_on_one_side"]\
+                   or t1_negs > self.options["max_times_on_one_side"]\
+                   or t2_affs > self.options["max_times_on_one_side"]\
+                   or t2_negs > self.options["max_times_on_one_side"]:
```
tienne-B, line 46:
```suggestion
                if max(t1_affs, t1_negs, t2_affs, t1_negs) > self.options["max_times_on_one_side"]:
```

tabbycat/draw/generator/graph.py
```diff
@@ -23,7 +24,7 @@ def avoid_conflicts(self, pairings):
         """Graph optimisation avoids conflicts, so method is extraneous."""
         pass
 
-    def assignment_cost(self, t1, t2, size, bracket=None):
+    def assignment_cost(self, t1, t2, size, bracket=None) -> Optional[int]:
```
kang-de-conqueror, line 27: You can use `int | None` instead of `Optional[int]`

teymour-aldridge:
Still TODO:

- Side balance penalty (hard constraint should be implemented)
- Pullup preference (pullup based on metric – lowest draw strength by wins for WSDC)
- History constraints (I think)

Qanpi:
Hey! 

I've been following this ticket for a while and now have more time to contribute and dive deeper into this, if fine by you. I have a few questions regarding the current status -- forgive me if they're somewhat scattered. 

1. Do I understand correctly that the intention is to have a separate class for (primarily) WSDC draw generation? 
2. The current pullup penalty does not seem to distinguish between the cost of pulling up against number of previous pullups vs. against weakest draw strength. Andrew's proposition assigns different cost multipliers based on this -- is the long-term plan to support that? 
3. At least the following constraints are still missing from the big picture: ~soft side balance~, history, fold, random number. Is that accurate? 
4. Is there a test suite / example case of 'perfect' WSDC draw generation to confirm against? 

Edit: reading further into point 2, I'm not actually sure if a different cost multiplier for these two cases would in any way change the outcome of the global minimum cost solution, since a pullup has to happen no matter what, so it only matters that the costs are appropriately relative to each other locally, which I believe the current solution supports already.

Edit 2: soft side balance seems to already be implemented. So does the hard constraint?

tienne-B:
@Qanpi Your help would always be appreciated!

1. Yes
2. Not really; the system here prioritizes number of pullups to draw strength, and I think it'd add extra complexity on us and on the users to have to calculate different penalties for really similar goals. (As you understand too!)
3. No, those constraints (except for random number which is done by shuffling teams before adding them to the graph), are already implemented in [graph.py](https://github.com/TabbycatDebate/tabbycat/blob/59c4d5fcda86e474be9454ea97873ca85207adf8/tabbycat/draw/generator/graph.py#L26-L51). They are there as they'd are not specific to power-pairing and could apply to random.
4. That's the next step ;P It would be good to add some test cases for it to do what we expect

Qanpi:
Cool! Thanks for a snappy response and overall for being so active on an open-source project :) 

I could start on some basic unit-like tests and build from there, based on what's available in `test_generator.py`. 

Does this sound reasonable to you, or did you already have some specific vision in mind for this?

tienne-B:
Sounds good! Are you able to pull down this branch or would you prefer if this were merged beforehand?

I would first test that each penalty gets correctly applied, and then some full draw tests

Qanpi:
I have no strict personal preference, and it's not a blocker, but I suspect the traceability might be clearer if this was merged first.

That's roughly along the lines of what I was thinking. I'll aim for a few of the tests by the end of this week so you can comment on the direction.

tienne-B:
Sure, merged!

PR #2613, opened 2025-04-06
Fix redis provisioning check in deploy script

merv1n34k:
Hi,

Current script version does not finish properly due to incorrect list index in redis provisioning check.

This PR will ensure `deploy_heroku.py` script works as it checks redis availability with more accurate regex.

### DIFF
--- a/deploy_heroku.py
+++ b/deploy_heroku.py
@@ -212,7 +212,12 @@ print_yellow("Waiting for Heroku Redis to provision (may take up to 5 minutes)..
 while not redis_provisioned:
     time.sleep(30)
     redis_output = subprocess.check_output(redis_status_command).decode().split("\n")
-    redis_provisioned = "available" in redis_output[2]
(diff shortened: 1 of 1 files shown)


deploy_heroku.py
```diff
@@ -212,7 +212,12 @@ def get_git_push_spec():
 while not redis_provisioned:
     time.sleep(30)
     redis_output = subprocess.check_output(redis_status_command).decode().split("\n")
-    redis_provisioned = "available" in redis_output[2]
+    for status in redis_output:
+        match = re.match(r"^Status:\s+available",status)
```
tienne-B, line 216:
```suggestion
        match = re.match(r"^Status:\s+available", status)
```
Linting is complaining

tienne-B (commented):
Hey! Sorry for forgetting about this PR. I look much less at the Heroku deployment steps, but this seems like it may be a little more resilient.

PR #2631, opened 2025-04-28
Master

HTM0410:


tienne-B:
Hey! In order to get the Heroku settings, set the `ON_HEROKU` env var. You can also set `TIME_ZONE` yourself; no need to change the defaults.

What are you trying to do with this PR?

PR #2632, opened 2025-04-28
Turn off redis certificate verification

### LINKED ISSUE #1841: Bypass SSL verification for Heroku Redis
Using an upgraded Heroku Redis dyno can cause an `SSLCertVerificationError` with the message:
```
Exception inside application: SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1108)')
```

The way to solve it is to get Redis to ignore the fault, which I wouldn't consider secure, as shown in the [Heroku help page](https://help.heroku.com/HC0F8CUS/redis-connection-issues).

Personally, think this is just less secure and the problem is on Heroku and their certs.

merv1n34k:
Hi,

From my current understanding of the problem, Heroku's Key-Value Store uses self-signed certificates, and redis attempt to verify certificate results in an error. This issue was raised previously (#1841). 

It was recommended (both on Heroku and channels_redis pages) to disable redis cert verification.

As I have been facing this error on my heroku installation, I assume the error is happening on any tabbycat heroku installation (at least since commit 0728fc5).

**This PR** resolves the issue by turning off redis cert verification.

Reference:
[Heroku key value store connection issues](https://help.heroku.com/HC0F8CUS/heroku-key-value-store-connection-issues)
[django instruction on channels_redis](https://github.com/django/channels_redis?tab=readme-ov-file#hosts)

PS: Heroku [official help page](https://devcenter.heroku.com/articles/connecting-heroku-redis#using-django-redis) provide solution for `django_redis` module which is either outdated or simply incorrect and must not be used.

### DIFF
--- a/tabbycat/settings/heroku.py
+++ b/tabbycat/settings/heroku.py
@@ -87,7 +87,10 @@ CHANNEL_LAYERS = {
     "default": {
         "BACKEND": "channels_redis.core.RedisChannelLayer",
         "CONFIG": {
-            "hosts": [environ.get('REDIS_URL')],
+            "hosts": [{
+                "address": environ.get('REDIS_URL'),
+                "ssl_cert_reqs": None,
+            }],
             # Remove channels from groups after 3 hours
             # This matches websocket_timeout in Daphne
             "group_expiry": 10800,


tienne-B:
Thank you! I was wondering why the Heroku docs weren't helping! You might also want to adapt the `CACHES` as well, as it falls back to Heroku Redis if there's no Redis Cloud.

merv1n34k:
For some reason the Heroku's proposed solution results in error as the `django_redis` complains about option `ssl_cert_reqs`, which for some reason is invalid. Skimming through `redis-py` [docs](https://redis.readthedocs.io/en/stable/examples/ssl_connection_examples.html#Connecting-to-a-Redis-instance-via-SSL,-while-specifying-a-self-signed-SSL-certificate.) that should not be the case. There was an [issue](https://github.com/jazzband/django-redis/issues/353) about it on `jazzband/django_redis`.

My assumption is `django_redis`/`redis-py` updated their source and the current option to define `ssl_cert_reqs` option in `CACHES` might be different. However, I did not look into it as for my case adapting only `CHANNEL_LAYERS` was sufficient. 

It probably would be better to adapt `CACHES` as well, but I am not sure how to do it, I am no expert in the field.

Issue #2646, opened 2025-05-16
Maximum side imbalance is non-quantifiable


Qanpi:
As a user, maximum side imbalance is a very vague preference. 

How is the imbalance measured? 
What value should I set it to? 

Additionally, the graph algorithm already attempts to minimize the imbalance, so hard constraining it doesn't bring that much added value.

My suggestion is that the preference is reverted to represent a hard cap on the number of times a team can debate one side, as per [WSDC rules](https://github.com/TabbycatDebate/tabbycat/issues/2513). That way, an organizer is explicitly made aware if there is a violation of the above-mentioned rule.

Alternatively, at the very least the name in code should be changed from `max_times_on_one_side` to something more accurate.

tienne-B:
> My suggestion is that the preference is reverted to represent a hard cap on the number of times a team can debate one side

~~Isn't it that? It should be!~~

That'd be an easy thing to change if you want a quick fix in your PR (something like this):
```py
max(t1.affs, t1.negs, t2.affs, t2.negs) < options['max_times_on_one_side']
```

I think I got confused with that one, sorry. It wouldn't really ever be a problem though, just based on the number of rounds there are!

Qanpi:
Not quite, unless I'm misunderstanding something.

You can verify this from the test I added in #2647. Team A is assigned two affirmatives, which should not be possible under the `max_times_on_one_side` hard constraint, but the cost is not `None`.

Edit: looks like you got there before me :)

PR #2647, opened 2025-05-16
WSDC graph matching tests

Qanpi:
WIP

- added max_side_balance penalty hard constraint test (might need modifications after #2646)


tienne-B:
Seems to be integrated in the current test suite (from #2648): https://github.com/TabbycatDebate/tabbycat/blob/f1d97f932b5314f9c513cd00f84c236e10be92f1/tabbycat/draw/tests/test_graph_allocations.py#L142-L148

Would seem to be okay to close.

PR #2653, opened 2025-05-21
Adding welcome message for registration form

linh-trant:


tabbycat/registration/templates/wizard_registration_form.html
```diff
@@ -7,6 +7,9 @@
 
     <form class="col-lg-8 ml-lg-auto mr-md-auto d-flex flex-column gap-3" action="." method="POST">
 
+      {% block registration_header%}
```
tienne-B, line 10: If anything, you could do something like add `register_message` to the template context data and then have:
```
    {% if register_message %}
        <div class="card">
            <div class="card-body pb-0">
                {{ register_message|safe }}
            </div>
        </div>
    {% endif %}
```
So that you don't need to create the new template files. What'd you think?

PR #2654, opened 2025-05-25
Add schedule to public page

TrudeauOkech:
In place of most tourneys having to write the schedule in the Welcome Page, we can add  a schedule page that displays the schedule. As it's not "timezone-aware", we also have an info card above the table that allows tabs to set the time zone todisplay on the info card.

This is not really an implementation of https://github.com/orgs/TabbycatDebate/projects/5 per-se, but could be a starting point.

image
image

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -17,7 +17,6 @@ from .models import tournament_preferences_registry
 from .types import MultiValueChoicePreference
 from .utils import validate_metric_duplicates
 
-
 # ==============================================================================
 scoring = Section('scoring', verbose_name=_("Score Rules"))
 # ==============================================================================
@@ -1179,6 +1178,15 @@ class FeedbackProgress(BooleanPreference):
     verbose_name = _("Enable public view of unsubmitted feedback")
     section = public_features
     name = 'feedback_progress'
+    default = False
+
+
+@tournament_preferences_registry.register
+class PublicSchedule(BooleanPreference):
+    help_text = _("Enables the public page showing the schedule")
+    verbose_name = _("Enable public view of shedule")
+    section = public_features
+    name = 'public_schedule'
     default = False
 
 
--- a/tabbycat/tournaments/views.py
+++ b/tabbycat/tournaments/views.py
@@ -8,6 +8,7 @@ from django.contrib.auth import get_user_model
 from django.db.models import Count, Q
 from django.shortcuts import redirect, resolve_url
 from django.utils.html import format_html_join
+from django.utils.timezone import get_current_timezone_name
 from django.utils.translation import gettext_lazy as _
 from django.views.generic.base import TemplateView
 from django.views.generic.edit import CreateView, FormView, UpdateView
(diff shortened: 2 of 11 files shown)


tabbycat/options/preferences.py
```diff
@@ -1182,6 +1181,25 @@ class FeedbackProgress(BooleanPreference):
...
+
+
+@tournament_preferences_registry.register
+class ScheduleTimezoneOptions(ChoicePreference):
+    help_text = _("The timezone to display in the schedule info card")
+    verbose_name = _("Schedule timezones")
+    section = public_features
+    name = 'schedule_timezones'
+    choices = get_common_timezone_choices()
+    default = 'UTC'
```
tienne-B: Why do we want to have a preference for this rather than use the TZ from the environment variables? Having two places for this will be confusing.

TrudeauOkech: Was thinking of tournaments that happen across multiple TZs, especially when their instance is set to a different TZ than they're actually in. But I actually see why it could be even more confusing!

TrudeauOkech: In hindsight, I probably over-thought this :)

tabbycat/tournaments/views.py
```diff
@@ -369,3 +371,99 @@ def dispatch(self, request, *args, **kwargs):
...
     page_subtitle = 'Contextual sub title'
+
+
+class SetTournamentScheduleView(AdministratorMixin, TournamentMixin, ModelFormSetView):
+    template_name = 'tournament_schedule_edit.html'
+    formset_model      = ScheduleEvent
+    form_class         = ScheduleEventForm
+
+    def get_edit_permission(self):
+        return 'schedule.change_scheduleevent'
```
tienne-B: You'll need to add this to `permissions.py`. Then, you could have something like:
```py
edit_permission = Permission.EDIT_EVENTS
view_permission = Permission.VIEW_EVENTS
```

tabbycat/tournaments/views.py
```diff
@@ -369,3 +371,99 @@ def dispatch(self, request, *args, **kwargs):
...
+            self.request.user,
+            self.get_edit_permission(),
+            self.tournament,
+        )
+        kwargs['can_edit']  = can_edit
+        kwargs['save_text'] = 'Save schedule'
+        return super().get_context_data(**kwargs)
+
+    def formset_valid(self, formset):
+        can_edit = has_permission(
```
tienne-B: We wouldn't need a permission check here as it's already handled beforehand.

tabbycat/tournaments/views.py
```diff
@@ -369,3 +371,99 @@ def dispatch(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        can_edit = has_permission(
+            self.request.user,
+            self.get_edit_permission(),
+            self.tournament,
+        )
+        kwargs['can_edit']  = can_edit
+        kwargs['save_text'] = 'Save schedule'
+        return super().get_context_data(**kwargs)
```
tienne-B: I think we could delete this method

PR #2672, opened 2025-06-20
Clarify error message about incorrect "ghost" ordering.

teymour-aldridge:


tabbycat/results/result.py
```diff
@@ -731,7 +731,9 @@ def merge_speaker_order(self, result: BaseDebateResult) -> list[ResultError]:
             if not self.get_ghost(side, pos) and result.get_ghost(side, pos):
                 self.set_ghost(side, pos, result.get_ghost(side, pos))
             elif self.get_ghost(side, pos) and not result.get_ghost(side, pos):
-                errors.append(ResultError("Inconsistent ghost order", "ghost", side, pos))
+                errors.append(
+                    ResultError("Inconsistent order of speeches marked as duplicate (also known as 'ghost' speeches).", "ghost", side, pos),
```
tienne-B:
```suggestion
                    ResultError("Inconsistent marking of duplicate (iron-person) speeches", "ghost", side, pos),
```
Shorter; there's not much space!

tienne-B (commented):
Oh, these messages were not really intended to be shown publicly! As they are, we should mark them for translation (try `gettext_lazy` like `from django.utils.translation import gettext_lazy as _`)

PR #2673, opened 2025-06-20
Add preference to control search engine indexing.

### LINKED ISSUE #1724: Create options for tournaments not to be listed in search engine
There has been several concerns of debaters that they (and their debating history) can be tracked down by googling their name (e.g. Name + "Tab"), and these individuals / institutions have expressed their concerns on their digital footprint (e.g. being tracked by future employers, etc.)

Was wondering if it can be added into list of possible configurations (or when it's deployed), to set whether certain tournament's site wants to be listed in search engine or not? Perhaps by adding `noindex` meta tag (am not sure how Tabbycat Page are generated, but it's just a random idea in my head).

teymour-aldridge:
Fixes #1724 

In the fullest implementation, I think we should also handle non-tournament pages (should users be able to disable indexing of these pages) as well as sites with multiple tournaments.

I think either of:
- add a global preference "disable_indexing" which disables indexing for the whole domain (overriding per-tournament preferences)
- add a preference to disable indexing for global pages (e.g. lists of tournaments) such that the disjoint sum of {global setting preference} + union of all the individual per-tournament settings covers every possible page.

could work?

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -1060,6 +1060,27 @@ public_features = Section('public_features', verbose_name=_("Public Features"))
 
 
 @tournament_preferences_registry.register
+class AllowSearchEngineIndexing(ChoicePreference):
+    help_text = _("Set whether or not search engines can index the tab."
+                  "This is done by adding a <meta name=\"robots\" content=\"noindex\">"
+                  "tag to appropriate pages. Please note that while most search"
+                  "engines (e.g. Google) do respect this, not all search engines"
+                  "will. While this may be useful in preserving participants"
+                  "privacy, you may wish to also consider other techniques"
+                  "(such as disabling public pages or password-protecting your)"
+                  "web server.")
+    verbose_name = _("Search engine indexing configuration")
+    section = public_features
+    name = 'search_engine_indexing'
+    choices = (
+        ('all', _("Allow all pages to be indexed by search engines.")),
+        ('homepage-only', _("Only allow the homepage to be indexed, do not allow any other pages (e.g. tabs, ballots, etc).")),
+        ('none', _("Do not allow any pages to be indexed by search engines.")),
+    )
+    default = 'all'
+
+
+@tournament_preferences_registry.register
 class PublicParticipants(BooleanPreference):
     help_text = _("Enables the public page listing all participants in the tournament")
     verbose_name = _("Enable public view of participants list")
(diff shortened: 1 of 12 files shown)


tabbycat/options/preferences.py
```diff
@@ -1059,6 +1059,27 @@ class SplitVotingBallots(BooleanPreference):
...
+class AllowSearchEngineIndexing(ChoicePreference):
+    help_text = _("Set whether or not search engines can index the tab."
+                  "This is done by adding a <meta name=\"robots\" content=\"noindex\">"
+                  "tag to appropriate pages. Please note that while most search"
+                  "engines (e.g. Google) do respect this, not all search engines"
+                  "will. While this may be useful in preserving participants"
+                  "privacy, you may wish to also consider other techniques"
+                  "(such as disabling public pages or password-protecting your)"
+                  "web server.")
+    verbose_name = _("Search engine indexing configuration")
```
tienne-B: This feels too technical... how about something like:

Help text:
> Adds a flag signaling search engines to not include tournament pages in search results. Note that it may get ignored, and the other available settings will provide greater privacy.

Verbose name:
> Include site in search engines

tabbycat/templates/base.html
```diff
@@ -22,6 +22,20 @@
...
+    {% if pref.search_engine_indexing == "all" %}
+    {% elif pref.search_engine_indexing == "homepage-only" %}
+        {% tournamenturl 'tournament-public-index' as tournament_homepage_url %}
+        {% if request.path == tournament_homepage_url %}
+        {% else %}
+            <meta name="robots" content="noindex">
+        {% endif %}
+    {% else %}
+        <meta name="robots" content="noindex">
+    {% endif %}
```
tienne-B: I would try simplifying the logic and scoping it with a block

```suggestion
    {% if pref.search_engine_indexing != "all" %}
      {% block noindex %}<meta name="robots" content="noindex">{% endblock %}
    {% endif %}
```

Then for the tournament index page, you can add something like:

```html
{% if pref.search_engine_indexing == "homepage-only" %}{% block noindex %}{% endblock %}{% endif %}
```

to override the default here.

tienne-B (commented):
How important would be hiding the site index (tournament list)? For hiding it, I don't really love either of the options, as I want to get rid of the global options, and don't want tournaments to affect others (but guess option 2 would be best).

PR #2676, opened 2025-06-21
Update default adj score range from 0-5 to 1-10

aalbayda:
In most competitions, including [WUDC](https://drive.google.com/file/d/1HNAHw1fcuobidsVSF7C_jbSI65ETz6nT/view), the judge range is 1-10, not 0-5.


tienne-B:
I think we should focus on updating the presets (in `options/presets.py`) if they are out-of-date rather than updating the global default.

PR #2680, opened 2025-06-25
Fixed problem with copying participants data to a CSV

### LINKED ISSUE #2664: Copying csv from the 'participants' page only includes the names (not other columns like speaker categories, adj. core flags etc.)
Image

Pressed the above on the 'Participants' page. Got a copy-pasteable text/csv format that had speaker/adj. names, but no other columns. For instance, the adjcore value did not carry over. 

Image

Not critical, but makes it marginally harder to export data to a spreadsheet. Not sure why this happens, haven't had time yet to look at the code.

JWPCode:
I believe I found the problem, the Boolean data was handled with a different function witch would create an icon that would not be recognised by the function to copy data to the clipboard. Changing this icon to a text emoji fixed this. To my knowledge and testing the emojis work with the CSV file format. 

This is my first Pull request, if there are steps I have missed or any other issues I will be happy to fix them up.

Fixes #2664

### DIFF
--- a/tabbycat/utils/tables.py
+++ b/tabbycat/utils/tables.py
@@ -110,8 +110,8 @@ class BaseTableBuilder:
         - `data` must be an iterable of booleans.
         """
         cells = [{
-            'icon': 'check' if datum else '',
             'sort':  1 if datum else 2,
+            'text': '✔️' if datum else '',
         } for datum in data]
         self.add_column(header, cells)
 
--- a/tabbycat/participants/views.py
+++ b/tabbycat/participants/views.py
@@ -46,7 +46,7 @@ class BaseParticipantsListView(TournamentMixin, VueTableTemplateView):
 
     def get_tables(self):
         adjudicators = self.tournament.adjudicator_set.select_related('institution')
-        adjs_table = TabbycatTableBuilder(view=self, title=_("Adjudicators"), sort_key="name")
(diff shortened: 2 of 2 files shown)


tabbycat/utils/tables.py
```diff
@@ -110,8 +110,8 @@ def add_boolean_column(self, header, data):
         - `data` must be an iterable of booleans.
         """
         cells = [{
-            'icon': 'check' if datum else '',
             'sort':  1 if datum else 2,
+            'text': '✔️' if datum else '',
```
tienne-B, line 114: I think embedding emojis in CSVs may cause problems when pasting into Excel, and using an emoji makes the table look weird as inconsistent with other symbols we use.

I would recommend modifying `tables/SmartTable.vue` instead for how the data is copied. There, we could make it so that the "check" icon gives back a "TRUE" value.

JWPCode: Coming back to this Your 100% right. However I'm not sure my environment is setup properly for some reason. Python code changes would happen perfectly, but none of my changes to SmartTable.vue have an effect, even going so far as to (Temporarily) delete the CSV export code, and the output would outputted CSV would not change .

tienne-B: Sorry to hear that! After modifying Vue code, you might have to run a few commands to bundle the JS:
```
npm run build && dj collectstatic
```

If the `LOCAL_DEVELOPMENT` env var is set, `npm run serve` should be able to auto-reload JS (on refresh).

PR #2701, opened 2025-07-07
Return proper error message when wrong `category` query parameter is provided.

teymour-aldridge:


tienne-B:
Sorry I haven't replied on GitHub about this yet. I just don't think we should be implementing such type checks, as that's the purpose of having serializers, which would be able to have more precise checks (such as that the category _exists_) and being more generalizable for all the possible query parameters that are passed.

I would recommend approaching this by creating new serializers that can take the query parameters, validate them, and then use them in the query. This may look like:
```py
class TeamStandingsView:
    class QueryParameters(Serializer):
        category = PrimaryKeyRelatedField()

    ...
    def get_queryset(self):
        self.QueryParameters(data=self.request.query_params).is_valid(raise_exception=True)
```

There would be more boilerplate, but now it's easier to add more params and have them tested (often with automatic error messages)

tienne-B:
Hey! I've gone ahead with creating serializers for query params (https://github.com/TabbycatDebate/tabbycat/commit/09bd133a54277cff838fd55cd6f2fe62dd57adab), and so I don't think this PR is pertinent anymore. Feel free to re-open if there's anything!

PR #2732, opened 2025-08-09
Implement initial tests for tournament and round APIs

CreatedBySeb:
WIP, posting for initial feedback on tests before proceeding to add more tests for these cases as well as for more resource types

### DIFF
--- a/tabbycat/api/tests/test_serializers.py
+++ b/tabbycat/api/tests/test_serializers.py
@@ -15,13 +15,81 @@ from options.presets import CanadianParliamentaryPreferences
 from participants.models import Adjudicator, Speaker, Team
 from tournaments.models import Round, Tournament
 from utils.misc import reverse_round, reverse_tournament
-from utils.tests import CompletedTournamentTestMixin
+from utils.tests import CompletedTournamentTestMixin, V1_ROOT_URL
 
 User = get_user_model()
 tz = zoneinfo.ZoneInfo('Australia/Melbourne')
 
 
+class TournamentSerializerTests(CompletedTournamentTestMixin, APITestCase):
+
+    def test_base(self):
+        tournament_url = V1_ROOT_URL + "/tournaments/" + self.tournament.slug
+
+        expected = {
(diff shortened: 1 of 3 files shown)


tienne-B (commented):
Few little nits:
* Could use `reverse_tournament` rather than always `reverse` with a kwargs dict (oh, I didn't do that well at this either)
* When passing in data, you shouldn't need to use `json.dumps()` (see `test_remove_breaking_team()`)

It looks like a good approach (and there's not much of a distinction it seems between `test_views` and `test_serializers`...)

CreatedBySeb:
Makes sense, I see there's also actually `reverse_url` on `CompletedTournamentTestMixin` which I can use and that also leverages `round_seq`.

Re. `json.dumps()`, I initially tried it without using this since I saw it wasn't used in one of the other tests, but I get a failure then with a 400, it only works if I call `json.dumps()`. This applies to both patch and post, so I'm not sure what the difference is that is causing it not to work here.

I wasn't entirely sure what the intended difference was between `test_views` and `test_serializers`, so I went with `test_views`. Currently it looks like `test_serializers` focuses more on validation cases while `test_views` is more looking at the API interactions? I can potentially remove the response validation parts of my tests and move them over to `test_serializers` if that is the case.

tienne-B:
> Re. `json.dumps()`, I initially tried it without using this since I saw it wasn't used in one of the other tests, but I get a failure then with a 400, it only works if I call `json.dumps()`. This applies to both patch and post, so I'm not sure what the difference is that is causing it not to work here.

Might be with the `Client` class, between DRF's and Django's different classes.

> I wasn't entirely sure what the intended difference was between `test_views` and `test_serializers`, so I went with `test_views`. Currently it looks like `test_serializers` focuses more on validation cases while `test_views` is more looking at the API interactions? I can potentially remove the response validation parts of my tests and move them over to `test_serializers` if that is the case.

That sounds about right, like making sure the serializer is correctly creating or raising validation errors, and views looking about permissions.

PR #2739, opened 2025-09-22
Created All preferences Page

### LINKED ISSUE #2586: Add a page that shows all preferences
Sometimes it's nice to be able to ctrl+F for a specific preference, which would be easier if there were an option to show all of them at once.

JWPCode:
I saw Issue #2586 requesting a page that had all preferences listed. This Page Has all preferences listed grouped by the section they appear. Under the assumption this page will be used with ctrl/cmd+f there are no section headers. Adding section headers requires more substantial code changes especially to templates and such, I also didn't think they would be necessary with the current seperate pages for each section.

### DIFF
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -1,5 +1,6 @@
 from decimal import Decimal
 
+from django import forms
 from django.core.validators import EmailValidator, MinValueValidator, validate_slug
 from django.forms import SelectMultiple
 from django.utils.translation import gettext_lazy as _
@@ -16,6 +17,25 @@ from tournaments.utils import get_side_name_choices
 from .models import tournament_preferences_registry
 from .types import MultiValueChoicePreference
 from .utils import validate_metric_duplicates
+
+
+# ==============================================================================
+all_sections = Section('all_sections', verbose_name=_("All Preferances"))
+# ==============================================================================
+
+
+@tournament_preferences_registry.register
+class DummyAllSectionsPreference(StringPreference):
+    section = all_sections
+    name = 'dummy_all_sections'
+    default = ''
+    verbose_name = 'All Sections Placeholder'
+
+    def get_field_kwargs(self):
+        kwargs = super().get_field_kwargs()
+        kwargs['widget'] = forms.HiddenInput
+        return kwargs
+
 
 # ==============================================================================
 scoring = Section('scoring', verbose_name=_("Score Rules"))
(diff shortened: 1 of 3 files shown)


tabbycat/options/preferences.py
```diff
@@ -17,6 +18,25 @@
 from .types import MultiValueChoicePreference
 from .utils import validate_metric_duplicates
 
+
+# ==============================================================================
+all_sections = Section('all_sections', verbose_name=_("All Preferances"))
```
teymour-aldridge:
```suggestion
all_sections = Section('all_sections', verbose_name=_("All Preferences"))
```

tabbycat/options/templates/preferences_index.html
```diff
@@ -82,6 +82,12 @@
       {% trans "Settings which can affect all tournaments on the site" as description %}
       {% include "components/item-action.html" with type="primary" emoji="🌏" text=title extra=description %}
 
+      {% tournamenturl 'options-tournament-section' section='all_sections' as url %}
+      {% trans "Preferance List" as title %}
```
teymour-aldridge:
```suggestion
      {% trans "Preference List" as title %}
```

tabbycat/options/templates/preferences_index.html
```diff
@@ -82,6 +82,12 @@
...
+      {% trans "All preferance options in one place" as description %}
```
teymour-aldridge:
```suggestion
      {% trans "All preference options in one place" as description %}
```

tabbycat/options/preferences.py
```diff
@@ -17,6 +18,25 @@
...
+class DummyAllSectionsPreference(StringPreference):
+    section = all_sections
+    name = 'dummy_all_sections'
+    default = ''
+    verbose_name = 'All Sections Placeholder'
+
+    def get_field_kwargs(self):
+        kwargs = super().get_field_kwargs()
+        kwargs['widget'] = forms.HiddenInput
+        return kwargs
```
tienne-B, line 37: Why do we need a dummy preference?

JWPCode: Its been a while since I did it sorry, but Its because of how the preferences are handled and how they have been put into sections. The default when there is no argument passed doesn't work with this set up, So we needed a new section to pass, that could then have the other sections added to, however it needs an initial preference to be returned or it throws an error

PR #2761, opened 2025-10-22
Always enable the API.

teymour-aldridge:


tienne-B:
I love net negative-line PRs!

PR #2765, opened 2025-10-30
Fix/skip missing local settings

abusumon:


tienne-B:
Hey @abusumon ! This looks like a nice PR, but I wonder why the Heroku deployment doesn't automatically set the `ON_HEROKU` environment variable (which would then prevent trying to load `local.py`)

PR #2828, opened 2026-03-01
Align local dev settings and inline scripts with the Vite workflow

Cyberaktiv21:
- Added the USE_VITE_SERVER flag and set it to be on by default in settings/local.example. This was done after renaming the old USE_WEBPACK_SERVER flag to the new USE_VITE_SERVER, so the hmr context flag now reflects the dev server introduced in the recent “Migrate Templates to local Vite HMR Setup” PR. Without adding this flag, participant lists and standings pages silently failed to load because the templates never received an active HMR signal.

- Updated all inline script blocks to declare type="module" (availability, draw admin/by, private URLs, results, tournaments, venues, etc.), keeping them consistent with Vite’s module graph. Previously these scripts were treated as classic scripts, causing buttons like “Generate Draw” (and other inline-script actions) to hang because Vite couldn’t process the inline code; marking them as modules resolves those interactions

### DIFF
--- a/tabbycat/availability/templates/availability_index.html
+++ b/tabbycat/availability/templates/availability_index.html
@@ -233,7 +233,7 @@
 
 {% block js %}
   {{ block.super }}
-  <script>
+  <script type="module">
     $(document).ready( function() {
       $("#createDraw").click(function(event) {
         $.fn.loadButton(event.target); // Prevent double submission
--- a/tabbycat/draw/templates/draw_display_admin.html
+++ b/tabbycat/draw/templates/draw_display_admin.html
@@ -446,7 +446,7 @@
 
 {% block js %}
   {{ block.super }}
-  <script>
+  <script type="module">
     $(document).ready( function() { // Use the fake submit buttons as real submission
       $("#triggerReleaseDrawForm").click( function() {
         $("#releaseDrawForm").submit();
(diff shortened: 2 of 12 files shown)


tienne-B (changes requested):
Careful; this commit has some merge conflicts and is reverting some recent commits

Cyberaktiv21:
I apologise for the errors - I forgot to pull the most recent commits. Looks like the JQuery timing issue has been fixed. However, on my local instance - the draws, participant info, standings (maybe pages with a certain class of vue components) fail to load. The issue is resolved when hot reload is enabled with the USE_WEBPACK_SERVER flag set to true (although web pack has been switched to vite , this flag might need a rename). I restarted the project multiple times (copied the static files as well with collectstatic) yet the issue persists.  I understand that enabling HMR for local development is part of the workflow but will it be sustainable for production?

tienne-B:
> I understand that enabling HMR for local development is part of the workflow but will it be sustainable for production?

@Cyberaktiv21 You're right that HMR shouldn't be activated for production; I don't know where that got broken, but I appreciate the work to fix it to be optional rather than mandatory.

Cyberaktiv21:
Should I create a new PR with a rename of USE_WEBPACK_SERVER to reflect the switch to Vite and maybe add HMR to local.example for now? I am quite new to this so I apologise if I am doing anything incorrectly - my apologies

PR #2829, opened 2026-03-01
docs: Document Redis requirement for local installs

Cyberaktiv21:
- Automatic adjudicator allocation was stalling on “Loading…” whenever a local developer hadn’t configured Redis and a background worker. Channels-based allocation tasks rely on Redis as the broker, so without it the modal never receives a response
- Updated the Linux install guide to include Redis in the short dependency list and added a dedicated subsection with install/start commands 
- Added a Redis installation subsection to the macOS guide with Homebrew instructions 
- Added a Redis prerequisite note to the adjudicator allocation guide so local users know to run Redis and a worker when the modal appears to hang 
- Extended the local settings example with a default CHANNEL_LAYERS config pointing to the local Redis instance


tienne-B (commented):
Looks good! Redis is also used for caching, if you'd want to note that as well.

PR #2832, opened 2026-03-05
Add option to hide metrics on public break page

Cyberaktiv21:
PR based on an idea shared by @tienne-B 
"As we release the break but want to avoid revealing speaks or extra metrics, we are left with two unideal options: temporarily removing speaks from the standings precedence, or not releasing the break on Tabbycat."

Fix : 
- Adds a "Hide metrics on public break page" toggle under Public Features in preferences.py
- When enabled, speaker score and draw strength  columns are hidden from the public break page, while keeping points, 1sts and 2nds visible

### DIFF
--- a/tabbycat/breakqual/views.py
+++ b/tabbycat/breakqual/views.py
@@ -85,6 +85,23 @@ class PublicBreakingTeamsView(PublicTournamentPageMixin, BaseBreakingTeamsView):
     public_page_preference = 'public_breaking_teams'
     cache_timeout = settings.PUBLIC_SLOW_CACHE_TIMEOUT
 
+    def get_table(self):
+        metrics_to_show = self.tournament.pref('public_break_metrics_to_show')
+        self.standings = self.get_standings()
+
+        if metrics_to_show != -1:
+            n = max(0, metrics_to_show)
+            self.standings.metric_keys = self.standings.metric_keys[:n]
+            self.standings._metric_specs = self.standings._metric_specs[:n]
+
+        table = TabbycatTableBuilder(view=self, title=escape(self.object.name), sort_key='Rk')
+        table.add_ranking_columns(self.standings)
+        table.add_column({'title': _("Break"), 'key': 'break'},
+                         [tsi.break_rank for tsi in self.standings])
+        table.add_team_columns([tsi.team for tsi in self.standings])
+        table.add_metric_columns(self.standings)
+        return table
+
 
 class GenerateBreakMixin:
 
--- a/tabbycat/options/preferences.py
+++ b/tabbycat/options/preferences.py
@@ -1190,6 +1190,17 @@ class PublicBreakingTeams(BooleanPreference):
 
 
 @tournament_preferences_registry.register
+class PublicBreakMetricsToShow(IntegerPreference):
+    help_text = _("How many metrics from the team standings precedence to show on the public break page. "
+                  "For example, 2 shows only the first two metrics set in Standings. "
+                  "Use 0 to hide all metrics, or -1 to show all of them.")
+    verbose_name = _("Number of metrics to show on public break page")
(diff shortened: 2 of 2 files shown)


tienne-B:
@Cyberaktiv21 I don't think we should have an allow-list. Rather, we should just hide everything after the first precedence metric.

Cyberaktiv21:
What are the metrics that we want to ultimately hide? I thought that we were removing speaks and draw strength as these are the two metrics that are sensitive...

tienne-B:
What would you think of this approach?
1. Have a setting "Number of metrics to show in public break" (or a better wording)
2. In `add_metric_columns()`, add a parameter to limit the number of metric columns that can be added
3. If admin or setting == -1, in `get_table()` show all columns, else pass the number to `add_metric_columns()`

I see a few situations where this preference could be used:
* If the tournament wants no metrics shown, they can use "0"
* If the tournament doesn't want to display speaks, they can use "1"

I don't think we should only disallow metrics based on speaks, as a table that shows points, and then firsts (like for BP) is misleading for teams whose break order doesn't reflect this apparent precedence as speaks were omitted.

Cyberaktiv21:
I am still a bit confused...
Is this the view that was originally desired in the context of solving the problem for WUDC?
image

Also when only speaks are to be hidden, is this the view that is expected?
image

I can work on your approach but I still need some more clarity on what combination needs to be hidden within [points, speaker scores, firsts, seconds, draw strength] for each numerical input in your suggested approach . If the options are only within the two screenshots that I have attached , then this can be a simple drop down list.

tienne-B:
> Is this the view that was originally desired in the context of solving the problem for WUDC?

Some tournaments may want that, yes. That would be obtained with "0" as the setting.

> Also when only speaks are to be hidden, is this the view that is expected?

No, as it should also exclude 1sts, 2nds, etc. Showing those metrics without speaks before is misleading, and tournaments that don't show speaks on break slides also don't include those subsequent metrics. This would be "1" as the setting.

linh-trant:
> I can work on your approach but I still need some more clarity on what combination needs to be hidden within [points, speaker scores, firsts, seconds, draw strength] for each numerical input in your suggested approach . If the options are only within the two screenshots that I have attached , then this can be a simple drop down list.

I don't think we would want any "combination for hiding" at all. Tab Directors can choose any metrics (and their priorities) for team ranking, and we have an option to hide them based on their priorities, regardless of what _are_ these metrics. And so, an approach to show the top `x` most priority metric is more ideal in this situation (especially when there are no such situations that you want to show the first metric (e.g, points in BP) and the 4th metric (e.g, number of seconds in BP), as it is misleading).

So what Étienne means here is we would want to have a feature `Number of metrics to show in public break`; for example, when this setting is set to 2, the first two metrics for ranking would be shown (regardless of what those metrics are; could be speaks, could be something else). When it is set to 0, no metric would be shown (like what you want to achieve with WUDC in your screenshot), and when it is set to -1, all metrics would be shown. 

Also, @tienne-B, I think hiding the ranking (the Break column) when we want to show no metrics is a good idea (or a disclaimer that ranking here is done alphabetically if there is not such a warning already (but I think we already did?)), just in case of some confusion.

Cyberaktiv21:
Thank you for the suggestion, @tienne-B ! The new approach and your explanation were very clear, and I appreciated the elegance of the solution once I started working on it.

Thank you as well, @linh-trant , for the detailed breakdown. Your explanation helped me realise that the break metrics are derived from the standings preference -  this made the whole picture much clearer and easier to follow, and helped me build a stronger mental model of how the solution needed to be built. I really appreciate the time you took to explain things so clearly and engage with my concerns so patiently. As a beginner, it means a lot and keeps me motivated.

Apologies for the slow response and development time. I'll aim to be more active going forward!

PR #2835, opened 2026-03-08
jQuery validate broken on ballot entry pages

merv1n34k:
Hey! 

Spotted another error on `develop`: when opening ballot edits, places are not rendered, dev log shows error with `JQuery` validation. Running on local docker install, see the report below.

### Problem

>$(...).validate is not a function error on all ballot entry pages (enter_results, public_enter_results, public_enter_results_error). Ballot form validation does not run — speaker positions, scores, and other fields are not validated client-side.

### Steps to reproduce

1. Navigate to any ballot entry page (admin or public)
2. Open browser console
3. `Error: jquery.js:4059 Uncaught TypeError: $(...).validate is not a function at HTMLDocument.<anonymous>`

### Root cause

`app.js` is loaded as a Vite module (`<script type="module">`), which is deferred. It imports its own jQuery and overwrites `window.$` / `window.jQuery`. The `jquery.validate.js` plugin is attached to the original jQuery instance loaded via `<script>` tag, but by the time `$(document).ready()` callbacks fire, `$` already points to the bundled instance — which lacks the validate plugin.

### Fix

Wrap the inline script in an IIFE `(function($) { ... })(jQuery)` to capture the jQuery reference at evaluation time (when the validate plugin is still attached), before `app.js` overwrites the global.

### DIFF
--- a/tabbycat/templates/js-bundles/jquery-shim.js
+++ b/tabbycat/templates/js-bundles/jquery-shim.js
@@ -1,0 +1,4 @@
+// Re-export the global jQuery loaded via <script> tag in base.html.
+// This ensures the Vite bundle (and Bootstrap) use the same jQuery
+// instance that jquery.validate.js and other plugins attach to.
+export default window.jQuery
--- a/tabbycat/templates/js-bundles/main.js
+++ b/tabbycat/templates/js-bundles/main.js
@@ -3,8 +3,8 @@ import { createApp, defineAsyncComponent } from 'vue'
 import { createPinia } from 'pinia'
 import * as Sentry from '@sentry/vue';
 import feather from 'feather-icons'
-import $ from 'jquery'
 import 'bootstrap' // Import bootstrap javascript plugins
+const $ = window.jQuery
 
 // Generic Templates
 import CheckboxTablesContainer from '../tables/CheckboxTablesContainer.vue'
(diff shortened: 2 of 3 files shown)


merv1n34k:
Hey @tienne-B! Have you got a chance to look at the PR?

tienne-B:
@merv1n34k Hey! Might the problem have been caused with https://github.com/TabbycatDebate/tabbycat/commit/80af8a09d57991e25b8f0712a5f324757cc0bd17#diff-57a03b05601e4873e6bc8e18bedacfa1c84dec0f78998f27e0fec0e0aa2617d3 ? If we remove that new line in `base.html`, does stuff get fixed? If not, we can try this.

merv1n34k:
Hmm, I think the root cause is that we have 2 JQuery instances present. The simples fix actually would be to remove `import $ from 'jquery'` in `main.js` and just use `const $ = window.jQuery`. 

What I see now my fix would only work on ballots, and any future use of JQuery validation will require new IIFE.

merv1n34k:
I can rebase my PR, but it's fine if we close it and you just commit to the `develop` yourself.

merv1n34k:
Fixed with more clean approach: The Vite bundle was importing its own copy of jQuery from `node_modules`, creating two jQuery instances and breaking `jquery.validate.js` which attached to the original global one. Fixed by aliasing all `jquery` imports to a shim that re-exports `window.jQuery`, so the bundle and vendor scripts share a single instance.

tienne-B:
Looks nice! Thank you!

PR #2838, opened 2026-03-20
Add ability to add and remove speakers from teams in admin interface

### LINKED ISSUE #2689: Create mechanism to add/remove speakers in a team
On an admin team page, it should be possible to open up a speaker form to create a new speaker with reg questions, and potentially delete existing speakers. This would avoid the common practice of renaming existing speakers in Edit DB.

TrudeauOkech:
Should fix #2689 
Screenshot 2026-03-21 at 02 13 20


tabbycat/participants/templates/team_registration_card.html
```diff
@@ -123,17 +123,33 @@ <h4 class="card-title mb-0">
             </ul>
           </details>
 
+          {% haspermission "add.team" as add_team_perm %}
```
tienne-B, line 126: Should this section be renamed? I think that having these action buttons in a section called "Custom Questions" could lead to people thinking that the answers are what will be deleted, not the speaker itself.

Also, a different permission should be used (like "delete speaker") for deletion.

tabbycat/participants/views.py
tienne-B, line 1: Small things, but we should use `form_class` for both new form views, and then add the object in `get_form_kwargs()` rather than overriding `get_form()`

Also, for creation, I'm not sure the registration form would be the best, as a lot of fields may be missing, and admins may not want to answer the custom fields.

tabbycat/participants/templates/team_registration_card.html
tienne-B, line 1: Just an idea, but maybe add an "Edit" button to the top of the registration card to send the user to the team's "Edit DB" page?

tienne-B (commented):
Sorry for the delay in the review!

tabbycat/participants/templates/team_registration_card.html
TrudeauOkech: Yeah that's a fair point, will rename it. And will add a separate permission for deletion.

tabbycat/participants/views.py
TrudeauOkech: Will do, good catch.

TrudeauOkech: Makes sense, I'll swap it out for a plain ModelForm on Speaker instead.

tabbycat/participants/templates/team_registration_card.html
TrudeauOkech: Good idea, I'll add that.

tabbycat/participants/templates/speaker_create.html
tienne-B, line 1: Maybe could this be based on the regular speaker form from `registration/forms.py`, but with all fields (and made optional)?

Also, don't think we need the participants subnav here.

TrudeauOkech: Hey, sorry hadn't had time to look at this. Let me make the change

PR #2840, opened 2026-03-28
Fix: Read PORT environment variable for Render deployment

Sabbir-Asif:
issue:

- Currently while deploying on renderer I was facing this issue
image

it is because it was not getting the port for binding.

Solution: 
- Now it reads the environment first otherwise it uses the default port


tienne-B:
Thank you, this change looks good (and Render was a little under my radar)

PR #2845, opened 2026-04-17
Add configurable adjudicator feedback score step

alekhbhardwaj:
Adds a new tournament preference (adj_score_step) to control the increment step allowed when entering adjudicator feedback scores. Options are any decimal, 0.5 increments, or whole numbers only.

The setting is editable via a new modal on the feedback overview page (accessible from the Score Step link in the breakdown card) and is enforced on all feedback submission forms via the NumberInput step attr.

### DIFF
--- a/tabbycat/adjfeedback/forms.py
+++ b/tabbycat/adjfeedback/forms.py
@@ -79,9 +79,11 @@ class BaseFeedbackForm(CustomQuestionsFormMixin, forms.Form):
         # Feedback questions defined for the tournament
         adj_min_score = self._tournament.pref('adj_min_score')
         adj_max_score = self._tournament.pref('adj_max_score')
+        adj_score_step = self._tournament.pref('adj_score_step')
         score_label = mark_safe(_("Overall score (%(min)d=worst; %(max)d=best)*") % {
                 'min': int(adj_min_score), 'max': int(adj_max_score)})
-        self.fields['score'] = forms.FloatField(min_value=adj_min_score, max_value=adj_max_score, label=score_label)
+        self.fields['score'] = forms.FloatField(
+            min_value=adj_min_score, max_value=adj_max_score, step_size=adj_score_step, label=score_label)
 
         self.add_question_fields()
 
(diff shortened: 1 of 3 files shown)


tienne-B:
Hey! Could we simplify this, and make it more in line with other range settings? The setting should take a number and default to 1, and we don't need a modal to edit it as it'll be in the settings with the min/max settings.

alekhbhardwaj:
Hey! Certainly. I also clamped the increment range to the maximum expected feedback score as precaution. I figured the _whole number | 0.5 step_ choices would be more intuitive but I hope speaker score step will be referenced and kept in mind- in case of confusion :) 
And removed the modal as well.

tabbycat/adjfeedback/forms.py
```diff
@@ -79,9 +79,13 @@ def _create_fields(self):
...
         adj_min_score = self._tournament.pref('adj_min_score')
         adj_max_score = self._tournament.pref('adj_max_score')
+        adj_score_step = self._tournament.pref('adj_score_step')
+        adj_score_step = max(0.001, min(adj_score_step, adj_max_score)) if adj_max_score > 0 else 1.0
         score_label = mark_safe(_("Overall score (%(min)d=worst; %(max)d=best)*") % {
                 'min': int(adj_min_score), 'max': int(adj_max_score)})
-        self.fields['score'] = forms.FloatField(min_value=adj_min_score, max_value=adj_max_score, label=score_label)
+        self.fields['score'] = forms.FloatField(
+            min_value=adj_min_score, max_value=adj_max_score, label=score_label,
+            widget=forms.NumberInput(attrs={'step': adj_score_step}))
```
tienne-B:
```suggestion
        self.fields['score'] = forms.FloatField(
            min_value=adj_min_score, max_value=adj_max_score, step_size=adj_score_step, label=score_label)
```

tienne-B (commented):
Looks better! I think we can simplify it even more though. I don't think we need the `max(0.001, min(adj_score_step, adj_max_score)) if adj_max_score > 0 else 1.0` nor the js block in `preferences_section_set.html‎`.

Also, lint your changes, so we aren't adding the extra whitespace.

alekhbhardwaj:
Made suggested changes, hopefully all good now!

tienne-B (approved):
Looks good! I'll merge when the CI passes

PR #2847, opened 2026-04-19
Improve registration form UI

### LINKED ISSUE #2844: Bad handling of forms choice options
Hey! 

The form also have some modes with slightly off rendering (see the image below).

Image

Image

If this is not an expected behaviour I can make a PR.

merv1n34k:
Hey!

Closes #2844.

`RadioSelect`, `CheckboxSelectMultiple`, and `CheckboxInput` widgets were wrapped with Bootstrap's form-control class, causing misaligned bordered containers. Add `is_choice_widget` template filter to detect these widgets and render them without `form-control`. Zero out extra label padding for form-check elements inside `list-group-items`.

The fix looks correct, other parts look unaffected, as in most cases they use Vue and not Django Framework.

Screenshot 2026-04-19 at 18 52 20
Screenshot 2026-04-19 at 19 41 00
Screenshot 2026-04-19 at 19 41 05

### DIFF
--- a/tabbycat/templates/components/form-field.html
+++ b/tabbycat/templates/components/form-field.html
@@ -11,7 +11,11 @@
     </label>
   {% endif %}
 
-  {{ field|addcss:"form-control" }}
+  {% if field|is_choice_widget %}
+    {{ field }}
+  {% else %}
+    {{ field|addcss:"form-control" }}
+  {% endif %}
 
   {% if field.help_text  %}
     <small class="form-text text-muted">{{ field.help_text|linebreaks }}</small>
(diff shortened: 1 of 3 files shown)


tienne-B (commented):
I think I prefer the big checkboxes in settings:
Image

But this PR seems to make them small:

Image

merv1n34k:
Okay, should we have big checkboxes in forms as well, for consistency?

tienne-B:
@merv1n34k I'm not quite sure; I think I'd prefer small checkboxes for the feedback form as I'm used to them being small there (also could say something about them being in a set and it could take too much space). What'd you think?

merv1n34k:
Yeah, make sense, I will revert only the ones in the setting

merv1n34k:
Hey! Now, the checkboxes in settings should be as before, and checkboxes in forms are small. 

If you want I can squash commits

PR #2856, opened 2026-04-24
Show ordinal pull-up count in draw remarks

alekhbhardwaj:
When a team appears in the draw with a pull-up flag, the draw remarks column now shows how many times they have been pulled up in total (up to and including the current round), as an ordinal — e.g. "Pull-up team (2nd pull-up)" instead of just "Pull-up team".

This makes it easier for tab directors to spot teams who have been repeatedly pulled up.

### DIFF
--- a/tabbycat/utils/tables.py
+++ b/tabbycat/utils/tables.py
@@ -796,10 +796,17 @@ class TabbycatTableBuilder(BaseTableBuilder):
             # conflicts is a list of (level, message) tuples
             conflicts = [("secondary", _draw_flags_dict.get(flag, flag)) for flag in debate.flags]
             if not debate.is_bye:
-                conflicts += [("secondary", "%(team)s: %(flag)s" % {
+                for dt in debate.debateteams:
+                    for flag in dt.flags:
+                        pullup_count = getattr(dt, '_pullup_count', 0)
+                        if flag == 'pullup' and pullup_count > 0:
+                            flag_text = _("Pull-up team (%(ordinal)s pull-up)") % {'ordinal': ordinal(pullup_count)}
+                        else:
+                            flag_text = _draw_flags_dict.get(flag, flag)
+                        conflicts.append(("secondary", "%(team)s: %(flag)s" % {
                             'team': self._team_short_name(dt.team),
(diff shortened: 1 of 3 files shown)


tienne-B (commented):
(Sorry for the late reply)

I love this idea! As a suggestion, we could add the "npullups" team metric to the admin draw page standings generator extra metrics
https://github.com/TabbycatDebate/tabbycat/blob/19ec9db3ca7281736c6b5c5814090bc87a3ad3c7/tabbycat/draw/views.py#L509-L510 and so we wouldn't need the other changes to `draw/prefetch.py` nor `draw/views.py`. Should simplify this change quite a bit!

This would only apply for the detailed draw page, but I think that'd be enough, and tab teams will necessarily see it before confirming (and can see afterwards too).

alekhbhardwaj:
Hi. 
Made changes as per suggestion, now piggybacking on the generator for the count. I prefer this route as well!

tabbycat/utils/tables.py
```diff
@@ -789,17 +789,27 @@ def fmt(dt):
...
             if not debate.is_bye:
-                conflicts += [("secondary", "%(team)s: %(flag)s" % {
+                for dt in debate.debateteams:
+                    for flag in dt.flags:
+                        if flag == 'pullup' and standings is not None:
+                            try:
+                                prev_pullups = standings.get_standing(dt.team).metrics.get('npullups', 0) or 0
+                            except ValueError:
+                                prev_pullups = 0
+                            flag_text = _("Pull-up team (%(ordinal)s pull-up)") % {'ordinal': ordinal(prev_pullups + 1)}
```
tienne-B, line 803: I think this is quite defensive; do you see any situations where we'd get a `None` or an exception?

```py
prev_pullups = standings.get_standing(dt.team).metrics['npullups']
flag_text = _("Pull-up team (%(ordinal)s pull-up)") % {'ordinal': ordinal(prev_pullups + 1)}
```

alekhbhardwaj: It is indeed. Perhaps I thought too far while assuming npullups is volatile in certain metric/extra_metric combinations.

PR #2857, opened 2026-04-25
Prioritize current tournament in navigation lists

merv1n34k:
Hey! 

This is a small, but neat QoL improvement, which puts currently edited tournament first in the list. Very useful when you can have many tournaments, and you don't have to scroll to the bottom of the page to see it. Does not modify original formset in DB.

### DIFF
--- a/tabbycat/utils/context_processors.py
+++ b/tabbycat/utils/context_processors.py
@@ -1,14 +1,16 @@
 from django.conf import settings
+from django.db.models import Case, Q, Value, When
 
 from tournaments.models import Tournament
 
 
 def debate_context(request):
+    tournaments = Tournament.objects.filter(active=True)
 
     context = {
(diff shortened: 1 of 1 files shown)


tabbycat/utils/context_processors.py
```diff
@@ -1,14 +1,16 @@
 from django.conf import settings
+from django.db.models import Case, Value, When
 
 from tournaments.models import Tournament
 
 
 def debate_context(request):
+    tournaments = Tournament.objects.filter(active=True)
```
tienne-B, line 8: Might want to add:
```py
Q(active=True) | Q(id=request.tournament.pk)
```
So that it'd work even if editing an inactive tournament?

merv1n34k: Done

merv1n34k:
Hey @tienne-B!

Do you think it is ready to merge?

PR #2880, opened 2026-06-24
Feature/n1 rule

maho10:
# N-1 Rule Tracking module 🐉
This is a small module aimed to help academic teams keep track of the n rule throughout the tournament. Allowing academic teams to be able to see which teams qualify for breaks, and to be able to ask for teams to pay waivers as early as it is not possible for an adjudicator to judge the minimum of rounds. This helps to make the process more transparent, and avoid leaking possible breaking teams while asking for teams to pay wavers.

### Assignments Page
Allows allocation of adjudicators to both institutions and independent teams. It is based on the same logic as adjudicator-institution conflicts, so it autofills and works largely in the same way. Adjudicators can be freely changed from institution, or deleted from their assigned institution and assigned to an independent team.

Sections are in the same forms, but visually divided, so that it is more intuitive

The assignments page is only available on the section of Import Data, as I felt it should be in the same space as  conficts, as the trcking and assignment is similar. It is also available from the compliance dashboard.

### Compliance Dashboard
It provides a real-time overview of the N-1 compliance across the tournament. For each institution and independent team, the dashboard displays whether the requirement has been met. Rows can be expanded to show assigned adjudicators and the number of rounds each has judged, so that it all can be easily tracked and team can be adviced with enough time. 

Waivers can be recorded directly within the module and are automatically considered when calculating compliance. In the same line, the minimum number of rounds (M) required for compliance can be configured per tournament and is automatically saved when modified.

There are different colors and emojis realted to the compliance so that it it visually easy to identify
#### For institutions
- ✅ Green: Fully compliant
- ⚠️ Orange: Partially compliant. They haven't met the requirements yet, but at leat one waiver has been payed. It will not turn green until the institution is fully compliant, either by adding more waivers or assigning adjudicators that compli with the rule minimum.
- ❌ Red: Not compliant. They can have adjudicators assigned, but they haven't participated in the minimum of rounds yet
#### For independent teams
- ✅ Green: Fully compliant
- ❌ Red: Not compliant. They can have an adjudicator assigned, but they haven't participated in the minimum of rounds yet

### Compliance Modes
Tournaments can choose between two modes for evaluating compliance:

- Strict mode (in spanish *Personalizada*): every required adjudicator must individually judge at least M rounds.
- Non-strict mode (in spanish *Despersonalizada*): adjudicators may collectively satisfy the requirement as long as the total number of judged rounds reaches the required threshold.
     - For example, if an institution has 3 teams, it requires 2 adjudicators. If the minimum rounds requirement is 3, the institution must provide at least 6 adjudicated rounds in total. This could be satisfied by 3 adjudicators judging 2 rounds each.
- N=N mode: For when organization decides, institutional teams can also be asked to have n adjudicators for n teams

-----------------------------------
Any coments or corrections are welcome!!
Majo Gil 🐉


tienne-B:
Wow, this is a huge PR and touches so much; I wonder if it'd be better to break it up into a few different PRs.

The main thing is that I don't think we should call this "N-1" as it would be to support more general adjudicator requirements (and it does support "N=N" right now), to be possibly extended to other requirements like N/2.

I've also never heard of "Strict institutional N-1 enforcement" being used for institutional break eligibility further than just that individual adj's eligibility.

Might you have an opinion on adding FK fields to `Adjudicator` for the inst/team they're fulfilling the requirements for, rather than a M2M relationship (like I don't see a situation where an adjudicator is fulfilling many insts' requirements)?

I'm not convinced about having a fining system just for this, maybe it'd be better left with https://github.com/TabbycatDebate/tabbycat/pull/2747 (should get back to working on it...) as a more general invoicing system? We could change the new model up to be something like "requirement exception" (and perhaps link which rounds it applies for?).

There might be some scope creep; why do we need to see which adjs for each inst in each round? In my mind it doesn't matter who as long as the requirements are fulfilled. I think we could simplify and use the existing requirements table (or expand), and if we've reached the limits of our tables system then we should get rid of that page in favour of the new component.

PR #2884, opened 2026-07-05
Feature/working category sort

### LINKED ISSUE #2517: Sorting: speaker categories
In the participants list, the button to sort by speaker categories does not work.
Same applies to speaker eligibility, cannot sort by those who's categories are checked or not

eldonmacdonald:
#fixes #2517 

I'm new here, let me know if there's anything wrong with the fix and I'll work to solve it. I've added tests for regression - it tests for participants, but the logic also fixes the problem with sorting by break category for teams.

### DIFF
--- a/data/fixtures/after_round_4.json
+++ b/data/fixtures/after_round_4.json
@@ -6361,24 +6361,12 @@
     }
 },
 {
-    "model":"participants.speakercategory",
-    "pk":1,
-    "fields": {
-        "tournament": 1,
-        "name": "Novice",
-        "slug": "novice",
-        "seq": 1,
-        "limit": 0,
-        "public": true
-    }
-},
-{
     "model": "participants.speaker",
     "pk": 1,
     "fields": {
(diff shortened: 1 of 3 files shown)


tienne-B:
@eldonmacdonald Thank you for the contribution! It seems like an effective improvement and it makes sense, so I've merged it.

How'd you get the Selenium tests to run? I normally just skip them.

eldonmacdonald:
> @eldonmacdonald Thank you for the contribution! It seems like an effective improvement and it makes sense, so I've merged it.
> 
> How'd you get the Selenium tests to run? I normally just skip them.

I was having problems with Selenium as well. I found the issue and solved it temporarily through my local.py settings, but I've created a more permanent solution to the Selenium testing problem and have created a new pull request for that solution: https://github.com/TabbycatDebate/tabbycat/pull/2885

I hope that the solution will help make UI testing easier (and more commonplace!)

PR #2885, opened 2026-07-06
Fixed Selenium testing

eldonmacdonald:
The SeleniumTestCase is a subclass of StaticLiveServerTestCase. StaticLiveServerTestCase starts a full Django server in the background that serves static files as though the server were running with Debug=True (e.g. serves from STATICFILES_DIRS instead of STATIC_ROOT) but templates HTML as though the server were running with Debug=False (e.g. uses paths to compressed static files in STATIC_ROOT instead of paths to uncompressed files in STATICFILES_DIRS). I have added a settings override for SeleniumTestCase that asks the server to template files with the paths to static files as they would be if Debug=True were set, and this fixes the discrepency without forcing the tester to run "dj collectstatic" with every test run.

### DIFF
--- a/tabbycat/utils/tests.py
+++ b/tabbycat/utils/tests.py
@@ -6,7 +6,7 @@ from unittest import expectedFailure
 from django.contrib.auth import get_user, get_user_model
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core.cache import cache
-from django.test import tag, TestCase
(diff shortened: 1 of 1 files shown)


eldonmacdonald:
To be more specific, when running Django tests, Django sets Debug=False no matter what is in your settings, but StaticLiveServerTestCase explicitly serves static files from STATICFILES_DIRS and not STATIC_ROOT.

PR #2889, opened 2026-07-10
Fix unwanted scrolling on draggable elements in Firefox

franga2000:
In Firefox, the allocation screens scrolls up while you drag the judge/panel/room. This is due to Firefox not exposing mouse position in the drag event handler. It's a [very old bug](https://bugzilla.mozilla.org/show_bug.cgi?id=505521) that doesn't look like it's going to be fixed any time soon.

The fix is to subscribe to the `dragover` event on the whole window, which gives you [effectively the same result](https://bugzilla.mozilla.org/show_bug.cgi?id=505521#c95).

Tested on Linux+Wayland+KDE+(Firefox,Chromium) - I don't have anything else available at the moment, but I'll try to remember to test on some other platforms as soon as I'm back from the tournament.

Greetings from Prague Debate Spring!

### DIFF
--- a/tabbycat/templates/allocations/DraggableItem.vue
+++ b/tabbycat/templates/allocations/DraggableItem.vue
@@ -61,7 +61,6 @@ const hideHovers = () => {
   <div
     draggable="true"
     :class="['d-flex m-1 align-items-center align-self-center', draggableClasses]"
-    @drag="drag"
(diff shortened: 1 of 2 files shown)


tienne-B:
So good to hear from you again, @franga2000 ! Looks good and I've also tested this out locally.

PR #2901, opened 2026-08-01
Add speaker standings and metrics by score criterion

### LINKED ISSUE #2900: Create standings for speakers per criteria
Could be nice to show speakers where they rank for each component (criterion) of their speaker scores.

https://www.facebook.com/groups/tabbycat.debate/posts/3466617490172850

TrudeauOkech:
Fixes #2900

### DIFF
--- a/tabbycat/options/forms.py
+++ b/tabbycat/options/forms.py
@@ -2,6 +2,9 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import gettext as _
 from dynamic_preferences.forms import preference_form_builder, PreferenceForm
 
+from standings.speakers import SpeakerStandingsGenerator
+
+from .fields import EMPTY_CHOICE
 from .preferences import tournament_preferences_registry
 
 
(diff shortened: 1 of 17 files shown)


tienne-B (commented):
How would this work with reply vs substantive criteria, and could you explain how the API now handles criteria scores? How could users get an API endpoint with the received criteria scores per round?

TrudeauOkech:
Criteria only aggregate over the speeches they apply to ie `get_position_filter()` mirrors `applies_to_position()`, and the per-round columns use the same filter.

One case your question turned up: the criterion tab was inheriting its eligibility filter from the substantive standings view, so it ranked on `standings_missed_debates`, which counts substantive speeches. In WSDC the reply comes from one of the substantive speakers. Reply-only criteria now use `standings_missed_replies`/`replies_count`.

TrudeauOkech:
On the API, each criterion adds `criterion_avg_<seq>` and `criterion_total_<seq>`, which work in `metrics` and `extra_metrics`:

    GET /api/v1/tournaments/<slug>/standings/speakers?metrics=criterion_avg_1

Per-round criteria scores aren't available anywhere at the moment. `BallotSerializer` accepts `criteria` on write and `speakers_as_dicts()` emits a `criteria` key, but on read that value is `criteria_scores[side][pos]` — a dict keyed by `ScoreCriterion` — while `CriteriaSerializer` expects objects with `.criterion`/`.score`. With no criteria defined a ballot GET returns 200 and just leaves the key out; with criteria it 500s on `'DebateSide' object has no attribute 'pk'`.

This PR doesn't touch either file, so it's not from here, but making the read path emit `[{"criterion": <url>, "score": n}, ...]` is what would actually get you per-round scores. I'm thinking of doing it in a separate PR.

tabbycat/options/forms.py
tienne-B, line 1: So the new metrics do show in the preference fields; however, trying to save them results in a ValidationError like "criterion_avg_1 is not a valid choice".

tabbycat/standings/speakers.py
```diff
@@ -259,6 +261,99 @@ class SpeakerScoreRankingsMetricAnnotator(SpeakerScoreQuerySetMetricAnnotator):
...
+    def choice_label(cls):
+        return (cls.name_format % {'criterion': cls.criterion_name}).capitalize()
+
+    def get_position_filter(self, round):
+        """Mirrors ScoreCriterion.applies_to_position()."""
+        if self.criterion.speech_type == ScoreCriterion.SpeechType.REPLY:
+            return Q(speakerscore__position=round.tournament.reply_position)
+        if self.criterion.speech_type == ScoreCriterion.SpeechType.SUBSTANTIVE:
+            return Q(speakerscore__position__lte=round.tournament.last_substantive_position)
+        return Q()
```
tienne-B: As the metrics for reply are fixed and don't have a setting to customize, I don't think we need to worry about reply speeches actually.

```python
Q(speakerscore__position__lte=round.tournament.last_substantive_position)
```
should always be applied (or based on what page is being viewed, between reply or substantive standings, so that it works with `SpeechType.ALL`)

tabbycat/standings/speakers.py
```diff
@@ -259,6 +263,102 @@ class SpeakerScoreRankingsMetricAnnotator(SpeakerScoreQuerySetMetricAnnotator):
...
+
+def criterion_metric_annotator_classes(tournament, standalone_criterion=None):
+    """Returns the metric annotator classes for every score criterion in the
+    tournament, keyed by metric key. Each binds its criterion, so that they can
+    be used like the classes in SpeakerStandingsGenerator's static dict.
+
+    `standalone_criterion`, if given, is the criterion whose page this is, and
+    so whose metrics should be labelled without repeating its name."""
+    classes = {}
+    for criterion in tournament.scorecriterion_set.all():
```
tienne-B, line 350: Exclude reply-only criteria?

Issue #2655, opened 2025-05-25
Enable judge-on-judge feedback for elimination rounds


tienne-B:
From [FB](https://www.facebook.com/groups/284702441616866/?multi_permalinks=9989383961148617):

> 1. Should we enable judge-on-judge feedback for outrounds?
>
> CAs, in spite of their best efforts, often get an incomplete picture of judge feedback after inrounds. As such, they make on-the-margins decisions based off of outround deliberations; this seems fair and meritocratic. However, this system has some flaws;
> 1. it compartmentalises feedback - you only get additional information from those who you ask, as opposed to a fuller picture. This is less important when a CA is themselves evaluating a judge in an outround deliberation and more so when they are unable to, and either don’t observe this judge at all or resort to a shadow observation from a trusted judge. 
> 2. It makes the aggregation of feedback more informal - this may seem marginal but the volume of information a CA team deals with in outround allocations is rather massive. The less of this information is informal - and thus, easy to forget or overlook - the better.
>
> It may seem like this is solving a rather narrow problem - giving information on the panel management quality of some judges at particularly big tournaments. But essentially it’s just adding an extra dataset to be used in any number of ways. People may have concerns about how outround data is noisier - debates are tighter, delibs more intense on occasion. But not only is it possible to opt out (you can simply unweight or unconfirm the feedback where useful), but often the most useful data is that which reflects the skill of chairing or panelling an outround effectively - which, as labelled above, is often very different to an in round. I think more data in an easily digestible manner is great and we should take more where we have an easy opportunity to!

PR #2903, opened 2026-08-01
Implement feedback collection preferences for elimination rounds

TrudeauOkech:
#2655


tabbycat/options/preferences.py
tienne-B, line 1: I'd think the feedback expectations would remain the same between rounds; why are we copying the feedback paths and not a pref like "Expect team feedback for these rounds" with options for prelim only or prelims + elims ?

How should this work for silent rounds?

TrudeauOkech: I've dropped feedback paths and replaced with two scope preferences (Expect feedback from adjudicators in and Expect feedback from teams in, each prelims only / prelims + elims, defaulting to prelims only).

For silent rounds, matching prelims today (teams are never expected to give feedback in a silent round in either stage, while adjudicators still are). 

What do you think? Does this work better?

docs/features/adjudicator-feedback.rst
```diff
@@ -104,6 +104,30 @@ to give feedback on all adjudicators including trainees. The options that only
...
+rounds too. These options only control *which rounds* feedback is asked for in.
+
+Teams are never asked for feedback in silent rounds, in either stage.
+
+.. note:: Feedback submitted during elimination rounds is recorded, displayed in
+  the feedback tables and on each adjudicator's feedback page, and counted
+  towards feedback progress. It does **not** count towards adjudicators' scores,
+  and is not included in the statistics on the feedback overview page. This is
+  deliberate: outround feedback tends to be noisier, so it is collected for
+  reference without disturbing the scores used by the automated allocation.
```
tienne-B, line 129: @TrudeauOkech Where did this requirement come from?

I don't see this rule being implemented in the code though...

If it's a real requirement, it should be made "ignored by default" so that adj cores can decide whether to include specific pieces. (So you can set `ignored=is_elim` when creating feedback)

PR #2801, opened 2026-01-12
Feature/schedule separation by day

JWPCode:
Changed the add scheduled events to table function so that it segregates them by day of the tournament. I loaded a fake mini schedule in to see how the visual separation works. Open to feedback to make changes. 
Screenshot 2026-01-12 at 9 31 07 pm

fixes #2758

PR #2911, opened 2026-08-15
Improve tournament schedule editor UX

idowolf:
image

## Summary

This PR replaces the dense schedule formset with a day-grouped, responsive editor while preserving the existing `ScheduleEvent` and Django formset workflow.

**Problem:** The current page repeats every field label and help text for every event, includes fixed blank forms, and makes schedule-wide editing difficult to scan. Native datetime controls also obscure the time at common admin widths.

**Solution:** Events are presented as compact rows grouped by local date, with time-only editing synchronized to the underlying datetime form fields. Administrators can add days and events dynamically, duplicate or delete rows, move events between dates, preview the public schedule, and drag events between time slots. Reordering deliberately moves the event into the destination time slot so persisted and public ordering remain chronological.

### Schedule editing flow

- Groups bound schedule forms by their local start date in the view and exposes an empty form template for client-side additions.
- Uses an explicit browser-compatible datetime format so existing values populate correctly.
- Retains Django's formset validation, CSRF token, edit permissions, and normal POST/save behavior.

### Editor interaction

- Adds day and event controls without a fixed limit of blank forms.
- Supports drag-and-drop reordering within a day using the existing allocation UI's handle and drag-state conventions.
- Supports keyboard reordering with the arrow keys from the focused drag handle.
- Synchronizes compact 24-hour controls with the submitted datetime values and provides a calendar action for moving an event to another day.
- Adds duration feedback, duplication, deletion, event-type highlighting, unsaved-change feedback, and a participant-facing preview.

### Automatic titles

- Makes the custom title optional. A blank title resolves to the selected round followed by the event type (or just the event type when there is no round).
- Shows the exact automatic title in the editor and updates it live as either dropdown changes.
- Keeps blank as the stored automatic state, so a later event-type or round change cannot leave a stale generated title behind.
- Uses the resolved title consistently in the public schedule and admin display, and exposes it as `display_title` in the API without changing the writable `title` field's meaning.

### Presentation

- Splits the editor into schedule, day, and event-row templates with page-specific JavaScript and SCSS.
- Adds responsive layouts for desktop and mobile, including a sticky save action and accessible labels for icon controls.

### Tests

- Adds view coverage for local-date grouping, browser-compatible datetime values, and dynamically created formset events.
- Covers automatic titles with and without rounds, custom-title overrides, blank-title form submission, and API representation.

## Testing

- `python tabbycat/manage.py test tournaments.tests.test_lookup tournaments.tests.test_round_checks tournaments.tests.test_schedule`
- `python tabbycat/manage.py test tournaments.tests.test_schedule`
- `python tabbycat/manage.py check`
- `python tabbycat/manage.py makemigrations tournaments --check --dry-run`
- `npm run build-vue`
- Sass compilation
- ESLint, Stylelint, and Flake8 on the changed source
- Browser checks at desktop and mobile widths for drag-and-drop, keyboard reorder, add-day behavior, synchronized time slots, automatic-title updates, public preview, field alignment, and horizontal overflow


tabbycat/tournaments/templates/schedule_editor.js
tienne-B, line 1: Looking at all this, I think using Vue would be the better call here; all this Vanilla JS feels quite brittle and does quite a lot that should be easier using a frontend framework.

It should also still be possible to use the Django formset with Vue by aligning field names and then serializing values and errors (but I also wouldn't be opposed to using API calls here like with motion release)

tabbycat/tournaments/templates/tournament_schedule_edit.html
```diff
@@ -6,15 +6,122 @@
...
+            Drag an event into another row's time slot to reorder it. Start and end times follow the slot.
+          {% endblocktrans %}
+        </small>
+      </div>
+      <div class="btn-group schedule-mode-switch mt-3 mt-lg-0" role="group" aria-label="{% trans "Schedule view" %}">
+        <button class="btn btn-secondary active" type="button" data-schedule-mode="edit">
+          <span data-feather="edit-3"></span>{% trans "Edit schedule" %}
+        </button>
+        <button class="btn btn-outline-secondary" type="button" data-schedule-mode="preview">
+          <span data-feather="eye"></span>{% trans "Public preview" %}
```
tienne-B: Does this preview align well with the actual public view? Have a screenshot?

Maybe implement day separation as was started in #2801 for the public view? (I'll reach out to them again)

tienne-B:
A few smaller things that you might want to consider too:
* Try using input fields with `type="time"` rather than `type="text"` with RegEx
* Do we need all those colours in the select fields, and the "AUTO" indication/style for titles?
* Make sure everything, including date/time/duration formats are localizable

tabbycat/tournaments/templates/schedule_editor.js
idowolf: Implemented in d13ffaffc. The editor is now Vue 3, split into a container, day and event-row components plus a schedule composable. It keeps the existing Django formset as the persistence boundary: Django serializes values, choices, permissions and errors through `json_script`, while Vue renders the exact formset field names and management fields. Invalid POSTs rehydrate their raw values/errors, deleted persisted forms remain submitted with `DELETE=on`, and CSRF plus server-side permission/validation behavior are unchanged. The drag handle reuses the existing generic `useDraggable` composable.

tabbycat/tournaments/templates/tournament_schedule_edit.html
idowolf: Implemented in d13ffaffc. The editor preview now uses the same `TablesContainer` component and table data shape as the public page. The public schedule is also separated into one full-width table per tournament-local day, following the direction discussed in #2801, with localized day headings and time-only cells.

Editor preview:

Schedule editor preview

Actual public schedule:

Public schedule

idowolf:
Addressed the remaining review feedback in d13ffaffc:

- Replaced the vanilla schedule editor with structured Vue 3 components while retaining one Django formset POST.
- Switched the visible controls to native `type="time"` inputs and widened them so localized 12-hour suffixes are not clipped.
- Removed event-type colours, the `AUTO` badge and dashed automatic-title styling. The live “Leave blank to show …” explanation remains.
- Routed editor strings through `useDjangoI18n`; day/time presentation uses locale-aware formatting, and the public tables use Django localized date/time formats.
- Reused `TablesContainer` for the preview and separated the public schedule into one table per local day.
- Preserved invalid values/errors, read-only mode, stable form indices, hidden deletion forms, CSRF, and server-side validation.
- Escaped user-entered titles before passing them to the table renderer.

Desktop editor:

Desktop schedule editor

Mobile editor:

Mobile schedule editor

Validation completed:
- 24 Django tournament tests pass.
- Vue production build and Sass build pass.
- ESLint, Stylelint, Flake8, `manage.py check`, and `makemigrations tournaments --check --dry-run` pass.
- Browser-tested add/save/delete/reload, mouse drag, keyboard reorder, desktop/mobile layouts, preview/public rendering, native localized times, and the hidden `DELETE=on` formset protocol.

PR #2914, opened 2026-08-18
Fix postgres:18 volume mount path in docker-compose.yml

maclask:
postgres:18+ images use a pg_ctlcluster-style, version-specific data directory layout and refuse to start when a volume is mounted directly at `/var/lib/postgresql/data`, even if that volume is empty.

This mounts the parent `/var/lib/postgresql` directory instead, so Postgres manages the versioned subdirectory itself — matching the image's own suggested configuration for `pg_upgrade --link` support.

Tested against a real deployment migrating from `postgres:12` to `postgres:18`: the `db` container failed to start with the old mount path (`Error: ... there appears to be PostgreSQL data in: /var/lib/postgresql/data (unused mount/volume)`) and started cleanly once switched to the parent directory.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       - POSTGRES_USER=tabbycat
       - POSTGRES_DB=tabbycat
     volumes:
-      - pgdata:/var/lib/postgresql/data
+      - pgdata:/var/lib/postgresql
 
   redis:
     image: redis:6


tienne-B:
@maclask Sorry for the delay, I'm just a little uncertain how I should treat a PR directly to `master`... This was fixed on `develop` with 60f0866f94865b86cc126ebb79cce0a3bbea772f, so maybe create a hotfix branch to get this fix out sooner?

maclask:
Ok, sorry, didn't mean to directly PR this to master. Great if it's already fixed on dev. Your call whether this needs to be a hotfix. From my point of view, I don't need it ASAP.

maclask:
Ok, sorry, didn't mean to directly PR this to master. Great if it's already fixed on dev. Your call whether this needs to be a hotfix. From my point of view, I don't need it ASAP.

tienne-B:
Hope it's fine that I've cherry-picked https://github.com/TabbycatDebate/tabbycat/pull/2921 though!

maclask:
Yeah, no problem

PR #2918, opened 2026-08-26
Add team column to speakers table on Private URL page

lazybones7:
### What
Adds a "Team" column to the Speakers table on the Private URLs page, so tab staff can match a URL to its team without cross-referencing elsewhere.

### Why
With large speaker lists, the table currently only shows name + URL. No way to see the entire team's URLs at a glance. 

### Notes
- Reuses the existing `_team_cell` helper, so it respects `team_code_names` and anonymisation settings automatically, same as other admin tables.
- Added `select_related('team')` to avoid an N+1 query.

Tested locally with code names on and off.

### DIFF
--- a/tabbycat/privateurls/views.py
+++ b/tabbycat/privateurls/views.py
@@ -93,14 +93,12 @@ class RandomisedUrlsView(RandomisedUrlsMixin, VueTableTemplateView):
         return table
 
     def get_speakers_table(self) -> TabbycatTableBuilder:
-        speakers = Speaker.objects.select_related('team').filter(team__tournament=self.tournament)
+        speakers = Speaker.objects.select_related('team').prefetch_related('team__speaker_set').filter(team__tournament=self.tournament)
         table = TabbycatTableBuilder(view=self, title=_("Speakers"), sort_key="name")
         table.add_speaker_columns(speakers, categories=False)
         table.add_column(
             {'key': 'team', 'tooltip': _("Team"), 'icon': 'users'},
-            [table._team_cell(speaker.team)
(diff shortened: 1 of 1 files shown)


tabbycat/privateurls/views.py
```diff
@@ -93,9 +93,15 @@ def build_link(person):
...
     def get_speakers_table(self) -> TabbycatTableBuilder:
-        speakers = Speaker.objects.filter(team__tournament=self.tournament)
+        speakers = Speaker.objects.select_related('team').filter(team__tournament=self.tournament)
         table = TabbycatTableBuilder(view=self, title=_("Speakers"), sort_key="name")
         table.add_speaker_columns(speakers, categories=False)
+        table.add_column(
+            {'key': 'team', 'tooltip': _("Team"), 'icon': 'users'},
+            [table._team_cell(speaker.team)
+                if not getattr(speaker.team, 'anonymise', False)
+                else table.BLANK_TEXT for speaker in speakers],
```
tienne-B: I don't think we need this; `team.anonymise` seems to only be for speaker standings, so it'll never be set here.

You may need to prefetch `team__speaker_set` as the team cell has the speaker list tooltip.

Also you might want to include the team emoji?

tienne-B (commented):
Looks good!  Just some little tweaks

tabbycat/privateurls/views.py
lazybones7: Hi, good catch. 
I've updated the PR with a new commit. 

I removed `team.anonymise` and prefetched `team__speaker_set`, and added `show_emoji=True`

PR #2919, opened 2026-08-27
Update scoring and draw rules for AP,BP Presets

lazybones7:
### What has changed: Asian Parliamentary

- Changed AP Speaker Score limits from 69-81 to 67-83. Changed replies to half of it: 33.5-41.5
https://docs.google.com/document/d/1JoJa0oqDfW06vAQb3eBcAX37oG9p2g0hRO44vvCHv_Q/edit?tab=t.0
[Appendix A, Page 29]

- Changed Odd bracket Resolution for AP from Pull-Up at Top to Intermediate Brackets with Bubble-up-Bubble-Down
https://docs.google.com/document/d/1yoRcSR3mufyzOTxbOTxnGvdfxS-ODscVPN3CzzoD3mQ/edit?tab=t.0
[Section 20, Page 12]

### What has changed: British Parliamentary

- Changed Upper Limit for Speaker Score from 99 to 100
https://drive.google.com/file/d/1IvBju1RDN1WH_lK-8DxSsVFQuEurq0pN/view
[Appendix A, Page 62]

- Added Odd bracket Resolution for BP: Pull-Up at Random*
https://docs.google.com/document/d/19Hk8imODwOIr6zLCUUwqpAhSamoSqmL0ZdoeemY_XD0/edit?tab=t.0
[Section 36.2.3.1]

---- 
Edit:

I'm unsure if BP Draw Generator Graph even uses the `_pullup_filter`. I think the last change I've added under BP might not even be required. I just realized that the BP Draw Generator already has this set.

### DIFF
--- a/tabbycat/options/presets.py
+++ b/tabbycat/options/presets.py
@@ -208,11 +208,12 @@ class BritishParliamentaryPreferences(PreferencesPreset):
     description  = _("2 vs 2 vs 2 vs 2. Compliant with WUDC rules.")
     show_in_list = True
 
+    # WUDC Constitution: https://docs.google.com/document/d/19Hk8imODwOIr6zLCUUwqpAhSamoSqmL0ZdoeemY_XD0/edit?tab=t.0
     scoring__score_min                         = Decimal('50')
-    scoring__score_max                         = Decimal('100') # From the latest WUDC Constitution
+    scoring__score_max                         = Decimal('100') # WUDC Schedule 1: 5.3
     scoring__score_step                        = Decimal('1')
     scoring__maximum_margin                    = 0.0
-    scoring__teamscore_includes_ghosts         = True  # WUDC 34.9.3.2
+    scoring__teamscore_includes_ghosts         = True  # WUDC 35.9.3.2
     # Debate Rules
     debate_rules__substantive_speakers         = 2
     debate_rules__teams_in_debate              = 4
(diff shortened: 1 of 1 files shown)


tabbycat/options/presets.py
```diff
@@ -227,6 +227,7 @@ class BritishParliamentaryPreferences(PreferencesPreset):
     # Draw Rules
     draw_rules__avoid_same_institution         = False
     draw_rules__avoid_team_history             = False
+    draw_rules__draw_odd_bracket               = 'pullup_random' # From the latest WUDC Constitution
```
tienne-B: The BP draw generator doesn't look at this setting so we don't need to set it

tabbycat/options/presets.py
```diff
@@ -359,18 +360,18 @@ class UADCPreferences(AustralsPreferences):
 
     # Rules source = https://docs.google.com/document/d/10AVKBhev_OFRtorWsu2VB9B5V1a2f20425HYkC5ztMM/edit
```
tienne-B: Could you update the links so that the sources are in the code?

tienne-B (commented):
Thanks for the updates; presets are hard to keep up-to-date.

We should change the comments though:
* Don't mention "latest" (rather, cite your sources with links above)
* Keep section numbers in the comments (like "From Rules 20.10") for quick reference
* The UADC scoring is from the handbook, not the constitution (and the reply score range _is_ now specified)

tabbycat/options/presets.py
lazybones7: Yea my bad, I kinda realized it after I made the commit that I should have mentioned the source in the code itself and not in my PR description. I'll commit the new changes. 

I will update the sources 👍

lazybones7:
Yea, I've made the changes now, let me know if there are any other changes required. I should have done this even before raising the PR. 😢

PR #2924, opened 2026-09-02
Add editable side pre-allocations matrix

tienne-B:
## Problem

Side pre-allocations can only be reviewed as a read-only matrix, making it cumbersome to lock individual team sides or seed a future round from sides already used in a confirmed draw.

## Solution

Make the next undrawn round editable inline with a reusable AJAX select table cell and save indicator. Drawn rounds show actual sides, while a bulk tool can copy the same or opposite sides from a confirmed round into any undrawn preliminary round. Changes are permission-checked and action-logged.

## Testing

- Open **Side Pre-Allocations** with confirmed and undrawn rounds; change a side, confirm the save indicator updates, and refresh to verify it persists.
- Sort the editable round column and copy the table; verify both use the current selected values.
- Bulk-apply same and opposite sides to an undrawn round; verify teams with no source result remain unchanged.
- Attempt to edit or bulk-apply to a round with a generated draw; verify the change is rejected.


tabbycat/draw/views.py
```diff
@@ -946,15 +950,155 @@ def get_table(self):
...
+                header = {'key': f'round_{round_.id}', 'title': escape(round_.abbreviation)}
+                data = [tsas.get((team.id, round_.seq), "—") for team in teams]
+                table.add_column(header, data)
+            return table
+
+        # Admin page: rounds with a generated draw show the team's actual
+        # side (read-only, since editing the pre-allocation at that point is
+        # a no-op); the first round without a draw yet is editable; any
+        # further undrawn rounds show their pre-allocation read-only.
+        editable_round = next((r for r in rounds if r.draw_status == Round.Status.NONE), None)
```
tienne-B, line 965: Why not allow editing any future round?

tienne-B (commented):
@PiechZ Thank you for the branch! I've left a question for you, but I hope we can get this past the finish line!

tabbycat/draw/views.py
PiechZ: I just did not want to make it too overwhelming and as as a proof of concept the next round was enough. However if the number of dropdowns does not bother anyone, it can definitely be set up for all rounds.

PR #2925, opened 2026-09-02
Add Karl Popper debate format support

tienne-B:
## Problem

Tabbycat does not currently support key Karl Popper rules: median aggregation across adjudicators, a solo adjudicator declaring a 2–1 split, or a preset that configures the format consistently.

## Solution

- Add a configurable panel score aggregation function, with mean as the default and Karl Popper median behavior as an option. Final team totals are derived from the aggregated speaker scores so result displays and standings remain consistent.
- Add self-declared 2–1 splits for genuinely solo-adjudicated voting debates, including ballot entry, merged results, vote totals, and result displays. Trainees remain excluded from split labeling.
- Add a Karl Popper preference preset covering its debate, scoring, draw, standings, feedback, and data-entry configuration.
- Add regression coverage for aggregation, self-split behavior, form visibility, and preset configuration.

## Testing

1. Create a tournament, apply the Karl Popper preset, and confirm that median scoring, per-adjudicator ballots, adjacent pairing, and self-split ballots are enabled.
2. Enter a three-adjudicator result with different speaker scores and confirm that each final speaker score is the median and each team total equals the sum of those final speaker scores.
3. Enter a solo-adjudicator result, select the self-declared split checkbox, and confirm the winner records 2/3 votes and the loser 1/3.
4. Add a trainee to that solo allocation and confirm the trainee is not labeled as self-declared split.
5. Open an individual ballot for a panel debate and confirm that the self-split checkbox is not shown.


tabbycat/results/templates/ballot/ballot_set.html
tienne-B, line 1: Should this field have more prominence, like if it were a full-width dropdown with the declared winner field?

tabbycat/results/result.py
```diff
@@ -64,6 +65,19 @@
 logger = logging.getLogger(__name__)
 
 
+def median_or_rounded_up_mean(values):
+    values = list(values)
+    if len(values) % 2 == 1:
+        return median(values)
+    return math.ceil(mean(values))
```
tienne-B, line 72: So the median is just so that it stays an integer? Should `statistics.median_high` be used instead, and a note in the option as it's not quite a median?

tabbycat/results/result.py
```diff
@@ -1174,18 +1201,17 @@ def teamscorebyadj_field_margin(self, adj, side):
...
-        return sum(self.get_score(adj, side, pos) for pos in self.positions if not self.get_ghost(side, pos))
-
     def teamscore_field_score(self, side):
         # Should be decision-decorated
         if not self.is_complete():
             return None
         if not self._decision_calculated and len(self.sides) == 2:
             self._calculate_decision()
-        return mean(self._teamscore_score_component(adj, side) for adj in self.relevant_adjudicators())
+        return sum(
```
tienne-B, line 1210: I think this is dangerous as it is meant to be the avg of all the voting/non-dissenting adjs for that team score

tienne-B (commented):
@PiechZ Got some comments for this one as well!

tabbycat/options/presets.py
```diff
@@ -621,6 +621,76 @@ def are_some_draws_not_random(tournament):
     standings__team_standings_precedence       = ['speaks_avg']
 
 
+class KarlPopperPreferences(PreferencesPreset):
+    name         = _("Karl Popper")
```
PiechZ: As far as I know, there are many varieties of Karl Popper format in other European organizations. I am thinking  whether it would be safer to add some note that this is applicable for the Czech version of the rules.

tabbycat/results/result.py
PiechZ: Yes, median was historically probably chosen because we need integer as a result for each speaker from the debate. 
In vast majority and in all competition debates, there is uneven number of adjudicators. Should it happen (very improbably) that there are two judges (for example if some judge steps down during a round), I needed some fallback option. That is an average from the two judges rounded up to an integer. and that is how we are calculating it, in that case we are not using `statistics.median_high`.

PiechZ: Maybe to add a note that in even numbered panels, this result into average rounded up?

tabbycat/results/result.py
PiechZ: In Karl Popper we include scoring of dissenting judges, I might have not taken into account the edge case of a configuration that would use median + margin includes dissenters to false. 
Do you have any recommendation?

tienne-B: I mean the change from `mean` to `sum`.

tabbycat/options/presets.py
```diff
@@ -621,6 +621,76 @@ def are_some_draws_not_random(tournament):
...
+    debate_rules__reply_scores_enabled         = False
+    debate_rules__require_substantive_for_reply = False
+    debate_rules__winners_in_ballots           = 'high-points'
+    debate_rules__preparation_time             = 60
+    debate_rules__enable_forfeits              = True
+    motions__enable_motions                    = False
+    motions__motion_vetoes_enabled             = False
+    motions__enable_motion_reuse               = True
+    # Draw Rules (penalty-weighted, as with APDA)
+    draw_rules__draw_odd_bracket               = 'pullup_random'
```
PiechZ: After organizational discussion, this was just recently changed to 'pullup_top'
I already have it in the fork as a commit.

tabbycat/options/presets.py
PiechZ: I am thinking about adding some descritpion, as this covers the current Karl Popper rules for Czechia. I know that some other organizations have their own, more or less used version of Karl Popper, that have differences.

tabbycat/results/templates/ballot/median_overview.html
```diff
@@ -1,54 +0,0 @@
-{% load debate_tags i18n team_name_for_data_entry %}
```
PiechZ: I know that this really does not suit the mainline Tabbycat, we have it as a practical helper for rewriting the debate results elsewhere, maybe we can have it just in a fork.

tabbycat/results/templates/ballot/ballot_set.html
PiechZ: That would be even better! However I wanted to sopmehow leave it as a checkbox, or maybe something similar to really have the non-self-split option as a default, as self-split is less usual than "normal" unanimous result.

tabbycat/results/result.py
PiechZ: I am sorry that I did not get it at the first sight, If I understand it correctly, I was unfortunately doing team speaker score as a median of sums of voting adjudicators, but it should really be sum of medians of voting adjudicators. 

Your fix should resolve it, AFAIK. I tried to do similarly in the fork.

PiechZ:
Hello @tienne-B , thank you very much for review, it was really valuable. I added some fixes in the fork, since I did not want to add unnecessary stuff here - https://github.com/szymsza/tabbycat-adk/pull/3

tabbycat/results/templates/ballot/median_overview.html
tienne-B: > rewriting the debate results elsewhere

What do you mean by that? Could the confirmed ballot in the API be used?

PiechZ: We are manually rewriting the results into statistiky.debatovani.cz (older PHP app, that has no API). It is used for storing the results of all debates.

PR #2930, opened 2026-09-05
Add tester tracking for adjudication cores

mahornagda:
Adjudication cores commonly assess adjudicators by sitting on their panels, and
then have to keep track of who has been watched, in what position, and who is
still unseen. That tracking is usually kept in a spreadsheet beside the tab and
updated by hand, which goes stale as soon as a draw changes.

### The page

A **Tester Tracking** page under Feedback. Adjudicators are listed against the
rounds that have a draw, and each cell shows the position the adjudicator held
when a tester sat on their panel — `C`, `P` or `T`.

| cell | meaning |
|---|---|
| `C` / `P` / `T` | tested, in that position |
| `·` | judged that round, no tester present |
| empty | not allocated |

The last two are deliberately distinct; collapsing them loses the difference
between "we haven't got to them" and "they've judged plenty, unwatched". Every
cell links to the feedback that adjudicator received that round.

### Who is a tester

Two-part, because in practice it is: members of the adjudication core always
count, and anyone else can be marked on the page itself with a checkbox, in the
manner of the existing breaking checkbox. Adjudication core members cannot be
unticked, since the flag would have no effect on them.

None of the coverage is typed in. It is derived from the draw and fills in as
rounds are allocated.

### Two summary columns

**Tested** counts the rounds an adjudicator has been watched in. **Chair** says
whether a tester has ever seen them chair, which is usually what an
elimination-round chair appointment rests on — an adjudicator who has chaired
but never with a tester present is flagged `unwatched`, as that is the gap most
easily missed.

Testers themselves are left out of both columns and out of the summary counts.

### Supporting changes

- `AdjudicatorFeedback` views accept `?round=`, so a link can address one
  round's feedback. Feedback hangs off either a source adjudicator or a source
  team, and both routes to the round are matched — filtering on only one
  silently drops the other, which is easy to miss when a tournament's feedback
  happens to be mostly from one side.
- `CheckCell.vue` accepts a `disabled` flag, so a checkbox reflecting a derived
  value can say so rather than accepting a click and silently refusing to save.
  Additive; existing tables are unaffected.

### Tests

15 new: nine covering the tester logic (adj core counts without being marked, a
lone tester does not test themselves, trainees count as tested, judging without
a tester is recorded but not as tested, testers are excluded from the counts),
and six covering the round filter, including that it finds feedback from both
sources


tienne-B:
Let's scale down this work a bit, and it could have two parts:
* Adding a column to the Feedback Overview to mark an adj as tester
* Adding an indicator to the feedback trend for the rounds where an adj got tested

This way we don't need to create new pages and explain the new table.

mahornagda:
Makes sense, I'll cut the separate page and do it as the two parts you describe: a tester column on the Feedback Overview, and a marker on the feedback trend for rounds where they were tested.
One question before I rebuild it. The thing an adjudication core actually decides with this is whether someone can chair an out-round, and that usually rests on having been watched in the chair specifically - not just watched. In the version I sent, that was a separate column flagging adjudicators who had chaired but never with a tester present.
Would you want the trend marker to distinguish "tested as chair" from "tested", or is a single tested marker enough? Happy either way, just want to build it once.

tienne-B:
The feedback trend graph already has colours to indicate their position in the round, so only a "tested" marker is needed.

mahornagda:
Rebuilt as you described. The separate page is gone; it's now a Tester column on the Feedback Overview, and rounds where a tester sat on the panel are ringed on the feedback trend with "tested" added to the tooltip. You were right that the colours already carry the position, so the marker only had to say whether a tester was there.

Two notes on what's in the diff:
- CheckCell gains a disabled flag. Adjudication core members count as testers whether or not the box is ticked, so theirs is ticked and disabled - otherwise unticking one appears to work and silently does nothing. Additive, existing tables unaffected.
- There's a migration for the new ActionType, matching 0014_alter_actionlogentry_type. I left the users app alone - its pending migration bundles an unrelated Add field users to group that's already on develop.

PR #2931, opened 2026-09-06
Preserve previous table sorts when breaking ties

### LINKED ISSUE #1750: Result sorting algorithm not stable
*URL*

https://gdc2021-cuhksz.herokuapp.com/gdc2021/results/round/1/

*Issue*

After sorting by field 1, if you sort by field 2, first sorting is shuffled.

*Proposed solution*

Use stable sorting algo (merge sort perhaps)

be-student:
Resolves #1750.

Sorting a new table column currently discards the previous ordering, and descending sorts reverse rows even when they compare equal. For example, sorting names descending and then an equal score column changes Charlie/Bravo/Alpha back to a different order.

Keep the most recently selected columns as ordered tie-breakers, applying each direction inside the comparator. Equal rows retain their input order, updated rows use the same criteria, and source data remains unmodified. This is implemented in the shared `useSortableTable` composable used by SmartTable; filtering continues to consume its sorted output.

Adds five dependency-free Node test-runner tests over the actual Vue composable, an `npm test` command, and a CI step. The contribution guide documents the contract and test command.

Validation:
- Four initial regressions failed on the original code and pass with the fix.
- `npm test`: five passed, covering column changes, descending ties, default sorting, reactive row updates, and priority/direction changes.
- ESLint on the changed JavaScript files and `git diff --check`: passed.
- `npm run build-vue`: passed, with existing large-chunk warnings.

Full Django/Selenium tests and browser interaction were not run locally; this change is confined to sorting logic and its frontend test wiring. No schema or styling changes.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -46,7 +46,8 @@ Development
 Frontend table tests
 --------------------
 
-Run ``npm test`` to exercise the Vue table composables without a Django server.
+Run ``npm test`` to run all frontend Node tests discovered through the standard
+``*.test.js`` naming convention, without a Django server.
 Sorting a new column keeps earlier column sorts as tie-breakers, in the order
 most recently selected. Descending order must preserve rows that compare equal,
 and sorting must not mutate the source rows. These tests also run in Django CI.
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -43,15 +43,6 @@ Development
 
     $ npm run serve
 
-Frontend table tests
---------------------
-
-Run ``npm test`` to run all frontend Node tests discovered through the standard
-``*.test.js`` naming convention, without a Django server.
-Sorting a new column keeps earlier column sorts as tie-breakers, in the order
-most recently selected. Descending order must preserve rows that compare equal,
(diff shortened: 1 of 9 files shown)


tienne-B (commented):
Thank you for the JS testing infrastructure! Could it be made more general though, like the `test` command to test future test files as well? (Update the CONTRUBUTING file as well)

It's a good feature to be able to sort by many columns, but should we indicate in the header the order (like adding <kbd>n</kbd> to the sorted columns)? Should there be a different interaction if people want to "reset" the sorting rather than adding to it?

be-student:
I updated `npm test` to use Node's general test discovery and updated CONTRIBUTING accordingly, so future `*.test.js` files are included without changing the command. The current interaction makes the most recently clicked column primary, keeps older columns as tie-breakers, and toggles direction when the same header is clicked. I left priority labels and a reset interaction out of this patch because either would add a new UI contract; I can follow whichever direction you prefer for those.

.github/CONTRIBUTING.rst
tienne-B, line 37: Add the line about `npm test` here, and make it succinct/general

tienne-B (commented):
How much work would it be to show the sorting? It could be in a separate commit in the PR if it'd help. Just think displaying the sorting is part of the feature.

PR #2935, opened 2026-09-09
Add pullup flag to lower-bracket team in intermediate brackets

lazybones7:
## What

Adds `pullup` flag to the team that joins an intermediate bracket from the lower bracket. 

This adds the existing `pullup` flag to that team at the point when the intermediate bracket is formed, and correctly transfers the flag if that team is later swapped out via a bubble-down conflict resolution in `intermediate_brackets_with_bubble_up_down`.

## Why

Helpful for Tab Directors handling Asian Parliamentary debates as UADC's constitution (Section 20.10) proposes preventing multiple pull-ups of a team from the lower bracket to the intermediate bracket.

Tried and tested locally

### DIFF
--- a/tabbycat/draw/generator/common.py
+++ b/tabbycat/draw/generator/common.py
@@ -95,6 +95,15 @@ class BaseDrawGenerator:
         flags = self.team_flags.setdefault(team, list())
         flags.append(flag)
 
+    def remove_team_flag(self, team, flag):
+        """Removes a previously-attached flag from a team, if present.
+        Used when a team's bracket placement changes after a flag was
+        provisionally added (e.g. bubble down swaps a team out of the
+        pullup role in an intermediate bracket)."""
+        flags = self.team_flags.get(team)
+        if flags and flag in flags:
+            flags.remove(flag)
+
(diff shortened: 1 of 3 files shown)


lazybones7:
Looks like CI Tests are failing, I'll look into that and suggest changes as well

lazybones7:
Finally... , it passed

I have updated 'pullup' flags for the teams that are supposed to get a pullup flag in the test generator.py based on the new conditions. Let me know if these are good to go

lazybones7:
This really is a band-aid fix, because Bubble-up-bubble-down might not need the pullup flag at all, as Australs doesn't really care if a team has been in the intermediate bracket twice. But if Australs ever updates its constitution to minimize pullups, then this will still have its use case

I'll raise the issue now to add a pullup restriction for Intermediate Brackets. I guess we can work that one out together :)

lazybones7:
Actually @tienne-B 

Please don't merge this, there's an issue with this, pls keep this open. I just realized that while this approach solves Intermediate Brackets, it makes it worse for Intermediate with bubble up bubble down. 

I'll figure out another workaround and raise a PR

PR #2940, opened 2026-09-19
Fix API ballot submissions with ghost speeches

tienne-B:
## Problem

Submitting a ballot with a ghost speech raises a `TypeError` because the API serializer calls `set_ghost()` without its required `is_ghost` argument. The ballot request fails instead of saving the result.

## Fix

Pass `True` when marking a ghost speech. Update the consensus ballot test to submit one ghost speech and verify it is returned as ghost.

Fixes BACKEND-ETP

## Manual verification

1. Submit a consensus ballot through the API with one speech marked `ghost: true`.
2. Confirm the request returns 201 and the saved result marks that speech as ghost.
3. Submit a ballot without ghost speeches and confirm it still succeeds.


lazybones7:
I literally came across this issue on Saturday and was about to raise a PR myself. Ahhh you beat me to it @tienne-B 
D:

Solid work though 👌

PR #2943, opened 2026-09-22
Update current weighted tooltip to include formatted score for adjudicators

lazybones7:
## What

Adds the exact weighted score to the tooltip on the "Current weighted score" column in the adjudicator feedback overview table, shown to 3 decimal places. 

(I am assuming 3 decimal places is enough to give clarity on exact score; could be changed to 4 as well if required.
ex: if 2 decimals are used, 7.473 and 7.469 also may round up to 7.47, again causing confusion)

## Why

Clarifies score differences between adjudicators who share the same 1-decimal weighted score. Helps adj-core rank adjudicators more precisely.


tienne-B:
If we were to put error bars for adjudicator feedback scores, they would be quite wide. Differences past 1 decimal point seem meaningless, so we shouldn't worry about showing it.

lazybones7:
Yea actually that makes sense, I only raised it cuz some of the core-adjudicators in a recent comp got a little confused when giving out Adjudicator Awards and they couldn't write the exact score for the adjes in the rank list.

